PR #26, opened 2017-06-13
should fix link to auth0 site in readme

lady3bean:


lady3bean:
Might want to add instructions for what options to chose for auth0 account, like whether we want authentication through google and third parties for our app. Should we answer no, and chose only the database connection option?

schuyler1d:
closing -- i think this change was folded in to another PR.

PR #27, opened 2017-06-13
package.json tweaks

schuyler1d:
query-string has conflicting version deps, so upgrading, and faker was a dupe from regular dependencies


shakalee14 (commented):
Great!

PR #34, opened 2017-06-15
adds --bind all flag to rethinkdb command

lady3bean:
Didn't know if this was relevant for everyone, as we added it in during setup while trying to get things to work


lady3bean:
No longer relevant after changes - closing.

PR #37, opened 2017-06-15
swap nf env variables for unix env variables

sreynen:
Many commands in package.json scripts rely on environment variables. When calling those commands via node foreman, the environment variables get loaded from .env, but that makes it difficult to call individual commands separately, which is sometimes useful, especially in development. The scripts can be called directly with regular unix environment variables, and node foreman still works like that as well, though this does add an extra step of explicitly sourcing the variables, e.g. `source .spoke-env`.


.gitignore
```diff
@@ -3,4 +3,4 @@
 npm-debug.log
 .DS_Store
 rethinkdb_data
-.env
```
schuyler1d, line 6: let's keep .env in .gitignore since it's often used for local env regardless of project.

shakalee14:
hey @sreynen since you are trying to implement pm2 in development and prod on local env, can we close this p.r.?

sreynen:
https://github.com/MoveOnOrg/Spoke/pull/63 is a work in progress and may end up not happening at all, so I'd like to keep this as an option until it's definitely not useful.

Issue #24, opened 2017-06-10
Easter Egg Invite Link


shakalee14:
- [ ] As a user, I can invite myself to the repo without directly going to port 8080 via a link on the splash page

sandramchung:
So this is to workaround the current steps 12 and 13 in the Getting Started part of the README, which are:

> 12. Because Spoke is invite-only you need to generate an invite: Go to the RethinkDB data explorer at http://localhost:8080/#dataexplorer and run: r.db('spokedev').table('invite').insert({is_valid: true}). Copy the generated key.
> 13. Use the generated key to visit an invite link, e.g.: http://localhost:3000/invite/e7bcc458-c8e9-4601-8999-a489e04bd45f. This should redirect you to the login screen. Use the "Sign Up" option to create your account.

PR #40, opened 2017-06-15
https://github.com/MoveOnOrg/Spoke/issues/24

sandramchung:
https://github.com/MoveOnOrg/Spoke/issues/24 Creates a hidden link on the homepage to take you to organization creation via the creation of an invitation link that redirects you to login. Better than the current method which is to manually create an invitation in rethinkdb and construct a link from the ID. To test, check out this branch, logout of your Spoke instance if you're logged in, go to the homepage (localhost:3000) and mouse over the bottom of the green box to see the hidden link.


lady3bean:
I'm running into the proxy error again, whether on master or checkout out to this PR. But I think that's a problem on my end. The code looks good to me. Much invite!

src/containers/Home.jsx
```diff
@@ -101,5 +118,26 @@ const mapQueriesToProps = () => ({
   }
 })
 
+const mapMutationsToProps = () => ({
+  // createCampaign: (campaign) => ({
```
mathemagica: probably need to add this back in so the rest of the campaign admin page works again

src/server/api/schema.js
```diff
@@ -139,6 +139,11 @@ const rootSchema = `
     userId: String
   }
 
+  input InviteInput {
+    id: String
```
mathemagica, line 143: did we need these in the input object? we're not using them, so worth testing whether we can remove them and have the input obj be empty

src/containers/Home.jsx
```diff
@@ -34,6 +34,23 @@ class Home extends React.Component {
     orgLessUser: false
   }
 
+  handleOrgInviteClick = async (e) => {
+    e.preventDefault()
+    const newInvite = await this.props.mutations.createInvite({
+      id: 'cats',
```
mathemagica: do we need these inputs? we're not using them, so worth testing whether everything still works when you remove them

sandramchung: An empty object definitely doesn't work. {is_valid = true} does work, though - probably because that's the only required attribute in the model definition.

src/containers/Home.jsx
sandramchung: Campaign admin is reverted to original - just forgot to delete the campaign mutation after copy pasting this part here in the homepage

src/server/api/schema.js
sandramchung: Empty input object throws and error. Just {is_valid: true} works, probably bc that's the only required attribute on the invite model https://github.com/MoveOnOrg/Spoke/pull/40/commits/36843534f9600f16d7d1f96d75c7106cc328b7a7

src/containers/Home.jsx
sandramchung: https://github.com/MoveOnOrg/Spoke/pull/40/commits/40decfb0352d10bd84db89d1f1ceb0fb19745ce1

src/server/api/schema.js
mathemagica: 👍

mathemagica:
👍

PR #44, opened 2017-07-07
Db migration 6servicemessages

schuyler1d:
I think this is ready for a PR.
It does two things:
1. removes the array fields message.service_message_ids and message.service_messages
2. fix some bugs that came up from previous db migrations.

There is still at least one outstanding issue with the client/server, but will address that in next batch of work.

### DIFF
--- a/src/server/api/lib/nexmo.js
+++ b/src/server/api/lib/nexmo.js
@@ -18,7 +18,7 @@ async function convertMessagePartsToMessage(messageParts) {
   const firstPart = messageParts[0]
   const userNumber = firstPart.user_number
   const contactNumber = firstPart.contact_number
-  const serviceMessages = messageParts.map((part) => part.service_message)
+  const serviceMessages = messageParts.map((part) => JSON.parse(part.service_message))
   const text = serviceMessages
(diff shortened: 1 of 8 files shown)


src/server/api/lib/nexmo.js
```diff
@@ -177,8 +177,9 @@ async function handleIncomingMessage(message) {
 
   const pendingMessagePart = new PendingMessagePart({
     service: 'nexmo',
-    parent_id: parentId,
-    service_message: message,
+    service_id: message['concat-ref'] || message.messageId,
+    parent_id: parentId, //do we need this anymore, now we have service_id?
```
shakalee14, line 181: i don't think we need it anymore ---> parentId ( parentId = message['concat-ref'] ) was previously created the same way so I assume it's serving the same purpose as service_id

shakalee14 (commented):
looks good to me!

src/server/api/lib/nexmo.js
schuyler1d: i'm going to leave this in for now -- i'm assuming some more bugs might come out with more testing/use, and the archaeology will be easier.  then we can clean it up.

PR #45, opened 2017-07-10
interaction step migration: part 1

sandramchung:
These changes are transitional prep for the migration to a relational database. They: 

- add new model fields to interaction_step while retaining the old model fields so things that depend on them still work
- provide a way to migrate data from old model to transitional model
- alter the instance save worker method to save the data in the new model fields

To check that this all works:

1. In the UI, create a campaign with title "Cats" and add interactions (a script) with at least one question with multiple answer options (you can change the campaign name if you don't like it - just make sure the filter in the query below matches).
2. Run this in the [data explorer](http://localhost:8080/#dataexplorer) to inspect the (script) interactions data before migration and resave:
`r.db('spokedev').table('interaction_step').innerJoin(r.db('spokedev').table('campaign'),function(interaction_step, campaign){return interaction_step["campaign_id"]==campaign["id"]}).filter({"right": {"title":"Cats"}}).pluck("left")`
My app database is named 'spokedev'; if yours is named differently you'll need to change the name where it appears in the query.
4. Run the migration (see src/migrations/2017-06-30-interaction-step-answer-option.js) to copy answer-option data from old model fields to transitional model fields. Note that if your db name is not 'spoke' (mine is 'spokedev') you'll have to edit it in the code before running the migration.
5. Repeat step 2 and note that the answer options are now in the answer_option fields on the response steps (as well as in the answer_options array fields on the question steps) and that interaction_parent_ids point to the question step id.
6. Navigate to campaign admin and edit and save your Cats campaign script. It should not hang on save, and the updated data should be inspectable in the db if you repeat step 2 a third time.

The next step is to fix all the things that will break with the new (postgres-compatible) interaction-step model. Once that all works, we can cut away all the old fields and migrate entirely to the new model in a future PR.

### DIFF
--- a/src/migrations/2017-06-30-interaction-step-answer-option.js
+++ b/src/migrations/2017-06-30-interaction-step-answer-option.js
@@ -1,0 +1,36 @@
+import { r } from '../server/models'
+
+// For each answer_option on interaction steps with question:
+// - Move the value to the answer_option field of the child interaction step
+// - Set the parent_interaction_id of the child to the id of the parent
+// - TODO: Drop the answer_options field
+
+// Run this migration on a running dev or prod instance with the following command:
+// ./dev-tools/babel-run-with-env.js ./src/migrations/2017-06-30-interaction-step-answer-option.js
+
+(async function () {
+  console.log('migrating interaction steps to updated schema');
+  try {
+    const parentSteps = await r.db('spokedev')
+      .table('interaction_step')
+      .filter((step) => step('answer_options')
+      .ne([]))
+    console.log(parentSteps)
+    const parentCount = parentSteps.length
+    for (let i = 0; i < parentCount; i++) {
+      const answerCount = parentSteps[i]['answer_options'].length
+      for (let j = 0; j < answerCount; j++) {
+        const parentId = parentSteps[i]['id']
+        const answerOption = parentSteps[i]['answer_options'][j]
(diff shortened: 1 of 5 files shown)


src/migrations/2017-06-30-interaction-step-answer-option.js
```diff
@@ -0,0 +1,36 @@
...
+// Run this migration on a running dev or prod instance with the following command:
+// ./dev-tools/babel-run-with-env.js ./src/migrations/2017-06-30-interaction-step-answer-option.js
+
+(async function () {
+  console.log('migrating interaction steps to updated schema');
+  try {
+    const parentSteps = await r.db('spoke')
+      .table('interaction_step')
+      .filter((step) => step('answer_options')
+      .ne([]))
```
shakalee14, line 17: what is this ```.ne([])) ``` doing?

sandramchung: Grabbing all instances where the answer_options array isn't empty. There's probably a better way to do it.

shakalee14: ok, so when ``answer_options`` isn't equal to an empty array, than it is constitutes as ``parentSteps``

schuyler1d:
:+1: looks good so far!

PR #48, opened 2017-07-12
Db migration nogeo

schuyler1d:
removing point from zip code since it's just unused data (for now)


shakalee14 (commented):
The change is representing location via two objects, latitude and longitude vs 1 nested location object? looks good to me.

PR #49, opened 2017-07-12
Webpack host environment variables

sreynen:
The environment variable named `JOB_DB_HOST` was being used to define `webpackHost`, not the actual job database host, so this changes the name to reflect that. Also, `JOB_DB_PORT` wasn't being used at all, while we do have a use for `WEBPACK_PORT`, so changing that as well. I imagine there probably is some potential use for `JOB_DB_` variables, but having them defined before we're actually using them seems to only add potential confusion.


schuyler1d:
:+1: doubly-good because JOB_DB_PORT wasn't used anywhere!  so confusing

webpack/server.js
```diff
@@ -3,9 +3,9 @@ import webpack from 'webpack'
 import config from './config'
 import { log } from '../src/lib'
 
-const webpackPort = 3000
+const webpackPort = process.env.WEBPACK_PORT
```
schuyler1d: maybe make this `const webpackPort = process.env.WEBPACK_PORT || 3000`
(and below: `const webpackHost = process.env.WEBPACK_HOST || 'localhost'` )

shakalee14 (commented):
looks good to me.

shakalee14:
@sreynen you want to merge this?

PR #52, opened 2017-07-17
adds stream-http module

lady3bean:


shakalee14:
sorry @lady3bean - what we were we adding stream-http for? should we merge this?

shakalee14:
i'm going to close this since it seems specific to your env?

PR #61, opened 2017-07-22
Db migration 3interactionsteps

sandramchung:
Yak shaving to make campaign_contact.questionResponses match pre-model change campaign_contact.questionResponses without array fields or `concatMap`. Verify by querying at localhost:3000/graphql with a database containing a campaign with an interaction script that has at least two question steps and at least two answer options for each question step; compare query results with master branch code to results with this branch's code. Sample query (change id to match your db):

```
{
  contact(id: "917987c3-929c-4c01-9fcd-61fbf1885edd") {
    questionResponses {
      value
      nextInteractionStep {
        id
        question {
          text
        }
        script
        questionResponse {
          id
        }
      }
      responders {
        id
      }
      responderCount
      question {
        text
        answerOptions {
          value
          responderCount
        }
        interactionStep {
          id
        }
      }
    }
  }
}
```

Also worth noting that this will probably not scale well - i.e. problems if there are a lot of contacts with question responses in a single campaign.


schuyler1d:
++ to the hard part :-)
Please actually delete the `answer_options` in the save and the model here:
https://github.com/MoveOnOrg/Spoke/blob/db-migration-3interactionsteps/src/workers/job-handler.js#L85
https://github.com/MoveOnOrg/Spoke/blob/db-migration-3interactionsteps/src/server/models/interaction-step.js#L21
(and then check the box here!)
https://github.com/MoveOnOrg/Spoke/blob/db-migration-3interactionsteps/docs/DB_MIGRATION.md#array

shakalee14 (commented):
LGTM!

sandramchung:
Ddoh, it's not done yet. I thought I'd tested with the new model in place, but it was still the transitional model - and there's another broken query to fix.

src/workers/job-handler.js
```diff
@@ -56,21 +62,34 @@ async function createInteractionSteps(job) {
...
 
-    interactionSteps.push({
-      id: newId,
-      campaign_id: id,
-      question: step.question,
-      script: step.script,
-      answer_option: answerOption,
-      parent_interaction_id: parentId
-    })
+    dbInteractionStep = await InteractionStep
```
sandramchung: Save fails for me after the first step unless this is `const dbInteractionStep = blah`

sandramchung:
Ready for re-review

src/server/api/campaign-contact.js
```diff
@@ -65,22 +65,70 @@ export const resolvers = {
...
+                "id": responseId,
+                "parent_interaction_id": questionStepParentInteractionId,
+                "question": question,
+                "script": script
+            },
+            "value":  responseValue
+          }
+        }
+      }
+      let responses = []
```
sreynen: Linting suggests to me that arrays should always be `const`, I think because the actual data in the variable is just a pointer to the start of the array and the pointer doesn't change when the array content changes.

src/server/api/campaign-contact.js
```diff
@@ -65,22 +65,70 @@ export const resolvers = {
...
+      for (var key in formatted) {
+        responses.push(formatted[key])
+      }
+      return responses
```
sreynen: I think this and the previous lines could be simplified to `return formatted.values()`. Babel will probably convert it back to something like this, but the ES2015 syntax makes it a little easier to read.

sandramchung: Interestingly, `formatted.values()` results in the error `formatted.values is not a function`, but `Object.values(formatted)` gives me what I want here.

sandramchung:
Okay, i think this is finally ready to merge

Issue #51, opened 2017-07-14
Remove gearshift.co references


sreynen:
Spoke was originally a project of gearshift.co and the codebase still has a lot of references to that. Now that it's open source, let's replace those references with more relevant links and email addresses.

sandramchung:
Includes https://github.com/MoveOnOrg/Spoke/issues/46

sreynen:
Related, also need to remove references to "Axle Factory", the company name.

PR #63, opened 2017-07-24
WIP: simplify config using PM2 for dev

sreynen:
This began with looking at https://github.com/MoveOnOrg/Spoke/issues/51, where a lot of hard-coded pieces of text related to the organization hosting a spoke instance need to be moved to configuration. Currently configuration is stored in many different places, so it would add a good deal of complication if implementing orgs needed to add their org name, privacy policy, contact email, etc. to different files for different contexts. So, this is an attempt to consolidate on a single config file.

Since we're using pm2 for prod and that pulls config directly from JS, the goal here is to get everything else pulling from that same JS file. I think the simplest way to do that is to make dev look more like prod. Specifically, by adding `--watch` flags to everything we're using in prod (pm2, babel, and webpack) so they can also be used for dev to get hot code reloading.

pm2 does `--watch` as a background daemon that can be stopped, or checked for logs, as needed. Neither babel nor webpack have that option, and macOS does daemons very differently from how linux does, so to keep it working in both I'm suggesting we use `screen` to daemon-ize webpack and babel.

Outside hot code reloading and cross-platform, I'm not sure what else we need in a dev environment, so would also love to hear what I'm missing so far.


README.md
```diff
@@ -13,19 +13,21 @@ This is generated from [react-apollo-starter-kit](https://github.com/saikat/reac
...
+6. Start `rethinkdb` and load `localhost:8080` to confirm RethinkDB is properly installed.
+7. Run `./dev-tools/babel-run-with-env.js ./dev-tools/db-startup.js` to populate the tables. Check http://localhost:8080/#tables to confirm the tables were created.
+8. Create an [Auth0](auth0.com) account. In your Auth0 account, go to Settings -> Clients -> and then grab your Client ID, Client Secret, and your Auth0 domain (should look like xxx.auth0.com). Add those inside your `spoke.config.js` file (AUTH0_CLIENT_ID, AUTH0_CLIENT_SECRET, AUTH0_DOMAIN respectively).
+9. Add the login callback and logout callback URL in `spoke.config.js` (default `http://localhost:3000/login-callback` and `http://localhost:3000/logout-callback`) to your Auth0 app settings under "Allowed Callback URLs" and "Allowed Logout URLs" respectively.
+10. Start compiling client code with `screen -S webpack npm run dev-build-client`. Then use `ctrl-a, d` to exit the screen.
+11. Start compiling server code with `screen -S webpack npm run dev-build-server`. Then use `ctrl-a, d` to exit the screen.
+12. Start server processes with `pm2 start spoke-pm2.config.js --watch`.
```
shakalee14, line 24: should we include instructions for the logs flag too?

README.md
```diff
@@ -13,19 +13,21 @@ This is generated from [react-apollo-starter-kit](https://github.com/saikat/reac
...
+13. Go to `localhost:3000` to load the app.
+14. Because Spoke is invite-only you need to generate an invite:
```
shakalee14, line 26: This should be a separate p.r., but @sandramchung handled these steps during the hackathon so that instead of having to visit `http://localhost:8080/#dataexplorer` and doing an insert statement to generate an invite, you can hover your mouse below the text on the spoke page and see a red easter egg link appear that allows you to sign in.

dev-tools/Procfile.dev
```diff
@@ -1,10 +0,0 @@
-server: npm run dev-start
```
shakalee14, line 1: just checking - but because we did steps 6 & 7, rethinkdb is still running?

shakalee14 (commented):
Just a few comments and questions.

README.md
sreynen: You mean the date format? I never got that to work.

dev-tools/Procfile.dev
sreynen: Yes, rethink is running as a separate daemon process in this scenario, never stops.

shakalee14:
@sreynen do you think we will pick this PR up again?

PR #64, opened 2017-07-25
make answerOptions more knex-supportable, and will do the same thing

schuyler1d:
I believe this does the same thing, because you end up selecting just from the left 'table' anyway.
Knex complains because the tables are the same, and so the fields fail -- probably will have to deal with that in another query, but for this one....


sandramchung:
👍

PR #65, opened 2017-07-25
remove job queues/locking: if we do not do this, will have to rewrite…

schuyler1d:
… for knex support anyway.

rethink-knex-adapter doesn't implement `group()`, `ungroup()` nor the weird filter() meta-function thing (`filter()` for standard dict queries work).

I don't see a need for the job queues -- if someone does, perhaps this would be a good task to take on "the right way"

### DIFF
--- a/src/workers/lib.js
+++ b/src/workers/lib.js
@@ -13,35 +13,17 @@ export async function updateJob(job, percentComplete) {
 }
 
 export async function getNextJob() {
-  const lockedQueues = await r.table('job_request')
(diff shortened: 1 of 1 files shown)


sreynen:
Looking back at the git history on this file, the queue names were originally campaign-specific, so I'm guessing this was added to prevent one campaign from occupying all the job processing. That's maybe only a concern in a software-as-a-service model, which I think we no longer care much about now that it's open source, so I'm ++ on removing. We can always add it back if a problem emerges.

PR #67, opened 2017-07-26
move jobs to different file, so they can be imported without triggeri…

schuyler1d:
…ng job loop


sreynen (approved):

shakalee14 (approved):

shakalee14 (commented):
@schuyler1d 
Actually - do we need to also change references where job-handler.js exists?
like in package.json --> line 28 and 29

PR #68, opened 2017-07-26
WIP: Pool Number Refresh

sreynen:
This will be a script to refresh a number pool, removing numbers in no-longer-used area codes, and also replacing numbers that have newer alternatives in the same area code. The script will be optional, as not everyone using Spoke will have a need to rotate numbers, and can be set to whatever schedule is appropriate via Cron, e.g. orgs might want to rotate more often during higher usage periods to mitigate increased risk of numbers getting blacklisted.


src/workers/pool-number-refresh.js
```diff
@@ -0,0 +1,94 @@
+import Twilio from 'twilio'
+// import { r } from '../server/models'
+
+/**
+ * Get all numbers contacted in last 30 days.
+ */
+const getRecentContactNumbers = () => {
+  // TODO: replace this with database call after thinky to knex switch
```
shakalee14, line 8: could we also do a call to the twilio api?

src/workers/pool-number-refresh.js
```diff
@@ -0,0 +1,94 @@
...
+}, {})
+
+/**
+ * Get numbers for Twilio poool from API.
+ */
+const getCurrentTwilioPoolNumbers = (twilio, sid) => {
+  const service = twilio.messaging
+    .services(sid)
+    .fetch()
+  return service
```
shakalee14: when we return service, does that return an object that has available phone numbers in it?

sreynen: Maybe? This part doesn't currently work at all. It's what https://www.twilio.com/docs/api/rest/messaging-services says should work, but in Spoke code, `twilio.messaging` is undefined.

src/workers/pool-number-refresh.js
sreynen: Probably, but since we have the numbers in the database, that should be faster and more reliable.

sreynen:
Turns out the TODOs on using the Twilio API can't been done with the current API, so this is stalled until new API features arrive.

Issue #57, opened 2017-07-20
Login component


lady3bean:
Make login link more prominent

PR #69, opened 2017-07-27
Login - Adds light and dark background links to theme and cleans up old links

lady3bean:
For issue #57


shakalee14 (commented):
LGTM but I also paired.

src/components/Pricing.jsx
```diff
@@ -1,65 +0,0 @@
-import React from 'react'
```
sreynen, line 1: https://github.com/MoveOnOrg/Spoke/blob/master/src/routes.jsx#L23 imports this file. I'd expect that to throw an error if the file is deleted.

src/containers/Home.jsx
```diff
@@ -24,8 +24,8 @@ const styles = StyleSheet.create({
     marginBottom: 15,
     color: theme.colors.white
   },
-  link: {
-    ...theme.text.link
+  link_dark_bg: {
```
sreynen, line 27: I suggest naming link types by function rather than appearance, to leave room for both alternate themes and accessibility via non-visual styles. E.g. if we're using dark background to add emphasis, maybe something like `link_important`.

shakalee14: are you suggesting using ``link_important `` and ``link `` ?

sreynen: Maybe? I'm just guessing at the function of these styles, so I'm more generally suggesting the function drive the name.

shakalee14: ok - i'll see what @lady3bean thinks. this link isn't necessarily more important, but will defer to them for naming convention.

shakalee14 (approved):

PR #74, opened 2017-07-27
change merge() methods to use eqJoin instead where it is easy.  these…

schuyler1d:
… may work in rethinkdb, but not sure

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -96,7 +96,7 @@ export async function assignTexters(job) {
   const payload = JSON.parse(job.payload)
   const id = job.campaign_id
   const texters = payload.texters
-  const currentAssignments = await r.knex.table('assignment')
+  const currentAssignments = await r.knex.from('assignment')
     .where('campaign_id', id)
     .join('campaign_contact', 'id', 'assignment_id')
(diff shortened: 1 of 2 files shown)


src/workers/jobs.js
```diff
@@ -169,16 +168,15 @@ export async function assignTexters(job) {
...
-        .getAll(row('id'), { index: 'assignment_id' })
-        .count()
-    }))
-    .filter((row) => row('count').eq(0))
+  const assignmentsToDelete = await r.knex.from('assignment')
+    .where('campaign_id', id)
+    .join('campaign_contact', 'id', 'assignment_id')
+    .groupBy('assignment_id')
+    .select('assignment_id')
+    .havingRaw('COUNT(campaign_contact.id) = 0')
```
shakalee14, line 176: hard deleting is fine here right? we don't need to archive assignments?

shakalee14 (approved):

src/workers/jobs.js
schuyler1d: i'm just doing what the app did before.  If we want to archive assignments, then let's make that a github issue to prioritize, etc.

PR #76, opened 2017-07-27
WIP user phone not required

sreynen:
Everything that assumes user_number needs to be updated to work differently, because user_number won't exist on messages once we move to Twilio deciding the numbers via messaging service.


src/server/models/pending-message-part.js
```diff
@@ -14,7 +14,7 @@ const PendingMessagePart = thinky.createModel('pending_message_part', type.objec
   service_id: requiredString(),
   parent_id: optionalString(),
   service_message: requiredString(), //JSON
-  user_number: requiredString(),
+  user_number: optionalString(),
```
shakalee14, line 17: we're changing this optionalString() because we may not have the user_number now that we aren't deciding?

sreynen: Yep.

PR #79, opened 2017-07-31
user_number not required

sreynen:
This allows Spoke to work with a Twilio messaging service. In a messaging service, the sender number is chosen by Twilio rather than Spoke, so these changes should all make Spoke work in that context.


schuyler1d (approved):
:+1: looks good!

src/workers/message-sender-234.js
```diff
@@ -9,8 +9,8 @@ const serviceMap = { nexmo, twilio }
 async function sendMessages() {
   const messages = await r.table('message')
     .getAll('QUEUED', { index: 'send_status' })
-    .filter((doc) => doc('user_number').match('[234]$'))
-    .group('user_number')
+    .filter((doc) => doc('contact_number').match('[234]$'))
```
shakalee14, line 12: can you remind me again why we we're matching for 234 and 56 specifically?

src/workers/message-sender-789.js
```diff
@@ -9,8 +9,8 @@ const serviceMap = { nexmo, twilio }
 async function sendMessages() {
   const messages = await r.table('message')
     .getAll('QUEUED', { index: 'send_status' })
-    .filter((doc) => doc('user_number').match('[789]$'))
-    .group('user_number')
+    .filter((doc) => doc('contact_number').match('[789]$'))
```
shakalee14, line 12: See above comment about matching for these specific numbers - these arent area codes?

src/workers/message-sender-234.js
sreynen: As far as I can tell, the message-sender workers are just broken up into arbitrary groups to spread the load among a few workers, and the last digit in a phone number has relatively even distribution.

src/workers/message-sender-789.js
sreynen: No, it's the last digit of the phone number.

src/workers/message-sender-234.js
shakalee14: yep, it's just counting up! great, thanks.

shakalee14 (approved):

PR #82, opened 2017-07-31
Moar tests

mathemagica:
tests for 'should create a test campaign', 'should create campaign contacts', and some test helper fn refactoring


sandramchung:
👍

PR #89, opened 2017-08-02
README: Link to Auth0 with an absolute path

hiemanshu:
This fixes the broken link to Auth0 in the README. Not using the absolute path tries to open it as a relative file link.


schuyler1d:
@hiemanshu welcome!
If you are checking Spoke out, we are actively working on a migration over to knex with rdbms (away from rethinkdb).  You can see this on the `knex` branch.  Ping us on the Open slack channel to discuss!

Issue #86, opened 2017-08-01
Invite logic needs updating in knex


sreynen:
Currently the invite link is going to /invite/null because it's not getting the ID back on creation. But even if it does get the ID back, we need to use something other than the ID for the URL, because IDs in PostgreSQL are not obscure hashes like they are in RethinkDB, they're easily guessable numbers.

hiemanshu:
RethinkDB used UUIDs as the primary key, you can have postgresql (using pgcrypto extension) or sqlite (using a plugin) generate and use UUIDs as their primary key. Postgresql has a [uuid](https://www.postgresql.org/docs/9.1/static/datatype-uuid.html) type, Sqlite can use string or binary blob to store UUIDs (binary blobs are faster, but less human readable)

schuyler1d:
it looks like knex queries have uuid for each query, as well, so maybe we can use those to fill those in.  might be worth trying to add it in rethink-knex-adapter (but as a separate field, in most cases).  invite though could be a primary field.

sreynen:
FYI https://github.com/MoveOnOrg/Spoke/tree/wip-86-invite-links has work in progress on this.

sreynen:
This should be resolved now.

PR #98, opened 2017-08-03
Invite links update

sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/86, this adds a hash column to the invite schema, populates it with UUID, and uses that instead of ID, so invite links are less guessable.

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -1,4 +1,3 @@
-import { log } from '../../lib'
 import { Campaign,
   CannedResponse,
   Invite,
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -1,3 +1,4 @@
+import { log } from '../../lib'
 import { Campaign,
   CannedResponse,
   Invite,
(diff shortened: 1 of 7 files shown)


src/server/api/schema.js
```diff
@@ -1,3 +1,4 @@
+import { log } from '../../lib'
```
shakalee14: Why are we importing log here? Are we using it/planning to use it in this file?

shakalee14 (approved):

shakalee14:
I think we need to wrap the inviteInput in an array because of knex psql? Not sure if this is the same issue, but in order to do an [insert statement and return the inserted id for knex and (specifically) psql, what you want to return needs to be wrapped in an array](http://knexjs.org/#Builder-insert). But this comment could be completely un related.

shakalee14 (commented):
When testing on production, I'm getting a GraphQL error when trying to create an organization. After you created an invite connected to a hash, were you able to create an organization?

sreynen:
Org creation should be fixed now. In my testing, the org I created doesn't get attached to my account, but I confirmed the org was created in the database, and I believe that also happens in master, so assuming it's not related to invite changes.

Issue #115, opened 2017-08-11
canned responses dont save in admin dashboard


shakalee14:
- [ ] As an admin, I should be able to create a canned response and save it.

shakalee14:
Currently, as an admin, I can create and save a canned response(s). But when I click `next` to create the campaign I receive the following error in the client:
`There was an error with your request. Try again a little bit!`

Error on the server: `model.uuid..` is not a function possibly coming from `./src/server/api/schema.js`

PR #122, opened 2017-08-15
Creates random id for new canned response

shakalee14:


shakalee14:
Resolves issue https://github.com/MoveOnOrg/Spoke/issues/115

Issue #97, opened 2017-08-03
Review security of app


schuyler1d:
any admin-only views should be admin only.

- [x] make sure /allmessages is secure (might need to do something per-organization?)
- [x] make sure /availablephonenumbers is secure
- [ ] check graphql queries for security issues **p1&p2**
 - [x] twilio should verify from twilio: https://www.twilio.com/docs/guides/how-to-secure-your-express-app-by-validating-incoming-twilio-requests **p1**

Note: nexmo incoming messages should verify that they are from nexmo but not a priority right now to implement.
Note: Need to write backend test that catch authentication issues

sreynen:
Currently /admin/1/incoming gets /allmessages, which returns every incoming message sent to the Twilio account, regardless of which org the message was part of. Also currently an admin on any org can go to /admin/1/incoming, even if they're not an admin on org ID 1. While client-side org check would be nice, I think the important check for security is on the server (since clients can lie), so my plan is to change /allmessages to /allmessages/{orgID}, confirm the current user is admin on {orgID}, and only return messages from the local database that are attached to {orgID}.

sreynen:
I went through all the resolvers and mutations to see how they limit access, and here's what I found, with things that I think need tighter security in bold.

Mutations:

- sendReply: only works in development mode, no security concern
- exportCampaign: requires user is org admin
- editOrganizationRoles: requires user is org owner or admin 
- [x] **joinOrganization**: currently anyone can become a texter in any org **p1** -- see https://github.com/MoveOnOrg/Spoke/pull/123
- updateTextingHours: requires user is org owner
- updateTextingHoursEnforcement: requires user is org owner
- [x] **createInvite**: currently anyone can create an invite **p2** 
- createCampaign: requires user is org admin
- unarchiveCampaign: requires user is org admin
- archiveCampaign: requires user is org admin
- startCampaign: requires user is org admin
- editCampaign: requires user is org admin
- [x] **createCannedResponse**: currently anyone can create a canned response for any user **p1** @shakalee14
- createOrganization: requires valid inviteId
- [x] **editCampaignContactMessageStatus**: currently anyone can edit any message status **p1** -- needs assignment security check -- see https://github.com/MoveOnOrg/Spoke/pull/131
- [x] **createOptOut**: currently anyone can create an opt-out for any contact **p1** -- needs assignment security check @shakalee14
- [ ] **sendMessage**: currently anyone can send a message **p2** 
[...]

PR #123, opened 2017-08-15
joinOrganization security

sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/97 this adds UUIDs to organization model and changes the join links to use UUID, so join links are no longer guessable.

**Note:** this includes a migration with a schema update to run on deploy. Should not require wiping database.

### DIFF
--- a/src/migrations/2017-08-10-organization_uuid.js
+++ b/src/migrations/2017-08-10-organization_uuid.js
@@ -1,0 +1,10 @@
+import { r } from '../server/models'
+
+async function migrate() {
+  await r.knex.schema.alterTable('organization', (table) => {
+    table.string('uuid');
+  })
+  console.log('added uuid column to organization table')
+}
+
+migrate()
--- a/webpack/config.js
+++ b/webpack/config.js
@@ -17,7 +17,12 @@ if (!DEBUG) {
   plugins.push(new ManifestPlugin({
     fileName: assetMapFile
   }))
-  plugins.push(new webpack.optimize.UglifyJsPlugin({ minimize: true }))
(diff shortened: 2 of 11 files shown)


src/migrations/2017-08-10-organization_uuid.js
```diff
@@ -0,0 +1,10 @@
+import { r } from '../server/models'
+
+async function migrate() {
```
shakalee14, line 3: ok so the idea is to have a uuid and an id, but just not expose the id in the url?

webpack/config.js
```diff
@@ -17,7 +17,12 @@ if (!DEBUG) {
   plugins.push(new ManifestPlugin({
     fileName: assetMapFile
   }))
-  plugins.push(new webpack.optimize.UglifyJsPlugin({ minimize: true }))
+  plugins.push(new webpack.optimize.UglifyJsPlugin({
```
shakalee14, line 20: are the changes in this file specific to your p.r.?

src/migrations/2017-08-10-organization_uuid.js
sreynen: It doesn't really matter if the ID is exposed, it just can't be used to join the org because it's easily guessable. And joining with UUID instead requires getting the UUID from the organization admin to the texter and then back to the app. URL is the easiest way to transmit the UUID.

webpack/config.js
sreynen: Sort of. This PR required a lot of WebPack rebuilding, and these changes sped that up.

shakalee14 (approved):
I think the changes to the webpack config should be another p.r. but i'm not blocking it.

PR #130, opened 2017-08-16
aws lambda environment support

schuyler1d:
docs probably need more details, as we determine exactly what's needed (deploy is a bit environment sensitive), but this gets us most of the way there.

Will also probably need some more work on the jobs/processes (since none of the non-server environment will run)


shakalee14 (commented):
from what I understand, it looks fine to me.

PR #131, opened 2017-08-16
editCampaignContactMessageStatus security

sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/97, adds assignmentRequired() check, and uses it.


schuyler1d:
Should being a campaign admin or ?Manager (whatever the middle role is) also be ok?

shakalee14 (approved):

PR #134, opened 2017-08-17
Removes create org invite handler from login link

shakalee14:
- Removes handler for creating an org invite from home page


sandramchung:
LGTM. We can also delete the handler code - easy enough to find it and reinstate it later as needed.

shakalee14:
created a way to reinstate via global variable @sandramchung

sandramchung:
👍

PR #138, opened 2017-08-18
#97 contacts security

sandramchung:
Per #97, secures contacts endpoint so that only campaign owners, admins, and texters who are assigned to the contact can load a contact.

### DIFF
--- a/src/server/api/errors.js
+++ b/src/server/api/errors.js
@@ -19,40 +19,18 @@ export function authRequired(user) {
   }
 }
 
-async function hasRole(userId, orgId, role) {
+export async function hasRole(userId, orgId, role) {  
   if (role) {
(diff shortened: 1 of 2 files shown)


schuyler1d (approved):
I like the abstraction of hasRole, however in your actual use of it, you just get all the roles as booleans.  DB-wise, this would be much faster with a single db call rather than 3, so why not just `r.knex('user_organization').where({user_id: userId, organization_id: orgId}).select('role')` and then map the results accordingly.

On a higher level, how many of our queries need to test based on role?  It seems like it might make sense to try saving the campaign roles of the user in the session somehow so we can test against that rather than for every query do all these permission calls again and again.

All that said.  this will work!  Yei better security!

shakalee14 (approved):

PR #140, opened 2017-08-18
Assignments security

sandramchung:
In testing this tiny change, I realized that I don't actually know which roles should have access to assignments. I made some assumptions about contacts. But in looking at the user_organization table in the db, the roles are set per organization, not per campaign like I assumed, and I don't know what their UX is supposed to look like. So marking this as WIP until we get some clarity.


mathemagica:
As an organizational admin, I should be able to assign contacts to campaigns. As a texter, I should not be allowed to view assignment screens or assign contacts to campaigns.

src/server/api/schema.js
```diff
@@ -690,12 +708,16 @@ const rootResolvers = {
       userRoles.forEach(role => {
         roles[role['role']] = 1
       })
+<<<<<<< HEAD
```
shakalee14: some merging issues here.

shakalee14 (approved):

shakalee14:
@sandramchung can you handle this conflict?

sandramchung:
@shakalee14 done

PR #143, opened 2017-08-21
remove mailgun api and just use a normal smtp mail server

schuyler1d:


src/network/response-middleware-network-interface.js
```diff
@@ -1,4 +1,5 @@
 import { createNetworkInterface } from 'apollo-client'
+import fetch from 'isomorphic-fetch'
```
shakalee14, line 2: to fix the fetch is not defined error?

shakalee14 (approved):

src/network/response-middleware-network-interface.js
schuyler1d: yes

PR #146, opened 2017-08-22
completes the process model for aws lambda and JOBS_SAME_PROCESS

### LINKED ISSUE #32: Speed up dev start time
`npm run dev` takes a long time (20 seconds for some people, over a minute for others) to get to a working server, which slows down development significantly for things that require frequent restarts. Let's see if we can speed that up.

### LINKED ISSUE #141: sent message creates a second record on twilio-message-report (rather than updating)
It seems like this happens because the first does not create a service_id value, and the update is only done through that.

schuyler1d:
fixes #141 
also pretty much fixes #32 

And then this should be it for lambda!


src/server/api/lib/twilio.js
```diff
@@ -5,8 +5,6 @@ import { log } from '../../../lib'
 import { getLastMessage } from './message-sending'
 import faker from 'faker'
 
-const defaultService = process.env.DEFAULT_SERVICE
```
shakalee14, line 8: why aren't we storing this in a variable anymore? just curious.

shakalee14 (approved):
we've added a few console.log's here to help someone see the job process?

src/server/api/lib/twilio.js
schuyler1d: This is the twilio file, so it should never be 'defaultService' -- it's 'twilio'.  The nexmo should probably have 'nexmo' hard-coded in the same way.  the DEFAULT_SERVICE environment variable is for the other parts of the app that need to decide which service to delegate to.

schuyler1d:
yeah, the console.logs will only be one-lines occasionally during the ping.  we can probably remove them, but log-wise, it seems a low lift at least while we're still working to understand the app.

Issue #133, opened 2017-08-17
RSVP Syncing with Action Kit


shakalee14:
- [ ] User rsvp for an event in spoke should sync with Action Kit

mathemagica:
- [ ]  write a query that gives us a csv of contacts + page_id, event_id, user_id (ak id) from data warehouse (or make sure your test data includes these fields for your initial contact list)
See issue --> https://github.com/MoveOnOrg/Spoke/issues/176

- the admin imports that csv to spoke on campaign creation 

- [ ] in ${handler}.js load metadata associated with the contact (event_id, user_id, page_id)

- [ ] In ${handler}.js try making an http call that includes AK_USER, AK_PASSWORD, AK_BASEURL like https://github.com/MoveOnOrg/python-actionkit/blob/public-master/actionkit/api/base.py

- [x] 3.5 store AK_USER, AK_PASSWORD, AK_BASEURL in .env (and eventually the config we deploy prod with too)

- [ ] report the http post to create the event signup like https://github.com/MoveOnOrg/python-actionkit/blob/public-master/actionkit/api/event.py

shakalee14:
lets split the last bullet point out.

shakalee14:
Step 1, we want a query to return an actual csv? thats downloaded to your computer, and then we import that csv right back into spoke? 
(question about step 1 and 2). and the csv is campaign specific that includes info on folks who RSVP yes (so could just be akids and the event_id). What is a page_id - thats ak related? is that info that needs to be initially included in the contacts import?

And then would the user story for these 2 steps be, I as a owner, can click a button that downloads a csv of rsvp folks, and then click on an upload button and upload the csv I just downloaded?

shakalee14:
- Custom fields is broken --> currently cant successfully use them if added which blocks implementation of this issue.  

Issue tracked here: https://github.com/MoveOnOrg/Spoke/issues/183

PR #149, opened 2017-08-23
General action framework

sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/133, this doesn't yet include any of the actual interaction with ActionKit, just a general framework for associating actions with responses. Mostly just adds the concept of an action to the many different places that deal with interaction steps.

### DIFF
--- a/src/server/action_handlers/actionkit-rsvp.js
+++ b/src/server/action_handlers/actionkit-rsvp.js
@@ -1,0 +1,9 @@
+export const displayName = () => 'ActionKit Event RSVP'
+
+export const available = () => {
+  return true
+}
+
+export const processAction = (questionResponse) => {
+
+}
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -196,7 +196,6 @@ const rootSchema = `
 
 async function editCampaign(id, campaign, loaders) {
   const { title, description, dueBy, organizationId } = campaign
-  log.info(campaign, 'campaign')
   const campaignUpdates = {
     id,
     title,
@@ -255,7 +254,6 @@ async function editCampaign(id, campaign, loaders) {
     }
   }
   if (campaign.hasOwnProperty('interactionSteps')) {
-    log.info(campaign.interactionSteps, 'campaign.interactionSteps')
     let job = await JobRequest.save({
       queue_name: `${id}:edit_campaign`,
       locks_queue: true,
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -695,7 +695,12 @@ const rootResolvers = {
     availableActions: (_, __, { user }) => {
       const allHandlers = process.env.ACTION_HANDLERS.split(',')
       const availableHandlers = allHandlers.filter(handler => {
-        return require(`../action_handlers/${handler}.js`).available()
+        try {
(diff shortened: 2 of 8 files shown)


src/server/action_handlers/actionkit-rsvp.js
```diff
@@ -0,0 +1,9 @@
+export const displayName = () => 'ActionKit Event RSVP'
+
+export const available = () => {
+  return true
```
schuyler1d, line 4: I know this is a stub, but at least for prod before it's deployed this should probably be `!!process.env.ACTIONKIT_BASEURL` or something

src/server/action_handlers/actionkit-rsvp.js
```diff
@@ -0,0 +1,9 @@
...
+}
+
+export const processAction = (questionResponse) => {
```
schuyler1d, line 7: I'm going to suggest you pass in the campaign_contact object, and the campaign as well here (I assume it's already in the context sending out to it, so might as well send it in here for when it's useful.

src/server/api/schema.js
```diff
@@ -685,6 +693,20 @@ const rootResolvers = {
     organizations: async(_, { id }, { user }) => {
       await superAdminRequired(user)
       return r.table('organization')
+    },
+    availableActions: (_, __, { user }) => {
+      const allHandlers = process.env.ACTION_HANDLERS.split(',')
+      const availableHandlers = allHandlers.filter(handler => {
+        return require(`../action_handlers/${handler}.js`).available()
```
schuyler1d: might want to try..catch here, if there's a bad handler name or something.

src/server/api/schema.js
```diff
@@ -685,6 +693,20 @@ const rootResolvers = {
...
+      })
+      const availableHandlerObjects = availableHandlers.map(handler => {
+        const handlerPath = `../action_handlers/${handler}.js`
+        return {
+          'name': handler,
+          'display_name': require(handlerPath).displayName()
```
schuyler1d, line 709: ditto on try-...catching or just have a whitelist hardcoded of the existing ones.

schuyler1d (approved):
This is an awesome example that I hope we keep around.  Please add a link to this PR in some docs/ADDING_FEATURES.md or something.  I will be copy-pasting from this for all my spoke work
:100:

shakalee14 (approved):

src/server/api/schema.js
sreynen: Added.

src/server/api/schema.js
sreynen: I returned false on the previous catch, so when it gets here, everything in availableHandlers must have a matching file.

src/server/action_handlers/actionkit-rsvp.js
sreynen: Yeah, waiting to see what environment variables are needed in processAction to check here.

src/server/action_handlers/actionkit-rsvp.js
sreynen: I'm not sure where exactly this will be called from, but makes sense to pass in all available context that might be useful.

PR #151, opened 2017-08-23
add external_id and backend naive-implementation for contact_sql upload

schuyler1d:
this is a WIP to data from a data warehouse.  We will need to expose the frontend SQL form.  However, I think we should merge this in early because:
1. it adds external_id to the campaign_contact table which will probably be useful for Scott's actionkit_event connector
2. the backend does work, and does no harm.

### DIFF
--- a/src/components/CampaignContactsForm.jsx
+++ b/src/components/CampaignContactsForm.jsx
@@ -173,7 +173,6 @@ export default class CampaignContactsForm extends React.Component {
 
   renderForm() {
     const { contactUploadError } = this.state
-
     return (
       <div>
         <GSForm
@@ -237,6 +236,7 @@ export default class CampaignContactsForm extends React.Component {
 }
 
 CampaignContactsForm.propTypes = {
+  datawarehouseAvailable: type.bool,
   onChange: type.func,
   optOuts: type.array,
   formValues: type.object,
--- a/src/components/CampaignContactsForm.jsx
+++ b/src/components/CampaignContactsForm.jsx
@@ -209,8 +209,10 @@ export default class CampaignContactsForm extends React.Component {
         Your upload file should be in CSV format with column headings in
         the first row. You must include <span className={css(styles.csvHeader)}>firstName</span>,
         <span className={css(styles.csvHeader)}>lastName</span>, and
-        <span className={css(styles.csvHeader)}>cell</span> columns. If you include a <span className={css(styles.csvHeader)}>zip</span> column,
+        <span className={css(styles.csvHeader)}>cell</span> columns.
+        If you include a <span className={css(styles.csvHeader)}>zip</span> column,
         we'll use the zip to guess the contact's timezone for enforcing texting hours.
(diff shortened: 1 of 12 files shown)


src/components/CampaignContactsForm.jsx
```diff
@@ -209,8 +209,10 @@ export default class CampaignContactsForm extends React.Component {
         Your upload file should be in CSV format with column headings in
         the first row. You must include <span className={css(styles.csvHeader)}>firstName</span>,
         <span className={css(styles.csvHeader)}>lastName</span>, and
-        <span className={css(styles.csvHeader)}>cell</span> columns. If you include a <span className={css(styles.csvHeader)}>zip</span> column,
+        <span className={css(styles.csvHeader)}>cell</span> columns.
+        If you include a <span className={css(styles.csvHeader)}>zip</span> column,
         we'll use the zip to guess the contact's timezone for enforcing texting hours.
+        An optional column to map the contact to a CRM is <span className={css(styles.csvHeader)}>external_id</span>
```
shakalee14, line 214: ok cool! so we would need to add this column.

schuyler1d: to the csv (and in the future as a column in the DB sql), yes, but then we can update/sync/send actions based on it

src/containers/AdminCampaignEdit.jsx
```diff
@@ -145,7 +145,8 @@ class AdminCampaignEdit extends React.Component {
             cell: contact.cell,
             firstName: contact.firstName,
             lastName: contact.lastName,
-            zip: contact.zip
+            zip: contact.zip,
+            external_id: contact.zip
```
shakalee14, line 151: why is this contact.zip too?

src/workers/jobs.js
```diff
@@ -60,6 +64,62 @@ export async function uploadContacts(job) {
   }
 }
 
+
```
shakalee14, line 67: so we will have an input box, and then this function will pass it through line 79 as long as there is a datawarehouse connection, it starts with a select and doesnt start with a ;, cool!

src/workers/jobs.js
```diff
@@ -60,6 +64,62 @@ export async function uploadContacts(job) {
...
+  if (!datawarehouse) {
+    log.error('No data warehouse connection, so cannot load contacts', job)
+    return
+  }
+
+  knexResult = await datawarehouse.raw(sql_query)
+  let fields = {}
+  let customFields = {}
+  const contactFields = {
+    first_name: 1,
```
shakalee14, line 83: these 1s here are for `true`?

schuyler1d: the value doesn't matter.  I just want to test whether it's in there.  I don't think es6 has sets, but i bet if i google it, they will.

src/containers/AdminCampaignEdit.jsx
schuyler1d: the zip line is just modified because of adding the comma

PR #152, opened 2017-08-23
Adds check for action handler before performing available action query

shakalee14:


src/server/api/schema.js
```diff
@@ -737,23 +737,27 @@ const rootResolvers = {
...
-      const availableHandlerObjects = availableHandlers.map(handler => {
-        const handlerPath = `../action_handlers/${handler}.js`
-        return {
-          'name': handler,
-          'display_name': require(handlerPath).displayName()
-        }
-      })
-      return availableHandlerObjects
+      if(!process.env.ACTION_HANDLERS){
+        return
```
schuyler1d: Does this work? You might need to send an empty array

PR #153, opened 2017-08-23
correct migration run command

sandramchung:
In case you still need to run the migration that updates interaction_step


sandramchung:
No need to merge this now that the new migration auto-run system is in place.

PR #154, opened 2017-08-24
Datawarehouse

schuyler1d:
This adds two features:
1. A frontend to the SQL interface to pull in campaign contacts from the data warehouse
2. A migrations regime that will have our dbs automatically run on startup (after pulling)


src/migrations/index.js
```diff
@@ -0,0 +1,71 @@
+import { r, Migrations } from '../server/models'
```
shakalee14, line 1: i like this better than having separate migration files.

src/server/index.js
```diff
@@ -33,6 +34,9 @@ setupAuth0Passport()
 if (!process.env.SUPPRESS_SEED_CALLS) {
   seedZipCodes()
 }
+if (!process.env.SUPPRESS_MIGRATIONS) {
```
shakalee14, line 37: did we add this to the .env.template file?

shakalee14 (approved):

src/migrations/index.js
sandramchung: ++

src/server/index.js
schuyler1d: no because I don't think it should be the default.  We could add it to docs somewhere, I suppose

PR #159, opened 2017-08-25
migration issues: yei! off-by-one errors

schuyler1d:
also depends on: https://github.com/MoveOnOrg/rethink-knex-adapter/pull/14


sandramchung (approved):
👍

Issue #158, opened 2017-08-25
[1 day] Bug: in campaign edit admin in the interaction section, answer options don't always save as expected


sandramchung:
Since actions were added, handleSubmit and handleSave() don't fire when the "next" ("save" if you are editing a previously created campaign) button in the interaction section is clicked and there is answer options content in the interaction script. The expected behavior is that clicking that button saves the content in the form section and advances you to the next section in the form, and that when you navigate back to the interaction section the content you just added has been saved. The next/save button does most of this when there are no answer options in the interactions form section content. Reloading the entire form will sometimes successfully save the changes to the interactions content. See debug-interactions-form branch for work in progress.

sandramchung:
Upon closer inspection, there are several bugs in the interactions form, and this one is observable intermittently: https://github.com/MoveOnOrg/Spoke/issues/166  probably has to do with polling and the delegation of interaction step saving to an asynchronous job.

PR #169, opened 2017-08-29
Debug interactions form

sandramchung:
#158 Finally tracked down a silent form validation failure. A few React bug/UX fixes along the way.

### DIFF
--- a/src/containers/AdminCampaignEdit.jsx
+++ b/src/containers/AdminCampaignEdit.jsx
@@ -124,6 +124,7 @@ class AdminCampaignEdit extends React.Component {
   }
 
   handleSave = async () => {
+    console.log("handleSave")
     let saveObject = {}
     this.sections().forEach((section) => {
       if (!this.checkSectionSaved(section)) {
@@ -133,6 +134,7 @@ class AdminCampaignEdit extends React.Component {
         }
       }
     })
+    debugger
     if (Object.keys(saveObject).length > 0) {
       // Transform the campaign into an input understood by the server
       const newCampaign = {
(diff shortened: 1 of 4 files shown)


src/containers/AdminCampaignEdit.jsx
```diff
@@ -123,6 +123,15 @@ class AdminCampaignEdit extends React.Component {
     })
   }
 
+  handleSubmit = async () => {
+    await this.handleSave()
+    this.setState({
+      expandedSection: this.state.expandedSection >= this.sections().length - 1 ||
+        !this.isNew() ?
+          null : this.state.expandedSection + 1
+    }) // currently throws an unmounted component error in the console
```
sreynen, line 132: Possibly related to the unmounted component error, when setting state based on previous state, React docs recommend using the updater method: https://facebook.github.io/react/docs/react-component.html#setstate which I think would look something like this:

`this.setState((previousState) => {
      expandedSection: previousState.expandedSection >= this.sections().length - 1 ||
        !this.isNew() ?
          null : previousState.expandedSection + 1
    })`

sreynen (approved):
One suggested change, but not sure if it matters, and doesn't need a re-review either way.

src/containers/AdminCampaignEdit.jsx
sandramchung: hmmm. This code works, but it still throws an unmounted component error:
```
this.setState((prevState) => ({
      expandedSection: prevState.expandedSection >= this.sections().length - 1 ||
        !this.isNew() ?
          null : prevState.expandedSection + 1
    }))
```

PR #173, opened 2017-08-30
added user object to prevent 500 error on texter todo list after message send

sandramchung:
Per today's errors during texting trial


mathemagica:
👍

PR #186, opened 2017-09-06
Add database pool size support for postgres

hiemanshu:
This change allows settings a pool size for Postgres.


shakalee14 (approved):

src/server/models/thinky.js
```diff
@@ -23,7 +23,8 @@ if (process.env.DB_JSON) {
       database: process.env.DB_NAME,
       password: process.env.DB_PASSWORD,
       user: process.env.DB_USER
-    }
+    },
+    pool: {min: process.env.DB_MIN_POOL, max: process.env.DB_MAX_POOL}
```
sandramchung: Per http://justonepixel.com/thinky/documentation/api/thinky/ the option is `buffer`, not `min`. And the defaults are 50 and 1000.

sreynen: Also, let's give these defaults in code, e.g. `process.env.DB_MIN_POOL || 50`, so a code update alone without manual config updates wouldn't break an existing instance.

hiemanshu: @sandramchung We don't use thinky for this anymore, but use knex. Knex documentation defines pool, and I've tested this on production and it works well (the pool size changes). 

http://knexjs.org/#Installation-pooling

sandramchung: @hiemanshu  Great. We could still use some defaults in the code so that Spoke instances without these variables set don't break.

hiemanshu: Yeah, I have the defaults defined in the env template, I'll update the actual config then.

sreynen: Yeah, our goal is for everyone with installs to be able to deploy code changes without worrying about breaking anything, and only need to update environment when they want to change something. At least until we have a better process for flagging changes that might break things.

sreynen (changes requested):
See inline comments.

hiemanshu:
I have updated the PR and fixed the conflicts.

PR #188, opened 2017-09-06
Allow settings nodemailer's secure settings

hiemanshu:
This allows testing locally with something like [MailCatcher](https://mailcatcher.me/)


sreynen (approved):
In retrospect, this probably should have been false by default, since most SMTP servers (including our own, it turns out) don't use SSL connections. Config is a good way to undo that mistake without breaking existing sites where it should be true.

PR #189, opened 2017-09-06
Invalidate session if deserialization fails

hiemanshu:
During testing I saw a deserialization failed error popup more than once. Invalidating the session will require the user to re-login and not have to see a 500 error.


shakalee14 (approved):

sandramchung:
Deployed to stage and tested. Yay!

PR #190, opened 2017-09-06
Remove tunnel being used on production

hiemanshu:
I don't see this being used anywhere, and haven't found a use for it.


shakalee14 (approved):

sandramchung:
Let's also remove the tunnel.py file, since we're removing references to it and no longer running it.

sandramchung:
Deployed and tested on stage.

PR #193, opened 2017-09-07
Bug custom fields script

shakalee14:


shakalee14:
- how to test this --> 
- create a campaign and upload contacts csv with custom fields
- create a script including custom fields
- assign a texter to the campaign and then switch to the texter view to see fields correctly populate

sandramchung:
Tested branch as described above and it works.  👍

src/lib/scripts.js
```diff
@@ -24,7 +24,8 @@ const getScriptFieldValue = (contact, texter, fieldName) => {
   } else if (TOP_LEVEL_UPLOAD_FIELDS.indexOf(fieldName) !== -1) {
     result = contact[fieldName]
   } else {
-    result = contact.customFields[fieldName]
+    let customFieldNames = JSON.parse(contact.customFields)
```
sandramchung: Nitpick: I think this this could be a const instead of let.

sandramchung:
LGTM

Issue #184, opened 2017-09-06
Diagnose how assigning texters currently works (or doesn't)


sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/177

shakalee14:
**How it doesnt work** 

-When I'm texting, the bubble for replies doesn't update after you handle your replies. Example, I as a texter send 2/4 messages, and get 2 replies back. The number in the replies bubble updates with the first replies I get. Then when I actually reply to the appropriate contacts, the number in the replies bubble does not update (doesn't recalculate after handling addressing replies) in the texter campaign dashboard. But in the actual texter to do list, it only shows me the texting conversations I need to reply to (so that part is not broken).

See **1 of 1 in the right corner** (meaning I only have 1 person to reply to)

<img width="1440" alt="screen shot 2017-09-08 at 9 42 35 am" src="https://user-images.githubusercontent.com/16676323/30214553-ea3e6d3a-947a-11e7-86fb-4eb911e4cea1.png">

But in the dashboard it says I have **2 replies** to send
<img width="1422" alt="screen shot 2017-09-08 at 9 48 31 am" src="https://user-images.githubusercontent.com/16676323/30214646-34cddd90-947b-11e7-9d7e-6ce0bedbdb06.png">

shakalee14:
I think this issue is two fold
- We currently attach auto assign to state and its automatically true when you visit this panel. So if you revisit this screen it'll try to auto-split contacts amongst texters
- Texting assignments are more stable after making sure information is saved. You can ensure it has saved after you press next and do a hard refresh. 

- Note - if you don't assign all the uploaded contacts to texters (say you upload 5 and only assign 3, then the panel for texters still stays yellow - which is misleading because you can still start the campaign)

shakalee14:
You can test this theory by checking out this PR:
https://github.com/MoveOnOrg/Spoke/pull/195

hiemanshu:
I'm currently looking at how texter assignments work. I'm doing this mainly to add a "Add All" button to the screen, so it adds all texters from that org onto the assignments screen. I'm getting this info from the `user_organization` table

shakalee14:
Could you make an issue @hiemanshu for what you're working on so we can see whats coming down the pipeline? That feature sounds awesome! If you want to see what we're working on, we've created milestones for the project.

hiemanshu:
[...]

PR #195, opened 2017-09-08
Changes autoSplit from true to false in campaigntexters form state

shakalee14:


shakalee14:
- To test this branch (which addresses https://github.com/MoveOnOrg/Spoke/issues/184) use the following instructions

- Check out branch
- Open admin and texter views
- Start creating a campaign, upload contacts
- Make sure autosplit is automatically off when you visit Texters panel (it should be off)
- Manually assign texters assignments. After you finish assigning contacts to texters, visit the next panel and refresh the page. Do this a few times (going back and forth between texters and script panels) to check if numbers persist after editing contact assignments.
- Start campaign 
- Check to see if texter assignment in dashboard is correct
- Go to admin panel and edit campaign - then visit texters panel again and try to reassign (visit next panel, refresh) and see if numbers persist. 
- Visit texter dashboard and see if number updates with new assignment

sandramchung:
Autosplit definitely works more coherently when it's off by default. It might make even more functional sense as a button that applies autosplit when it's clicked rather than as a toggle that enforces autosplit as you go. I tested with only six contacts and three texters, and I didn't test the case where some of the texters had already texted some contacts - so I'm not sure how functional that case is. Per Scott's comment on #184, it's hard to say whether or not this change address all the issues; but it is definitely an improvement.

sandramchung (approved):
👍

PR #196, opened 2017-09-11
absolute url from ENV makes apollo happy again

sandramchung:
Fixes the error `UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: ###) Error: Network error: only absolute urls are supported`  that we get on some graphQL requests, ostensibly since we added isomorphic-fetch to the repo.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,7 @@ AUTH0_DOMAIN=XXXX.auth0.com
 AUTH0_CLIENT_ID=
 AUTH0_CLIENT_SECRET=
 BASE_URL=http://localhost:3000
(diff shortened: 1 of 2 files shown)


shakalee14:
@anasauce see pr - this should fix that horrible error we were seeing friday

schuyler1d:
what does GRAPHQL_URL need to be in production?
Happy to +1 this to test on stage, but think it's important to see it working there.

schuyler1d (approved):

sandramchung:
Flagging that this will break people's locals when they merge it, until they set GRAPHQL_URL in .env

Issue #162, opened 2017-08-28
Make the volunteer texting screen auto-refresh after completing a round of texting


mathemagica:
When volunteers get responses, they should see those responses in the text screen automatically, without needing to reload the page.

pollingInterval: 5000

sandramchung:
Setting the pollInterval in TexterToDoList graphql query to 5000 makes the screen refresh quickly; but I'm concerned that lots of texters leaving this screen open means they're all hitting the database every five seconds. I think we should implement an event-based refresh instead - triggered when new assignments come in or when this texter sends or receives messages.

sandramchung:
Branch 162-textertodo-autorefresh and https://github.com/MoveOnOrg/Spoke/pull/197

sreynen:
Let's do both, do the fix we can do now, and then plan the better fix for later.

PR #197, opened 2017-09-11
Took down the polling interval on TexterToDo list from 10 seconds to 5 seconds

sandramchung:
This is a quick fix for #162, with the caveat that we may need to revisit it if it makes the DB cack. Tested on my local by starting a new test campaign with myself as a contact and texter, loading my texter todo list (/app/[userid]/todos), sending myself a message, and then replying to myself and checking to see that the number of replies needed in the new test campaign updates appropriately.


schuyler1d (approved):
:+1:

PR #204, opened 2017-09-12
Don't defaul to true when EMAIL_HOST_SECURE is false

sreynen:
I believe this was the goal with https://github.com/MoveOnOrg/Spoke/pull/188


sandramchung (approved):
👍

PR #219, opened 2017-09-13
fixes #216: closing context should return to the right organization /…

### LINKED ISSUE #216: Closing texter mode should go directly to /app/:organizationId rather than /app
I'm in an organization of my own ("scooby doo", orgid=4). When I go to 'switch to texter' (path=/app/4/todos) and then enter, then I click the 'X', it redirects me to path=/app which in turn redirects me to path=/app/3/todos (another org I'm a member of)

So this is a bug, but will affect few people.

schuyler1d:
…app even with multiple organizations

needed to do something small while getting back into the code base.


sandramchung:
👍

sandramchung (approved):
LGTM

PR #221, opened 2017-09-13
add travis status to readme

sreynen:


schuyler1d (approved):
might be good to make sure it builds and passes, before merging.  Right now the travis build fails at 'rake not installed' or something.

PR #223, opened 2017-09-14
194 new texter reply screen

shakalee14:
- only WIP because I haven't provided testing notes yet.

### DIFF
--- a/src/containers/AssignmentTexterContact.jsx
+++ b/src/containers/AssignmentTexterContact.jsx
@@ -27,17 +27,15 @@ import { withRouter } from 'react-router'
 import wrapMutations from './hoc/wrap-mutations'
 const styles = StyleSheet.create({
   mobile: {
-    '@media(min-width: 600px)':{
+    '@media(min-width: 425px)':{
       display: 'none !important'
     }
   },
-
   desktop: {
-    '@media(max-width: 600px)':{
+    '@media(max-width: 450px)':{
       display: 'none !important'
     }
   },
-
   container: {
     margin: 0,
     position: 'absolute',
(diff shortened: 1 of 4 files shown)


src/containers/AssignmentTexterContact.jsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react'
+import $ from 'jquery'
```
schuyler1d: Can we import it as `jQuery` instead of `$` -- it's really hard to see where it's used.

src/containers/AssignmentTexterContact.jsx
```diff
@@ -482,43 +527,84 @@ class AssignmentTexterContact extends React.Component {
...
-          {this.renderNeedsResponseToggleButton(contact)}
-          <RaisedButton
-            label='Canned responses'
-            onTouchTap={this.handleOpenPopover}
-          />
-          <RaisedButton
+    const size = $(window).width()
+    let toolBar = null
+
+   if (messageStatus === 'needsResponse' &&  size < 450 || messageStatus === 'messaged' && size < 450 ){
```
schuyler1d: could we achieve this with css breakpoints or max/min-width?  if too much a pain, then this is fine.

schuyler1d (approved):

src/containers/AssignmentTexterContact.jsx
sreynen: If I'm reading this right, it's only used in one place, `$(window).width()`. I'd like to push to not add jQuery at all, and find a way to do this with native JS, as jQuery is a pretty big package.

shakalee14:
In order to test, you must have replies working or a reply history in your texter contact messages
- Check out branch
- Create a campaign and assign yourself or texter contacts
- Make sure your campaign has a script longer than 1 interaction
- Switch to texter
- Change to responsive mode (make sure you refresh)
- Send a message (screen should look the same as on production since these changes should only affect mobile reply screen)
- Receive a reply
- Go to texter dashboard and click on `send replies`
- Screen should look like mock (except with a slightly different icon. Original icon in mock was created in google draw. Arrow icon used is from material-ui svg icons).
- Also check to see if regressions occur on desktop or ipad size screens for replies (make sure to refresh browser)

Issue #208, opened 2017-09-12
Opt-out reply cursor is buggy


schuyler1d:
Use case: want to say "Oh, no problem"   -- 
Experience: other letters start typing on the other side of of the cursor/built in reply.  cursor is a problem

Reproducing it:
1. click opt-out on a user
2. place your cursor with the mouse at the beginning of the text
3. start typing
Result: After typing a character the cursor moves to the end buggily

sreynen:
Update: this seems to be an issue with material-ui and text inputs, documented here: https://github.com/callemall/material-ui/pull/7006

Updating material-ui to a version including that PR doesn't directly fix it, I think because the fix in that PR simple stops material-ui from causing a re-render, and Spoke code is causing a re-render in other ways.

In looking at other dialogs in Spoke and how they've avoided this problem, I've only found one other dialog that has a simple text input (vs. something that looks like a text input, but is actually more complex, with {token} strings, a select list, etc.): the "Invite new texters" dialog on the admin people page. That dialog has the same bug, but it doesn't matter in that context because the text input isn't being used for input, only copying a URL.

sreynen:
After more testing with the latest stable version of material-ui (0.19.1), I haven't found a configuration that fixes this problem short of removing the dialog entirely, so I'm going to work on a non-dialog version of the opt-out interface.

PR #227, opened 2017-09-14
Move opt-out form from dialog to card

sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/208. This is a work-around to an issue with text inputs in material-ui dialogs. The work-around is to take the opt-out form out of a dialog and instead make it appear in a "card" below the opt-out button. Here's what it looks like after pressing the button:

screen shot 2017-09-14 at 2 58 00 pm

That hides a lot on mobile sizes, but only while it's being used, which seems okay. Pressing "cancel" makes it go away.

### DIFF
--- a/src/containers/AssignmentTexterContact.jsx
+++ b/src/containers/AssignmentTexterContact.jsx
@@ -12,9 +12,6 @@ import IconButton from 'material-ui/IconButton/IconButton'
 import { Toolbar, ToolbarGroup, ToolbarSeparator } from 'material-ui/Toolbar'
 import { Card, CardActions, CardTitle } from 'material-ui/Card'
 import Divider from 'material-ui/Divider'
-
-
-import Dialog from 'material-ui/Dialog'
 import { applyScript } from '../lib/scripts'
 import gql from 'graphql-tag'
 import loadData from './hoc/load-data'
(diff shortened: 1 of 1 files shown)


src/containers/AssignmentTexterContact.jsx
```diff
@@ -10,6 +10,10 @@ import NavigateCloseIcon from 'material-ui/svg-icons/navigation/close'
 import { grey100 } from 'material-ui/styles/colors'
 import IconButton from 'material-ui/IconButton/IconButton'
 import { Toolbar, ToolbarGroup, ToolbarSeparator } from 'material-ui/Toolbar'
+import { Card, CardActions, CardTitle } from 'material-ui/Card'
+import Divider from 'material-ui/Divider'
+
+
 import Dialog from 'material-ui/Dialog'
```
schuyler1d: do you need this import anymore?

schuyler1d (approved):
:+1: 
So cards work (are less buggy) than dialogs?

src/containers/AssignmentTexterContact.jsx
sreynen: Nope, don't need it.

Issue #233, opened 2017-09-19
Opt-out button with mobile view overlays other buttons and is unavailable on iPad


schuyler1d:
![bugoptoutmobile](https://user-images.githubusercontent.com/52117/30600057-67f45884-9d2c-11e7-84df-b75a5260eb00.png)

shakalee14:
<img width="358" alt="screen shot 2017-09-19 at 11 26 40 am" src="https://user-images.githubusercontent.com/16676323/30602369-6212247c-9d32-11e7-9783-5a5c922fd01c.png">

PR: https://github.com/MoveOnOrg/Spoke/pull/236

PR #236, opened 2017-09-19
Adds white background and zIndex to fix optout overlay bug on mobile

shakalee14:


shakalee14:
Need to test.

schuyler1d (approved):
Does this also resolve the confusing 'double send button' visible issue?

shakalee14:
yes - on mobile, you will only see one send button
screen shot of what it looks like here --> https://github.com/MoveOnOrg/Spoke/issues/233

PR #243, opened 2017-09-20
Heroku one click deploy

shakalee14:


shakalee14:
Can test deployment on fork because app.json (deployment environment variable file for heroku) has to be in the root directory of master

https://github.com/shakalee14/Spoke

app.json
```diff
@@ -0,0 +1,279 @@
...
+      "value": "assets.json"
+    },
+
+    "CAMPAIGN_ID": {
+      "description": "",
+      "required": false,
+      "value": "campaign-id-hash"
+    },
+
+    "DB_HOST": {
```
schuyler1d: I think all of these can be setup automatically by Heroku without the user needing to enter them.
We would probably need something like this in the src/server/models/thinky.js file:
```
+} else if (process.env.DATABASE_URL) {
+  var databaseType = process.env.DATABASE_URL.match(/^\w+/)[0]
+  config = {
+    client: (/postgres/.test(databaseType) ? 'pg' : databaseType),
+    connection: process.env.DATABASE_URL,
+    pool: {
+      min: process.env.DB_MIN_POOL || 2,
+      max: process.env.DB_MAX_POOL || 10
+    }
+  }
```

app.json
```diff
@@ -0,0 +1,279 @@
...
+      "value": "<changeme>"
+    },
+
+    "BASE_URL": {
+      "description": "based on your herokuapp name or domain name",
+      "required": true,
+      "value": "https://<changeme>.herokuapp.com"
+    },
+
+    "GRAPHQL_URL": {
```
schuyler1d: I pushed against this before, so I'm being a hypocrite but I realize we shouldn't make this another env variable you have to set.  Let's make it optional, and default to `BASE_URL + '/graphql`' (which I think is exactly what @sreynen was suggesting)

app.json
```diff
@@ -0,0 +1,279 @@
...
+      "value": ""
+    },
+
+    "ROLLBAR_ENDPOINT": {
+      "description": "rollbar api integration",
+      "required": false,
+      "value": "https://api.rollbar.com/api/1/item/"
+    },
+
+    "EMAIL_HOST": {
```
schuyler1d: Probably something for later, but might be nice to leverage this:
https://devcenter.heroku.com/articles/smtp
e.g. sendgrid:starter seems to be free.  If that just changes WHAT creds you need to put in, I guess that's useless/dumb.  But if there's some kind of auto-base usage level, we could make these optional when whatever the heroku ENV declares to exist.

docs/HEROKU_DEPLOY.md
```diff
@@ -0,0 +1,8 @@
+# Instructions for after you click - deploy to heroku
```
schuyler1d: this is a random thought from Scott's documentation link a while ago.  I wonder if we should prefix all our doc file names with e.g. HOWTO_ or REFERENCE_ etc.  thoughts?

app.json
```diff
@@ -0,0 +1,279 @@
+{
+  "name": "spoke",
+  "description": "Spoke",
```
schuyler1d: maybe something like "Mass p2p texting tool"

app.json
```diff
@@ -0,0 +1,279 @@
...
+  "repository":"https://github.com/moveonorg/spoke",
+  "success_url":"/setup.html",
```
schuyler1d: can we put envirionment variables in this url?  If so, it might be cool to bring them automatically to the organization creation page (i.e. a first invite) (which maybe makes them a superuser)

README.md
```diff
@@ -10,6 +10,11 @@ Spoke was created by Saikat Chakrabarti and Sheena Pakanati, and is now maintain
 
 This is generated from [react-apollo-starter-kit](https://github.com/saikat/react-apollo-starter-kit).  Look at that project's README for info on some of the libraries used.
 
+## Deploy to Heroku
+[![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy?template=https://github.com/shakalee/Spoke)
```
schuyler1d: don't forget to change `shakalee` here at the end.

schuyler1d (commented):
i know you're still working on this, but took a preliminary peek

app.json
shakalee14: db_host  and db_ creds are derived from heroku database credentials page so i have to figure out a way to populate these values from the other page

app.json
shakalee14: i'll take it out.

README.md
shakalee14: yea - i cant figure out a way to test this without master having an app.json so i'm using my fork to test for now.

Anasauce:
I didn't need the `GRAPH_QL_URL` and make sure you `http` not `https` for your Auth0 callbacks

shakalee14:
you right @Anasauce

shakalee14:
Can test with my fork : https://github.com/shakalee14/Spoke

shakalee14:
Note - that in my heroku instance, user experience tested from creating campaign, to sending texts, to replies

README.md
```diff
@@ -10,6 +10,15 @@ Spoke was created by Saikat Chakrabarti and Sheena Pakanati, and is now maintain
 
 This is generated from [react-apollo-starter-kit](https://github.com/saikat/react-apollo-starter-kit).  Look at that project's README for info on some of the libraries used.
 
+## Deploy to Heroku
+
+<a href="https://heroku.com/deploy?template=https://github.com/shakalee14/Spoke">
```
schuyler1d: change this to `MoveOnOrg` from shakalee :-)

schuyler1d: Also, we should have a warning here that this template incurs costs and is not supported on a free instance/version (or if it can, we should say that too)

app.json
schuyler1d: See here, for example:
https://devcenter.heroku.com/changelog-items/438

schuyler1d (changes requested):
A couple changes, but this is great!  Really exciting
(after changes, no need for another review)

Issue #235, opened 2017-09-19
Opt-out send button is easily confused with 'regular' send button


schuyler1d:
When the opt-out option is 'active', you can still click the regular send button.
Some things that could improve that:
* grey out/disable regular send button while opt-out card is displayed
* rename/recolor opt-out 'send' button to 'Send and Opt-out'

PR #246, opened 2017-09-20
hide other options while opt-out is active, also lint

sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/235 this does two things in a single commit, which I realize complicated review a bit, but I didn't know a good way to avoid. Suggestions on that welcome. The first thing it does is generally linting, which I started to reduce my own confusion in editing the code. The other thing, the main change, is hiding all the other options when the opt-out option is active, as none of them make sense to use when doing an opt-out. Those changes all relate to the `optOutDialogOpen` state variable.

### DIFF
--- a/src/containers/AssignmentTexterContact.jsx
+++ b/src/containers/AssignmentTexterContact.jsx
@@ -9,7 +9,7 @@ import FlatButton from 'material-ui/FlatButton'
 import NavigateCloseIcon from 'material-ui/svg-icons/navigation/close'
 import { grey100 } from 'material-ui/styles/colors'
 import IconButton from 'material-ui/IconButton/IconButton'
-import { Toolbar, ToolbarGroup, ToolbarSeparator } from 'material-ui/Toolbar'
+import { Toolbar, ToolbarGroup } from 'material-ui/Toolbar'
 import { Card, CardActions, CardTitle } from 'material-ui/Card'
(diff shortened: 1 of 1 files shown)


src/containers/AssignmentTexterContact.jsx
```diff
@@ -695,55 +693,55 @@ class AssignmentTexterContact extends React.Component {
...
-                    // pressing the Enter key submits
-                    this.handleClickSendMessageButton()
-                  }
-                }}
-              />
-              {this.renderCorrectSendButton()}
-            </GSForm>
-          </div>
+          {message}
           {this.renderActionToolbar()}
```
schuyler1d: I think rather than bring in the state to renderActionToolbar, I would vote that the logic here be something like:
{(this.state.optOutDialogOpen ? this.renderActionToolbar() : '')} (or however that's done in React/jsx-ese
My thinking is that the action toolbar as a component (I understand it's not (yet) one), shouldn't "know" about the optout dialog -- it should just be called when it should render or not.  I think it'll also be a little clearer in the code with the control flow that way, since if someone is reading the 'top' render view, then know how it works, rather than having to go down to the sub-(pseudoish)-components

schuyler1d (approved):
I had a suggestion for checking the state slightly differently, but it's optional.
:+1:

PR #251, opened 2017-09-21
do not save on section change so that admins can view campaign data w…

schuyler1d:
…ithout unintentionally saving it


sandramchung:
👍

PR #252, opened 2017-09-21
limit tasks per message job

sandramchung:
Trying to sort out the pending message parts piling up on prod.


schuyler1d (approved):
fingers crossed

PR #253, opened 2017-09-21
temporarily remove polling on texter todo list

sreynen:
This is a temporary fix to an apparent problem of too much database traffic when many texters have this interface open. Longer-term we can probably find a better fix for this.


sandramchung (approved):
👍

PR #258, opened 2017-09-22
remove all server-side rendering/hydration -- unnecessary and causing…

schuyler1d:
… errors on production


mathemagica:
👍 though i dont see how this would impact the number of connections the server is making to the db

PR #261, opened 2017-09-23
run migrations with an await, and track handleIncomingMessages so we …

schuyler1d:
…can diagnose why it is running only once per process

### DIFF
--- a/src/workers/job-processes.js
+++ b/src/workers/job-processes.js
@@ -96,27 +96,30 @@ export async function handleIncomingMessages() {
   setupUserNotificationObservers()
   console.log('Running handleIncomingMessages')
   // eslint-disable-next-line no-constant-condition
+  let i = 0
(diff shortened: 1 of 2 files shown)


src/workers/job-processes.js
```diff
@@ -96,27 +96,30 @@ export async function handleIncomingMessages() {
...
     } catch (ex) {
-      log.error(ex)
+      log.error('error at handleIncomingMessages', ex)
     }
   }
 }
 
 export async function runDatabaseMigrations(event, dispatcher) {
-  runMigrations(event.migrationStart)
+  await runMigrations(event.migrationStart)
```
schuyler1d, line 122: this wasn't important to run the migrations, but is apparently important for the models to trigger ensureIndex. why? :confused:

mathemagica:
👍

Issue #244, opened 2017-09-20
Texter assignments change after being saved


ErikKirkstein:
Texter Assignments are changing after being saved. 

Example 1: I setup Texter A with 500 texts. I save. i then open the campaign to add another Texter, Texter B. When I open the tab to assign tTexter, Texter A's assignment has changed to a larger or smaller number. I add Texter B and assign 500, then adjust Texter A's assignment back to 500. I open the campaign again to add Texter C, and Texter A or B's assignment has changed. I adjust and save. Rinse. Repeat.

Example 2: Texter D was assigned 500. 20 minutes later they message me on Slack asking if I added more texts to their assignment, or say that they're see 1048 texts, not the 500. Or they say their text disappeared.

It appears that something is happening where the assignment are not "locked in" in any way, but instead seem to change on their own, and for what appears to be arbitrary reasons. This does not appear to happen to all campaigns in a given day, just some, and for unknown reasons.

mathemagica:
https://github.com/MoveOnOrg/Spoke/pull/262 *might* fix that/those issues

mathemagica:
Next steps: download sky's testing csv and try to reproduce the issue locally. we are tentatively sure we fixed this on prod.

PR #262, opened 2017-09-25
simplify logic for auto-assigning texts, possibly dodging a race cond…

schuyler1d:
…ition seen in #244 (might fix it?)

### DIFF
--- a/src/components/CampaignTextersForm.jsx
+++ b/src/components/CampaignTextersForm.jsx
@@ -152,9 +152,9 @@ export default class CampaignTextersForm extends React.Component {
       }
     })
 
-    let extra = totalNeedsMessage + totalMessaged - this.formValues().contactsCount
-    if (extra > 0) {
-      // 2. With extra texter capacity beyond contacts, remove contact counts from the top
+    let extraTexterCapacity = totalNeedsMessage + totalMessaged - this.formValues().contactsCount
+    if (extraTexterCapacity > 0) {
+      // 2. With extraTexterCapacity beyond contacts, remove contact counts from the top texters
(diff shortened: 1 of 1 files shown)


src/components/CampaignTextersForm.jsx
```diff
@@ -168,16 +171,17 @@ export default class CampaignTextersForm extends React.Component {
...
+      // 3. if we don't have extras and auto-split is on, then fill the texters with assignments
       const factor = 1
       let index = 0
       let skipsByIndex = new Array(newFormValues.texters.length).fill(0)
-      if (newFormValues.texters.length === 1 && this.state.autoSplit) {
+      if (newFormValues.texters.length === 1) {
         const messagedCount = newFormValues.texters[0].assignment.contactsCount - newFormValues.texters[0].assignment.needsMessageCount
         newFormValues.texters[0].assignment.contactsCount = this.formValues().contactsCount
         newFormValues.texters[0].assignment.needsMessageCount = this.formValues().contactsCount - messagedCount
-      } else if (newFormValues.texters.length > 1 && (extra > 0 || (extra < 0 && this.state.autoSplit))) {
```
sreynen, line 179: Look like this is removing a case where this code would run with autosplit off and extra > 0, but it's not clear to me reading this code what `extra` is, or how it might end up greater than 0, so I don't know if removing that is a mistake or intentional. Maybe we can walk through how this works?

schuyler1d: Yes, that's the missed condition I believe is causing the bug.

Ill add a change to rename extra -> unassignedContactCount which I think is more clear and descriptive

PR #263, opened 2017-09-25
offline dev, removing opt-outs on client-side, and job_request.result_message

schuyler1d:
i.e. stuff I did on an airplane
possibly fixes:
#248  (Allow a server-job to communicate a failure message back to the client)
#249 (AdminCampaignEdit loads all optout numbers from server)

PR #268, opened 2017-09-29
Removes unnecessary logging

shakalee14:


schuyler1d:
Yeah, that was annoying, right?  It's solved in this patch already:
https://github.com/MoveOnOrg/Spoke/pull/263/files#diff-d82d53538123a15a45b9aa398d605c1aL97
I'll merge that since it was approved

Issue #247, opened 2017-09-20
Add lint to automated Travis testing


sreynen:
https://github.com/MoveOnOrg/Spoke/blob/master/CONTRIBUTING.md says the project has 2 quality requirements: all tests passing and no lint errors. Those both seem like good requirements. But so far we're only testing the first one, not automatically checking for lint errors at all. Many lint errors are simply syntax preferences, so only marginally increase the difficulty of contributing. Some lint errors suggest flaws in how the code functions, more directly impacting end users.

To be more explicit on what "complete" looks like on this issue: opening a PR with lint errors should result in Travis tests failing and opening a PR without lint errors (also without test errors) should result in Travis tests passing.

On the latter part of that, currently we have many lint errors, so keeping clean PRs at passing requires either fixing all those errors before adding the Travis check, or somehow setting up the test in a way that only runs lint on files relevant to the current PR.

Marking this "easy," because it should be. But unlike most easy tasks, this one is likely to take a long time, as we have a lot of small lint errors to fix.

PR #275, opened 2017-10-02
lint autofixes

sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/247, this is just what `npm run lint` does automatically.


schuyler1d:
+1 as long as it merges cleanly with stage-master. Otherwise I suggest we run this after the current pr queue

shakalee14 (approved):

PR #284, opened 2017-10-04
fixes #282: wrong boolean sign on jobs with result_messages to mark completed

### LINKED ISSUE #282: Bug preventing campaign from starting
Erik created campaigns 203 and 204 today and experienced the error that even after he had set up the campaign to run properly, the "run campaign' button remained grayed out, and is_started never got set to true in the DB. Sandra and I think this was introduced with https://github.com/MoveOnOrg/Spoke/pull/263/commits/bc9c9f7593bc5421850c72941d2beb12e81743ca but have not yet tested reverting it. To temporarily workaround this to get Erik started on campaign 203 we just set is_started=t. Campaign 204 still exhibits the error behavior can be used for testing.

additionally, even when is_started=t is hacked into place, message receiving is also broken. Perhaps due to clearing out jobs older than 2min which may still be valid jobs?

schuyler1d:
testing for NOT error to block was wrong -- sort of a double negative.

### DIFF
--- a/src/containers/AdminCampaignEdit.jsx
+++ b/src/containers/AdminCampaignEdit.jsx
@@ -242,8 +242,8 @@ class AdminCampaignEdit extends React.Component {
 
   checkSectionSaved(section) {
     // Tests section's keys of campaignFormValues against props.campaignData
-    // * Determines greyness of button
-    // * Determine if section is marked done (in green)
+    // * Determines greyness of section button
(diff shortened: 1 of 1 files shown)


mathemagica:
I think this'll fix not being able to reply to messages. But noting that this probably doesnt address the issue where newly created campaigns are not startable even after all campaign sections have been filled out.

👍

PR #287, opened 2017-10-04
multi-send on messages when people hold down enter key (before submit completes)

### LINKED ISSUE #286: Texting using keyboard "enter" button sent contacts multiple texts
Volunteer texters who used the "enter" or "return" button on their keyboards to send text messages in Spoke reported receiving replies from individual members indicating they had received multiple texts, sometimes up to 20. This seems to be connected to a volunteer "holding down" the enter button as opposed to clicking it repeatedly, but I am not certain I can say the later with confidence.

schuyler1d:
fixes #286


mathemagica:
👍

Issue #276, opened 2017-10-02
Create ability to track active user count


schuyler1d:
Using Auth0, we can use this API:
https://auth0.com/docs/api/management/v2#!/Stats/get_active_users

Otherwise we could use # of unique texters for sends in a given time window.

Tracking should support amazon cloudwatch, but also heroku metrics if such a thing exists

sandramchung:
What temporal resolution do we need here? It looks the best we can get from the Auth0 API stats endpoint is daily login count. We could alternatively parse the response from Auth0 log endpoint to get login successes within an arbitrary time period.

sandramchung:
Of these ideas, counting unique texters among sends and receipts in a given time window will probably give us the best figure to hold up against load on the db/service at any given time.

schuyler1d:
so total sends per [5 minutes], total receipts per [5 minutes], total unique texters-that-sent-a-message per [5 minutes], is what i think we came to at standup

sandramchung:
Per standup today: aim for 1 hr temporal resolution via querying unique texters in the message table, and since I've already written code to send it to cloudwatch, send it to cloudwatch.

sandramchung:
Hmm, I heard 1 hour resolution and one query.

sandramchung:
Manual queries documented here: https://github.com/MoveOnOrg/Spoke/blob/active-user-count/docs/TEXTER_ACTIVITY_QUERIES.md

sandramchung:
https://github.com/MoveOnOrg/Spoke/pull/289 We can run these queries against our data warehouse to generate reports without burdening the production DB at all. Will implement cloudwatch metrics later.

PR #289, opened 2017-10-04
Active user count

sandramchung:
Can run these queries manually to grab user activity stats. Per #276


schuyler1d (approved):
:+1: 
optional: do we want to rename this to our 'new' naming scheme (prefixing with reference/howto/etc)?

shakalee14 (approved):

PR #290, opened 2017-10-04
Fix unmessaged count

schuyler1d:
This makes the messaged count for past campaigns still report the right number.  This was a dumb "feature" -- probably related to weird behavior for timezones/send-later and stopping the messages, but we should unhack this and just ask for what we want on the client/server vs. due unpredictable things related to due by date

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ rethinkdb_data
 __test__/test_data/*
 production-env.json
 claudia.json
+test.sqlite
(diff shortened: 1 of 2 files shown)


mathemagica:
Does any functionality related to viewing pastDue campaigns right now break for the mobile team? Did they use this earlier?

schuyler1d:
Before this, no one knew what past Due did.  Erik says that he's generally made it 'well past' whenever the campaign runs for, but didn't know what it did.

Looking at the code, it looks like there's an obscure feature that if we enable timezones for a campaign, and then someone clicks 'send later' for when the timezone is available, then it would only send if the campaign were not past due.  This all seems like an anti-feature -- sending messages the next day when the texter is no longer around.  I'll await feedback from others on whether we should excise fhis from the code altogether.  But the reports should work, and there's no reason to break them just because the campaign completed.

mathemagica:
👍

PR #291, opened 2017-10-04
bugfix: increasing assignment used to double assignments

### LINKED ISSUE #269: Texter assignment form behavior is confusing/unpredictable
Texter assignment form behavior can get very confusing as the campaign progress and you return to add more contacts and/or texters and/or distribute or redistribute contacts between the texters. This is at least in part because the permutations of possibilities in this form are too many to test and think through comfortably, and we're not testing or thinking through all of them yet.
Propose:
1) Writing tests for the form permutations and using those to update the texter assignment algorithm
2) Converting autosplit into a simple button that does exactly one thing: divides the assignable contacts between the listed texters as evenly as possible, regardless of how many contacts the texters have already contacted.

schuyler1d:
… rather than just increment the assignments
fixes #269

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -288,9 +288,11 @@ export async function assignTexters(job) {
       unchangedTexters[assignment.user_id] = true
       return null
     } else {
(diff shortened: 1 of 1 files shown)


mathemagica:
👍

PR #293, opened 2017-10-05
testing bootstrap

schuyler1d:
fixes startup of database

### DIFF
--- a/__test__/_setup.test.js
+++ b/__test__/_setup.test.js
@@ -1,0 +1,16 @@
+import { sleep } from '../src/workers/lib';
+import { r } from '../src/server/models/';
+
+beforeAll(async () => {
+  let testDbExists = false
+  while (testDbExists) {
+    const testDbExists = await r.knex.schema.hasTable('job_request')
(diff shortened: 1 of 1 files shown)


__test__/_setup.test.js
```diff
@@ -0,0 +1,16 @@
+import { sleep } from '../src/workers/lib';
+import { r } from '../src/server/models/';
+
+beforeAll(async () => {
+  let testDbExists = false
+  while (testDbExists) {
```
sreynen: I think this should be `while (!testDbExists) {`

__test__/_setup.test.js
```diff
@@ -0,0 +1,16 @@
...
+    const testDbExists = await r.knex.schema.hasTable('job_request')
```
sreynen: I think `const` here should be dropped, since `testDbExists` is declared above.

sreynen (changes requested):
See inline comments.

sreynen (approved):

PR #297, opened 2017-10-05
add database migration instructions for lambda

schuyler1d:


mathemagica:
👍  yay for dated migrations!

PR #299, opened 2017-10-05
Fix post-test DB wipe, pre-test DB check, and foreign key dependency order

sreynen:
This generally ensures everything around table creation and dropping happens (dropping wasn't working before), and happens in the right order to remove any errors around foreign keys.


schuyler1d (approved):
:+1:

Issue #279, opened 2017-10-02
Audit staging to allow for external testers, add features to support volunteer QAers


schuyler1d:
https://github.com/MoveOnOrg/Spoke/pull/272 does some work for this.

mathemagica:
Context: progcoder friends have volunteered to help test spoke. Make sure our staging env and test data are in a state that we consider to be friendly to the public, and that onboarding docs are ready

mathemagica:
Just to make the goal of this task explicit: we want to be able to let external, non-staff folks into the system to help test. To do this safely, we need to make sure that these folks dont contact any of our member data, dont incur costs that bill to us without our consent, and dont send messages "as" us that could be attributed to us. So to make this happen, we're making a special test org that has twilio message sending disabled, and that is logically isolated from other org data.

PR #300, opened 2017-10-06
Per org maxcontacts service

schuyler1d:
This adds some per-organization features (currently, only settable with direct database access) including which service is used, and if there is a maxContacts count.

This addresses #279, because we can now make QAers admins or set the maxContacts for only our own orgs to actually send messasges on a service or send them but have a maxContacts setting.

For the same purpose, we also have a `fakeservice` that joins twilio and nexmo for DEFAULT_SERVICE as an option, which means it won't try to send through twilio, etc and you can use, e.g. just the `/admin/1/campaigns/5/send-replies` url to fake-send replies.  This will also be useful for testing sans-service.


sandramchung (approved):
👍  fc49a7d916fae99bcc60ff0bfdaed5b4e4a8d545

src/server/api/schema.js
```diff
@@ -322,7 +323,7 @@ async function editCampaign(id, campaign, loaders, user) {
 
 const rootMutations = {
   RootMutation: {
-    sendReply: async (_, { id, message }, { loaders }) => {
+    sendReply: async (_, { id, message }, { user, loaders }) => {
       if (process.env.NODE_ENV !== 'development') {
```
sandramchung: I imagine we'll want to use this to test in a production-like environment - so we'd have to deploy it in dev mode to do that?

schuyler1d: ah, i missed that.  i'm going to make is so development node env OR have correct access

mathemagica:
Belatedly, thanks for this- this'll make testing much easier!

Issue #294, opened 2017-10-05
Twilio Incoming Messages Sometimes Fail


shakalee14:
Not sure if we always receive incoming messages. While testing on stage, I had to send a message twice and only received the second reply. Checked first in `pending_message_part` table to see if it just wasn't processed but saw 0 rows. Then checked twilio and saw this:

<img width="938" alt="screen shot 2017-10-05 at 11 32 16 am" src="https://user-images.githubusercontent.com/16676323/31236018-e6c0bbf8-a9c0-11e7-83f9-b427b7f17f5c.png">

shakalee14:
This issue may occur because lambda instance hasn't 'warmed up' yet and could be rectified by using a different fallback url in our `inbound settings` on twilio

shakalee14:
https://github.com/MoveOnOrg/Spoke/pull/310

PR #310, opened 2017-10-06
adds sqs message processing function

shakalee14:


.env.example
```diff
@@ -35,6 +35,7 @@ TWILIO_APPLICATION_SID=
 TWILIO_AUTH_TOKEN=
 TWILIO_MESSAGE_SERVICE_SID=
 TWILIO_STATUS_CALLBACK_URL=
+TWILIO_SQS_QUEUE_URL=
```
schuyler1d, line 38: I don't think we should make these optional items in `.env.example` (feel free to prefer something different) but we should definitely add them to `docs/REFERENCE-environment_variables.md`. In this case, probably include a link the SQS twiliio blog post https://www.twilio.com/blog/2017/07/handling-high-volume-inbound-sms-and-webhooks-with-twilio-functions-and-amazon-sqs.html

schuyler1d (approved):
and deploy to stage for testing!

src/workers/job-processes.js
```diff
@@ -137,6 +137,7 @@ const processMap = {
 const syncProcessMap = {
   // 'failedMessageSender': failedMessageSender, //see method for danger
   handleIncomingMessages,
+  processSqsMessages,
```
schuyler1d: I just realized that as currently implemented, we'll process 10 messages per 5 minutes.  I think we'll need to add a process loop in job-processes that runs `await processSqsMessages()` inside a sleep loop the same way we do with some of the others. (like here: https://github.com/MoveOnOrg/Spoke/pull/310/files#diff-d82d53538123a15a45b9aa398d605c1aL28)  
but even as-is, this should still prove that it will actually load the data from sqs and process it anyway.

shakalee14:
fixes https://github.com/MoveOnOrg/Spoke/issues/294

src/workers/job-processes.js
```diff
@@ -137,6 +149,7 @@ const processMap = {
...
```
schuyler1d: I think you want this to be `checkMessageQueue` now

src/workers/job-processes.js
```diff
@@ -42,6 +42,18 @@ export async function processJobs() {
   }
 }
 
+export async function checkMessageQueue() {
+  console.log('checking if messages are in message queue');
```
schuyler1d: To stop continually running something that is only used during a setting, let's add the same
```
if(!process.env.TWILIO_SQS_QUEUE_URL) {
   return
}
```
to the top of the function here, so we don't loop at all when the setting isn't there.

schuyler1d (changes requested):
after changes :+1:

src/workers/job-processes.js
shakalee14: oh yea of course

schuyler1d (approved):
:+1:

Issue #302, opened 2017-10-06
Change tests in backend.test.js to re-use data


sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/277

Currently each test creates the data it needs, which won't give us a great measure of performance, since tests later in a process will take much longer than tests earlier in the process. Let's change this so each tests builds on the previous, re-using the already-created data.

Issue #277, opened 2017-10-02
Spoke perf testing framework: end-to-end message sending test case


mathemagica:
Context: we need a way to simulate having many texters on spoke at the same time so we can measure how well the system performs at scale without undue risk of prod breakage. Goal of these perf testing tasks is not to load test the entire system end-to-end, but be able to load test just the parts of the code we change, so that we can have some sense ahead of time whether code changes we're about to roll out will causing scaling issues. 

Using __test__/backend.test.js as an example of how to mock out the current logged in user, and https://github.com/MoveOnOrg/Spoke/blob/master/src/server/api/schema.js#L323 as an example of how to mock twilio interaction, write a test that mock logs in as a user, sends a configurable number of mocked text messages associated with a mocked campaign, and responds to mocked replies. 

Document how to run this test on a locally run server using your local postgres database, and also pointing at the staging postgres database.

ibrand:
@sreynen @mathemagica @schuyler1d where did this leave off? As far as I know, we don't have a test like this but we do now have e2e cypress stubs (@matteosb) for more solid integration testing

Frydafly:
> @sreynen @mathemagica @schuyler1d where did this leave off? As far as I know, we don't have a test like this but we do now have e2e cypress stubs (@matteosb) for more solid integration testing

following up! lifting up @ibrand 's comment above

schuyler1d:
This doesn't yet exist in our main repo.  I had a thought that cypress could support this, but that would be more difficult for specifically *performance* testing -- It might be better to use e.g. Artillery or @tmc has considered using https://k6.io/docs/using-k6/scenarios

Arique1104:
Hello,

My name is Arique (they/them) and I am the Community Manager for Spoke. First of all, thank you so much for creating this issue and for everyone’s contributions to this ticket.

I am going through all of our issues and beginning to close all issues created in 2018 or older. Your issue qualifies.

I am giving everyone one week to answer the following question: 

- **Has this issue been resolved during the subsequent Spoke Updates?**

On November 16th I will be closing this issue if:
1. There is no response
2. You respond yes

If you are reading this message after November 16th, please note that if you tag me and answer the above question, I will be happy to reinstate the issue.

ibrand:
This has not been resolved. I'd leave this open. Load testing would be cool to have in Spoke one day even if it's not super high priority.

Arique1104:
Thank you!

PR #315, opened 2017-10-09
Persist test data

sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/302 (part of https://github.com/MoveOnOrg/Spoke/issues/277), this avoids recreating data in each test, to avoid data creation changes having an outsized impact on test time.


schuyler1d (approved):
:+1:  as-is.
It would be a bit more idiomatic, I think, if you put the persisted objects along with the `it()`s that use them within a `describe(.... { }) block.

Issue #303, opened 2017-10-06
Add test that adds texter account


sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/277

PR #318, opened 2017-10-09
add texter test

sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/303 (part of https://github.com/MoveOnOrg/Spoke/issues/277), this adds a test for adding a texter.

This builds on https://github.com/MoveOnOrg/Spoke/pull/315, so only the last commit here is relevant.

### DIFF
--- a/__test__/backend.test.js
+++ b/__test__/backend.test.js
@@ -25,14 +25,14 @@ let testTexterUser
 
 // data creation functions
 
-async function createUser() {
-  const user = new User({
-    auth0_id: 'test123',
-    first_name: 'TestUserFirst',
-    last_name: 'TestUserLast',
-    cell: '555-555-5555',
-    email: 'testuser@example.com',
-  })
+async function createUser(userInfo = {
+  auth0_id: 'test123',
+  first_name: 'TestUserFirst',
+  last_name: 'TestUserLast',
+  cell: '555-555-5555',
+  email: 'testuser@example.com',
+}) {
+  const user = new User(userInfo)
   try {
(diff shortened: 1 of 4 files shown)


__test__/backend.test.js
```diff
@@ -204,34 +206,39 @@ it('should convert an invitation and user into a valid organization instance', a
...
-//
-//
-//
-// })
+it('should add texters to a organization', async () => {
+  testTexterUser = await createUser({
+    auth0_id: 'test456',
+    first_name: 'TestTexterFirst',
+    last_name: 'TestTexterLast',
+    cell: '555-555-6666',
```
schuyler1d, line 226: not sure we use this anywhere, but are `-`s allowed? (or is that part of the test)?

__test__/backend.test.js
```diff
@@ -204,34 +206,39 @@ it('should convert an invitation and user into a valid organization instance', a
...
+  expect (contact.campaign_id).toBe(parseInt(testCampaign.data.createCampaign.id))
 })
 
-// it('should add texters to a organization', async () => {
-//
-//
-//
-// })
+it('should add texters to a organization', async () => {
+  testTexterUser = await createUser({
```
schuyler1d, line 222: We don't need to do this as part of this PR, but I'm tempted to suggest we move the `createUser()`, etc to a library file, and then also have a `createTexter()` which fills in arbitrary user data, so we can run it 100 times in one test easily. #future

schuyler1d (approved):
:+1:

Issue #304, opened 2017-10-06
Add test that assigns campaign contacts to texter


sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/277

PR #323, opened 2017-10-10
texter assignment

sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/304 (part of https://github.com/MoveOnOrg/Spoke/issues/277), adds texter assignment test. Also 
adds JOBS_SYNC flag, so tests can wait on related jobs (like texter assignment) to complete.

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -277,16 +277,10 @@ async function editCampaign(id, campaign, loaders, user) {
     })
 
     if (JOBS_SAME_PROCESS) {
-      console.log('JOBS_SYNC')
-      console.log(JOBS_SYNC)
-      console.log('global.JOBS_SYNC')
-      console.log(global.JOBS_SYNC)
       if (JOBS_SYNC) {
-        console.log('assignTexters sync')
         await assignTexters(job)
       }
       else {
-        console.log('assignTexters async')
         assignTexters(job)
       }
     }
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -92,6 +92,7 @@ import { uploadContacts,
 const uuidv4 = require('uuid').v4
 
 const JOBS_SAME_PROCESS = !!process.env.JOBS_SAME_PROCESS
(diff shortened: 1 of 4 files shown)


src/server/api/schema.js
```diff
@@ -244,7 +245,7 @@ async function editCampaign(id, campaign, loaders, user) {
       payload: compressedString.toString('base64')
     })
     if (JOBS_SAME_PROCESS) {
-      uploadContacts(job).then()
+      uploadContacts(job)
```
schuyler1d, line 248: we need to keep these `.then()`s same as below, at least for prod-ish environments.

Were you intending to complete this as:
```
if (JOBS_SAME_PROCESS) {
   (JOBS_SYNC ? await uploadContacts(job) : uploadContacts(job).then())
}
```

src/server/api/schema.js
```diff
@@ -276,7 +277,12 @@ async function editCampaign(id, campaign, loaders, user) {
     })
 
     if (JOBS_SAME_PROCESS) {
-      assignTexters(job).then()
+      if (JOBS_SYNC) {
+        await assignTexters(job)
+      }
+      else {
+        assignTexters(job)
```
schuyler1d, line 284: same for here, etc.

schuyler1d (approved):
need to bring back .then()s please update before merge.  if you restore them all then :+1:

schuyler1d (changes requested):
changes needed

src/server/api/schema.js
sreynen: I believe these `.then()` calls aren't doing anything useful. I think they're only cloning the promise object returned by the async function, without changing it at all.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then is the closest thing I could find to directly describing an empty `.then()` call:

> If one or both arguments are omitted, or are provided non-functions, then then will be missing the handler(s), but will not generate any errors.

schuyler1d (approved):
Ok, after testing this, @sreynen is right -- the `.then()` doesn't do anything.  sorry about the hangup
:+1:

PR #324, opened 2017-10-11
Adds logging to AssignmentTexterContact component and getContact()

shakalee14:


schuyler1d (approved):
:+1: should help seeing what's going on

Issue #160, opened 2017-08-25
Add ActionKit survey action


sreynen:
- [ ] As an admin, I should be able to mark that a question response should trigger completion of an ActionKit survey.
- [ ] As a texter, I should be able to choose a response associated with the ActionKit survey action handler, and complete the survey on the contact's behalf.

Requested by MoveOn's Field Team, I _think_ this only makes sense for surveys with no required fields, so simply completing the survey is meaningful information, but worth discussing more the exact use cases.

PR #325, opened 2017-10-11
fix bugs from first draft for actionkit-rsvp #160

schuyler1d:


src/server/action_handlers/actionkit-rsvp.js
```diff
@@ -73,13 +76,23 @@ export async function processAction(questionResponse, interactionStep, campaignC
...
+          if (err || (body && body.error)) {
+            console.error('error: actionkit event sign up failed', err, userData, body)
           } else {
+            if (httpResponse.headers && httpResponse.headers.location) {
+              const actionId = httpResponse.headers.location.match(/action_id=([^&]+)/)
+              if (actionId) {
+                //save the action id of the rsvp back to the contact record
+                customFields['processed_event_action'] = actionId[1]
+                await r.knex('campaign_contact')
+                  .where('campaign_contact_id', campaignContactId)
```
shakalee14: where does this `campaign_contact_id` come from? is this a new column, or should it be `campaign_contact.id`?

schuyler1d: good catch!

mathemagica:
👍

shakalee14 (approved):

shakalee14:
it would be nice to have a `HOWTO_INTEGRATE_WITH_ACTIONKIT.md`, thoughts @schuyler1d ? even if it was to just make sure certain variables are set?

PR #327, opened 2017-10-11
fixes #322: when client admin has different messagedCount than has ...

### LINKED ISSUE #322: Resaving assignments while texters are active accidentally increases assignment totals
Description:
1. Campaign admin assigns some texters, e.g. Alice 100, Bob 100
2. The campaign begins and Alice and Bob start sending texts
3. Campaign admin revisits assignments, opening up assignment screen.  So far Alice and Bob have sent 25 texts, so the screen shows 75 for each.
4. While admin has the screen open Alice and Bob continue sending messages and they send 25 more texts
5. Admin updates Bob's assignment to 200 and clicks save

The desired result is that Bob should get a total of 225 and Alice, should have a total of 100.

However, because the client just sends the current 'needsMessageCount' for each (and all) users, Alice will end up with 125 total assigned and Bob will have 250.  The 25 count difference is because of what happened between opening that page the saving.

I propose two solutions:
1. assignment changes should only be for texters with changed assignments (rather than all texters, each time)
2. Additionally, or alternatively we can send the client's 'current' sent count and then figure out the deltas implied by the client instead of getting the absolutes.

schuyler1d:
occurred while page is open, then we still want to "credit" the texter with the messages they have sent since the admin opened the client

### DIFF
--- a/src/containers/AdminCampaignEdit.jsx
+++ b/src/containers/AdminCampaignEdit.jsx
@@ -204,7 +204,8 @@ class AdminCampaignEdit extends React.Component {
       if (newCampaign.hasOwnProperty('texters')) {
         newCampaign.texters = newCampaign.texters.map((texter) => ({
           id: texter.id,
-          needsMessageCount: texter.assignment.needsMessageCount
+          needsMessageCount: texter.assignment.needsMessageCount,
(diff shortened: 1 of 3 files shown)


mathemagica:
👍

PR #334, opened 2017-10-12
WIP: User data editing

schuyler1d:


shakalee14:
I can't get this branch to work locally, I get this error:

`ApolloError {graphQLErrors: Array(1), networkError: null, message: "GraphQL error: Expected Output type."}`

schuyler1d:
Well, it IS work-in-progress :-)
Is it possible you're sending a user id that doesn't exist?  Not sure how to reflect it in the types, but it might be getting hung up on this line: `if (!userRes || !userRes.length) { return null }`

schuyler1d:
will re-open with user-data-editing2

Issue #317, opened 2017-10-09
Sending texts very slow: texters reporting 10 sec between sends


mathemagica:
During current campaign 10/9 6pm ET, Erik reported there are only 2 texters with a list size of 880on and they are reporting very slow texting speed- 10 sec between sends. but this only happens intermitently

DB CPU is low - like ~3%, so doesn't look like DB is a bottleneck this time. Lots of errors in the cloudwatch logs like this though:
2017-10-09T22:07:28.384Z	3d12cfab-ad3e-11e7-9ed8-4d2d6fe39bbd	WARNING: Attempting to listen on socket /tmp/server354.sock, but it is already in use. This is likely as a result of a previous invocation error or timeout. Check the logs for the invocation(s) immediately prior to this for root cause, and consider increasing the timeout and/or cpu/memory allocation if this is purely as a result of a timeout. aws-serverless-express will restart the Node.js server listening on a new port and continue with this request.

also seeing lots of log messages like this:
2017-10-09T22:12:40.617Z	8c17157a-ad3e-11e7-9b7c-fb39a422414f	entering handleIncomingMessages. round: X 

where X steadily increments, like this:

2017-10-09T22:09:40.868Z	8c17157a-ad3e-11e7-9b7c-fb39a422414f	entering handleIncomingMessages. round: 1
2017-10-09T22:09:41.369Z	8c17157a-ad3e-11e7-9b7c-fb39a422414f	entering handleIncomingMessages. round: 2
...
2017-10-09T22:12:40.617Z	8c17157a-ad3e-11e7-9b7c-fb39a422414f	entering handleIncomingMessages. round: 358

and then

2017-10-09T22:12:40.965Z 8c17157a-ad3e-11e7-9b7c-fb39a422414f Task timed out after 180.10 seconds

schuyler1d:
the handleIncomingMessages is expected behavior.  That logging was meant to confirm that it gets past `round: 1` which it wasn't when our DB was throttled.

Presumably when the users complain about sending taking 10sec, they mean the UI, rather than it taking 10 seconds for the message to be received by the contact?

schuyler1d:
Would also be worth checking whether the '10 second lag' is actually related to the send, vs, the `getContact()` for the next view.
If the latter, then probably the longterm solution, at least is to load several contacts at once rather than per-page

mathemagica:
[...]

PR #344, opened 2017-10-18
Follow aws lambda best practices for import/handler code separation

schuyler1d:
may related to #317
See:
http://docs.aws.amazon.com/lambda/latest/dg/best-practices.html#function-code
"Separate the Lambda handler (entry point) from your core logic"


mathemagica:
👍

Issue #349, opened 2017-10-18
Remove unused dependencies


sreynen:
There's at least one dependency (stripe) that isn't used anywhere. Unused dependencies slow down anything that involves dependencies, e.g. build and deploy, so should be removed.

PR #350, opened 2017-10-18
remove stripe dependency

sreynen:
For #349, this removes stripe from the dependency list. Stripe doesn't appear anywhere in the code: https://github.com/MoveOnOrg/Spoke/search?utf8=%E2%9C%93&q=stripe&type=


sandramchung (approved):
👍

PR #351, opened 2017-10-18
declare dev-only dependencies as devDependencies

sreynen:
None of these dependencies exist in /src, so shouldn't be needed for production, but I think all are used for development.


sandramchung (dismissed):
👍

sandramchung (commented):
`npm install --production` fails unless I add "babel-plugin-transform-runtime": "^6.23.0" and "babel-preset-es2017": "^6.24.1" back to dependencies. And then I can't `npm run dev` without nf.

sreynen:
I'd forgotten `postinstall` does prod-build, which adds a bunch of dev requirements to `npm install`, so I think it doesn't actually make sense to move most of this. Thanks for catching that!

PR #352, opened 2017-10-18
remove bonzo, unused

sreynen:
For #349, this dependency isn't used anywhere: https://github.com/MoveOnOrg/Spoke/search?utf8=%E2%9C%93&q=bonzo&type=

https://www.npmjs.com/package/bonzo says this is for DOM manipulation, which we're doing with React, so not sure why this was ever added as a dependency.


sandramchung (approved):
👍

Issue #353, opened 2017-10-20
Question Reply Info Not Included in Data Export


shakalee14:
- Currently, when you export, you receive a csv with all the information except contacts responses to different questions

Example: If the question was `Will Volunteer?`, a column will be created for that question --> `question[Will Volunteer?]` but the contact's response (that is marked by the texter) is blank for all rows. All the other columns seem to update with the correct information.

PR #354, opened 2017-10-23
Changes interaction_step_id to stepId comparison to be less strict

shakalee14:
StepId and interaction_step_id types are different - so uses abstract comparison

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -551,14 +551,21 @@ export async function exportCampaign(job) {
 
       const questionResponses = await r.table('question_response')
         .getAll(contact.id, { index: 'campaign_contact_id' })
+
       Object.keys(allQuestions).forEach((stepId) => {
         let value = ''
         questionResponses.forEach((response) => {
-          if (response.interaction_step_id === stepId) {
+          if (response.interaction_step_id == stepId) {
             value = response.value
           }
         })
+
+        // console.log('value before:', value);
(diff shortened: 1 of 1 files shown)


shakalee14:
To test:
- Must have AWS related variables set in environment variables
- Must have a question step in your interaction script
- Must have a contact who replied and a texter who recorded the answer

shakalee14:
fixes #353

src/workers/jobs.js
```diff
@@ -551,14 +551,21 @@ export async function exportCampaign(job) {
 
       const questionResponses = await r.table('question_response')
         .getAll(contact.id, { index: 'campaign_contact_id' })
+
       Object.keys(allQuestions).forEach((stepId) => {
         let value = ''
         questionResponses.forEach((response) => {
-          if (response.interaction_step_id === stepId) {
+          if (response.interaction_step_id == stepId) {
```
schuyler1d: I've been fixing this in other places with an explicit `parseInt(...)` on the right side.  some very future date, we might try to move it all to integers (or back to strings!) and having it explicit will probably be good for that moment.

schuyler1d (approved):
nice catch!
:+1: 
(optional nit)

src/workers/jobs.js
shakalee14: response.interaction_step_id is a number
stepId is a string

PR #364, opened 2017-10-31
if no db entry, do not set timezone_offset=undefined_undefined and al…

schuyler1d:
…low to dispatch seedZipCodes by process

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -23,9 +23,11 @@ async function getTimezoneByZip(zip) {
       return `${rangeZip[2]}_${rangeZip[3]}`
     } else {
       const zipDatum = await r.table('zip_code').get(zip)
-      if (zipDatum) {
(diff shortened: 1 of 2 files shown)


src/workers/jobs.js
```diff
@@ -23,9 +23,11 @@ async function getTimezoneByZip(zip) {
       return `${rangeZip[2]}_${rangeZip[3]}`
     } else {
       const zipDatum = await r.table('zip_code').get(zip)
-      if (zipDatum) {
+      if (zipDatum && zipDatum.timezone_offset && zipDatum.has_dst) {
         zipMemoization[zip] = `${zipDatum.timezone_offset}_${zipDatum.has_dst}`
         return zipMemoization[zip]
+      } else {
+        return ''
```
shakalee14, line 30: so if it isn't found, it'll return an empty string - lgtm!

shakalee14 (approved):

PR #367, opened 2017-11-02
npm script to automate tunnelling on dev environment using Ngrok

benmort:


package.json
```diff
@@ -34,7 +34,8 @@
     "dev-start": "nodemon -e js,jsx,md,mustache -w ./src --exec ./dev-tools/babel-run -- ./src/server",
     "generate-contacts": "./dev-tools/babel-run ./dev-tools/generate-contacts.js",
     "jest-test": "jest __test__/backend.test.js",
-    "dev": "nf start -w --procfile ./dev-tools/Procfile.dev"
+    "dev": "npm run tunnel | nf start -w --procfile ./dev-tools/Procfile.dev",
+    "tunnel": "node ./tunnel.js"
```
schuyler1d, line 38: I think we want something like `if [ "$NGROK_AUTH_TOKEN" != \"\" ] ; then node ./tunnel.js; fi` for the tunnel job -- basically it should be able to work without ngrok.

.env.example
```diff
@@ -46,3 +46,6 @@ EMAIL_HOST_PASSWORD=
 EMAIL_HOST_USER=
 EMAIL_HOST_PORT=
 EMAIL_FROM=
+PORT=_integer_
```
schuyler1d, line 49: Does the `_integer_` etc provide info to some IDE?  If so, this is cool.  However, if it's just a textual hint, I'd rather they be blank in this file, so when people copy it over, if they don't modify the line at all, then it stays an empty value (and e.g. tests if the value exist will do the right thing)

schuyler1d (changes requested):
changes just to make sure it's truly optional

schuyler1d:

Frydafly:
Closing since this is pretty old, and created an issue https://github.com/MoveOnOrg/Spoke/issues/1951 in order to not abandon this project!

PR #368, opened 2017-11-03
added log table for twilio service response and timetamps to message …

benmort:
…table for latency tracking

### DIFF
--- a/src/migrations/index.js
+++ b/src/migrations/index.js
@@ -46,7 +46,28 @@ const migrations = [
        table.string('reason_code').nullable().default('')
      })
    }
-  }
+  },
+  { auto: true, //4
+    date: '2017-10-23',
+    migrate: async function migrate() {
+      await r.knex.schema.alterTable('message', (table) => {
+        table.timestamp('queued_at');
+        table.timestamp('sent_at');
+        table.timestamp('service_response_at');
+      })
+      console.log('added action timestamp columns to message')
+    }
+  },
+  { auto: true, //5
+    date: '2017-10-23',
+    migrate: async function migrate() {
+      await r.knex.schema.createTable('log', (table) => {
+        table.string('message_sid');
+        table.json('body');
+        table.timestamp('created_at').default('now()');
+      })
+      console.log('added log table')
+    }
   /* migration template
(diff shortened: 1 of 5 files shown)


src/migrations/index.js
```diff
@@ -46,7 +46,28 @@ const migrations = [
...
+    date: '2017-10-23',
+    migrate: async function migrate() {
+      await r.knex.schema.createTable('log', (table) => {
+        table.string('message_sid');
+        table.json('body');
+        table.timestamp('created_at').default('now()');
+      })
+      console.log('added log table')
+    }
   /* migration template
```
benmort: need }

src/server/api/lib/twilio.js
```diff
@@ -159,14 +160,15 @@ async function sendMessage(message) {
 async function handleDeliveryReport(report) {
   const messageSid = report.MessageSid
   if (messageSid) {
+    await Log.save({ message_sid: report.MessageSid, body: report })
```
schuyler1d, line 163: will this work on non-postgres (i.e. non JSON-supporting) dbs?  I think we might be safer if `body: JSON.stringify(report)`

src/migrations/index.js
```diff
@@ -46,6 +46,28 @@ const migrations = [
...
+        table.timestamp('sent_at');
+        table.timestamp('service_response_at');
+      })
+      console.log('added action timestamp columns to message')
+    }
+  },
+  { auto: true, //5
+    date: '2017-10-23',
+    migrate: async function migrate() {
+      await r.knex.schema.createTable('log', (table) => {
```
schuyler1d, line 64: I'm a little nervous about a race condition here -- `createTable` will also end up being run in your line if server/models/log.js:
`const Log = thinky.createModel('log', type.object().schema({`
Maybe just do `r.knex.schema.createTableIfNotExists` instead of `createTable` and we should be fine.

schuyler1d (changes requested):
needs tweaks, but very close.

jmcarp:
Just curious--why keep these logs in a sql table? Hopefully operators of spoke will send application logs to a heroku logging provider (or equivalent for aws), so why not output these logs to stdout like everything else?

PR #369, opened 2017-11-03
add conditional join to campaign link from texters form interface

benmort:


src/server/api/schema.js
```diff
@@ -444,6 +444,26 @@ const rootMutations = {
       }
       return organization
     },
+    assignUserToCampaign: async (_, { campaignId }, { user, loaders }) => {
+      let campaign
```
schuyler1d, line 448: Joining an organization requires the uuid, which is at some level the security layer for this api.  So, besides `campaignId` we need an argument `organizationUuid` (just like with joinOrganization) which should be tested against the campaign's organization.

src/server/api/schema.js
```diff
@@ -444,6 +444,26 @@ const rootMutations = {
...
+        const assignment = await r.table('assignment')
+          .getAll(user.id, { index: 'user_id' })
+          .filter({ campaign_id: campaign.id })
+          .limit(1)(0)
+          .default(null)
+        if (!assignment) {
+          await Assignment.save({
+            user_id: user.id,
+            campaign_id: campaign.id,
+            max_contacts: (process.env.DEFAULT_MAX_CONTACTS || 1)
```
schuyler1d, line 461: It looks like `max_contacts` is a new field which we either need to add in this PR, or we can remove this line and include in a max_contacts PR separately.

schuyler1d (changes requested):
some small changes, but besides that, looks good!

PR #371, opened 2017-11-03
message char limit

benmort:


src/server/api/schema.js
```diff
@@ -664,6 +664,13 @@ const rootMutations = {
 
       const { contactNumber, text } = message
 
+      if (text.length > (process.env.MAX_MESSAGE_LENGTH || 280)) {
```
schuyler1d, line 667: For backwards compatibility, please test only IF there is a process.env.MAX_MESSAGE_LIMIT, i.e.
`if (process.env.MAX_MESSAGE_LIMIT && text.length > process.env.MAX_MESSAGE_LENGTH) {`

src/containers/AssignmentTexterContact.jsx
```diff
@@ -513,7 +513,7 @@ class AssignmentTexterContact extends React.Component {
   }
 
   messageSchema = yup.object({
-    messageText: yup.string().required("Can't send empty message")
+    messageText: yup.string().required("Can't send empty message").max(window.MAX_MESSAGE_LENGTH)
```
schuyler1d, line 516: maybe something like `max(window.MAX_MESSAGE_LENGTH || 99999)`

src/server/middleware/render-index.js
```diff
@@ -68,6 +68,7 @@ export default function renderIndex(html, css, assetMap, store) {
       window.SUPPRESS_SELF_INVITE="${process.env.SUPPRESS_SELF_INVITE}"
       window.PRIVACY_URL="${process.env.PRIVACY_URL}"
       window.ALLOW_SEND_ALL=${process.env.ALLOW_SEND_ALL}
+      window.MAX_MESSAGE_LENGTH=${process.env.MAX_MESSAGE_LENGTH || 280}
```
schuyler1d, line 71: let's make it fallback to something far bigger, e.g. `99999`

schuyler1d (changes requested):
Twilio converts to MMS, so we haven't been including a max limit and it's been fine, so the changes requested are to allow a message limit without forcing a group to have one. (fwiw, our biggest message was around ~600chars)

Issue #379, opened 2017-11-08
Opt Out Count on Campaign Stats View Returns GraphQL Int Error


shakalee14:
I receive this error when I visit this screen and opt-outs number doesn't show: 
GraphQL error: Int cannot represent non 32-bit signed integer value [object Object]
![screen shot 2017-11-08 at 11 38 58 am](https://user-images.githubusercontent.com/16676323/32572378-58553a68-c498-11e7-9ba7-6014f7e34fb1.png)

PR #383, opened 2017-11-09
Resolves count result to return an integer

shakalee14:


shakalee14:
Fixes https://github.com/MoveOnOrg/Spoke/issues/379
@benmort

schuyler1d (approved):
:+1: 
knex's habit of returning row results for everything is a bit annoying, and we should probably decide on a preferred pattern for these.  knex does have a `.first()`  for single row returns, but here it would still need a `.then((res) => res.count)` so not much gained.

Anyway, something to think about.

PR #387, opened 2017-11-10
Decreases row max to 6 and removes fixed height

shakalee14:
- handles regression introduced by https://github.com/MoveOnOrg/Spoke/commit/c13df2233b786cead4ef2f9a0dacb6bc01f43ee6


schuyler1d (approved):
If this is accomodating specific devices then can we add a comment by the rules -- that way we can avoid oscilating between fixes across different device issues.
:+1:

PR #360, opened 2017-10-27
moveon master > Getup dev

benmort:

PR #398, opened 2017-11-18
[WIP] Lint on travis.

jmcarp:
There's a lot of lint to clean up, probably because we're not running `eslint` on travis. This patch adds `eslint` to our travis config, checks in the automatic fixes made by `eslint`, and begins to resolve other lint errors. This is dull work but potentially useful--the linter is already uncovering unused code and missing variables, so I'm guessing we have some lurking bugs here.

Before I spend much time fixing errors, can we confirm that the current `eslint` rules are what we want to use @schuyler1d ?

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -1,6 +1,6 @@
 import { r, datawarehouse, Assignment, Campaign, CampaignContact, InteractionStep, Organization, User } from '../server/models'
 import { log, gunzip, zipToTimeZone } from '../lib'
-import { sleep, getNextJob, updateJob } from './lib'
+import { updateJob } from './lib'
 import serviceMap from '../server/api/lib/services'
 import { getLastMessage, saveNewIncomingMessage } from '../server/api/lib/message-sending'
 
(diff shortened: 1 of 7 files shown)


src/workers/jobs.js
```diff
@@ -58,22 +52,22 @@ export async function processSqsMessages() {
...
-          if (err) console.log( err, err.stack ) // an error occurred
-          else     console.log( data )           // successful response
+  sqs.receiveMessage(params, async (err, data) => {
+    if (err) {
+      console.log(err, err.stack)
+      p.reject(err)
+    } else if (data.Messages) {
+      console.log(data)
+      for (let i = 0; i < data.Messages.length; i ++) {
+        const body = data.Messages[i].Body
```
schuyler1d, line 62: @shakalee14 This may explain our sqs not working as expected.

schuyler1d (commented):
@jmcarp I think we can take this PR, since it's hitting a file that 1. does need to be fixed for a current issue we're having, and 2. I think is off the mainline work of GetUp (see https://github.com/MoveOnOrg/Spoke/pull/360 and the other benmort PRs).

However, I think we should wait before running this on other files, because it will make merging the GetUp work that much more difficult.

jmcarp:
@schuyler1d: as written, this patch is all-or-none, since it adds `eslint` to travis. If we only fix some linter errors, builds will fail. If you want, I can back out the change to the travis config and leave the updates to the worker as they are. WDYT?

Issue #391, opened 2017-11-13
Replace deprecated `apollo-server` library


jmcarp:
It's been replaced with https://www.npmjs.com/package/apollo-server-express.

schuyler1d:
@jmcarp This is described here:
https://github.com/MoveOnOrg/Spoke/issues/224
Because it would take some time to figure out how to refactor/make all the changes for the new server, we have been procrastinating this.
If you'd like to take this on, let us know.  It would certainly be good to be up-to-date.

Issue #229, opened 2017-09-14
Remove Apollo tracer


sreynen:
Related to https://github.com/MoveOnOrg/Spoke/issues/224, currently including https://www.npmjs.com/package/apollo-tracer which is marked as deprecated on https://github.com/apollographql/graphql-tracer. The name and use in /server/index.js suggests it's a debugging tool, so removing shouldn't impact anything for production code.

PR #399, opened 2017-11-19
Update to latest apollo-server and handle breaking changes.

### LINKED ISSUE #224: Update Apollo GraphQl
Old (backwards-incompatible) version of Apollo running.  Upgrade all the things to make it work.

This is probably difficult and will take a good amount of work across the code base, both client and server

jmcarp:
[Resolves #224 #229 #391]

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,8 +60,7 @@
   "dependencies": {
     "aphrodite": "^0.4.1",
     "apollo-client": "^0.4.7",
-    "apollo-server": "^0.1.5",
-    "apollo-tracer": "^0.1.4",
+    "apollo-server-express": "^1.2.0",
     "aws-sdk": "^2.6.3",
     "aws-serverless-express": "https://registry.npmjs.org/aws-serverless-express/-/aws-serverless-express-3.0.2.tgz",
     "axios": "^0.16.1",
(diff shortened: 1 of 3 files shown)


shakalee14:
When testing this branch I get the following errors on campaign creation:
- After attempting to submit by clicking `save and goto next section`, I receive an error in the UX/Client: `There was an error with your request. Try again in a little bit!`
Server error: ```3:47:32 PM server.1               |  Error: Resolve function for "Campaign.datawarehouseAvailable" returned undefined```
- Then I refresh the screen, and data for that portion of the campaign creation form has been saved and I can continue

@jmcarp - did you see this same behavior?

example screen shots:
Error message:
screen shot 2017-11-20 at 3 47 55 pm

After refreshing, I am able to click next part of campaign creation form:
screen shot 2017-11-20 at 3 51 22 pm

jmcarp:
Good catch @shakalee14! Looks like the `datawarehouseAvailable` resolver returns `undefined` instead of `false` when the user is a super-admin and the data warehouse is not available. I added a small fix so that the resolver returns `false` in this case. Ready for review when you have time!

shakalee14 (approved):

PR #404, opened 2017-11-24
Migrate prop-types to external package.

jmcarp:
`React.PropTypes` is deprecated in react 15.6 and gone in 16.x. Migrated to external package as described in https://reactjs.org/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes.


shakalee14:
i'll test this branch - but I agree and we already utilize the `prop-types` library with our other react projects.

shakalee14 (approved):

PR #370, opened 2017-11-03
small styling additions to campaigns on todos

benmort:

PR #407, opened 2017-11-27
Merge GetUp changes with generalizations/fixes/configuration

schuyler1d:
This is a followup to @benmort 's #360 PR.
Incidentally, it also merges in #371 and #368 
While not merging  #369 or #370 it includes those changes from the #360 'big patch'

It does roughly the following:
1. adds a bunch of style options for campaigns (color, logo, html pre-text)
2. dynamic assignment mode, where people can join a campaign and begin texting immediately, and will be replenished with a new contact as they work them (with a configurable maximum contact value `MAX_CONTACTS_PER_TEXTER` or what is set by the campaign admin for the user)
3. An optional `TERMS_REQUIRE` page when the user first logs in.  At the moment, this text is hard-coded, but might also be a good point for configuration in the future.
4. `NOT_IN_USA` configuration option, required to use any bulk sending feature
5. An hierarchical script editing interface, which also works post-campaign start.
6. Some other [miscellany](https://github.com/MoveOnOrg/Spoke/pull/360#pullrequestreview-73176506) (e.g. logging twilio message statuses better)

My hope with this, even if it's a big patch is that we can QA it this week and if it's tested successfully, we can merge before the hackathon next week, so we can clean up our backlog of PRs.


jmcarp:
Since this includes #368, here's my question from that patch: https://github.com/MoveOnOrg/Spoke/pull/368#issuecomment-346707561

> Just curious--why keep these logs in a sql table? Hopefully operators of spoke will send application logs to a heroku logging provider (or equivalent for aws), so why not output these logs to stdout like everything else?

schuyler1d:
I commented the same thing on their original! That might be worth doing later, but it's worth noting that in this case it's just moving raw message record data away from message.response_text to the log table. So it's no worse than before. This raw response data may have other info useful for programmatically handling the message (e.g. flagging unreachable numbers in some way).

src/server/api/interaction-step.js
```diff
@@ -14,8 +19,19 @@ export const resolvers = {
...
       'id',
-      'script'
+      'script',
+      'answerOption',
+      'answerActions',
+      'parentInteractionId',
+      'question',
+      'isDeleted'
     ], InteractionStep),
+    questionText: async(interactionStep) => {
```
shakalee14, line 29: When I run this, using dynamic assignment, I get the following error on the texter contact screen:

Client:
```
minilog.js:17 client  
ApolloError {graphQLErrors: Array(1), networkError: TypeError: Cannot read property 'interactionSteps' of undefined at new AssignmentTexterContact (…, message: "GraphQL error: Cannot read property '0' of undefin…not read property 'interactionSteps' of undefined"}
graphQLErrors
:
[{…}]
message
:
"GraphQL error: Cannot read property '0' of undefined↵Network error: Cannot read property 'interactionSteps' of undefined"
networkError
:
TypeError: Cannot read property 'interactionSteps' of undefined at new AssignmentTexterContact (http://localhost:3000/assets/bundle.js:211469:83) at ReactCompositeComponentWrapper._constructComponentWithoutOwner (http://localhost:3000/assets/bundle.js:23568:28) at ReactCompositeComponentWrapper._constructComponent (http://localhost:3000/assets/bundle.js:23550:22) at ReactCompositeComponentWrapper.mountComponent (http://localhost:3000/assets/bundle.js:23469:22) at Object.mountComponent (http://localhost:3000/assets/bundle.js:21841:36) at ReactCompositeComponentWrapper.performInitialMount (http://localhost:3000/assets/bundle.js:23658:35) at ReactCompositeComponentWrapper.mountComponent (http://localhost:3000/assets/bundle.js:23532:22) at Object.mountComponent (http://localhost:3000/assets/bundle.js:21841:36) at ReactCompositeComponentWrapper.performInitialMount (http://localhost:3000/assets/bundle.js:23658:35) at ReactCompositeComponentWrapper.mountComponent (http://localhost:3000/assets/bundle.js:23532:22)
__proto__
:
Error
```

Server:
```
3:45:06 PM server.1               |  TypeError: Cannot read property '0' of undefined
3:45:06 PM server.1               |      at getTopMostParent (src/lib/interaction-step-helpers.js:90:10)
3:45:06 PM server.1               |      at _callee7$ (src/server/api/campaign-contact.js:168:14)
3:45:06 PM server.1               |      at tryCatch (/Users/shaka/moveon/Spoke/node_modules/regenerator-runtime/runtime.js:62:40)
3:45:06 PM server.1               |      at GeneratorFunctionPrototype.invoke [as _invoke] (/Users/shaka/moveon/Spoke/node_modules/regenerator-runtime/run
3:45:06 PM server.1               |  >  time.js:336:22)
```

When I add some console.logs, it looks like `allInteractionSteps` here isn't updating in the getAvailableInteractionSteps() function in the `AssignmentTexterContact` component. It remains an empty array which then get's passed to  `getTopMostParent` helper function.

jmcarp:
This could be a good place to work on test coverage. As we fix bugs like the one @shakalee14 reported above, we can fill in regression tests so that we don't let those bugs come up again. I'd be happy to help or pair on that if you all are interested.

src/server/api/interaction-step.js
schuyler1d: I believe this will be fixed with 637c8e5

PR #408, opened 2017-11-27
Readme tweaks

schuyler1d:
improvements to README setup instructions from @laveshaparker feedback

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,14 +33,14 @@ Follow up instructions located [here](https://github.com/MoveOnOrg/Spoke/blob/ma
 9. Run `npm run dev` to start the app. Wait until you see both "Node app is running ..." and "Webpack dev server is now running ..." before attempting to connect. (make sure environment variable `JOBS_SAME_PROCESS=1`)
 10. Go to `http://localhost:3000` to load the app.
 11. As long as you leave `SUPPRESS_SELF_INVITE=` blank and unset in your `.env` you should be able to invite yourself from the homepage.
-If you set that variable, then spoke will be invite-only and you'll need to generate an invite. Run:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,18 +27,20 @@ Follow up instructions located [here](https://github.com/MoveOnOrg/Spoke/blob/ma
 3. `npm install`
 4. `npm install -g foreman`
 5. `cp .env.example .env`
-6. Run `npm run dev` to create and populate the tables.
-7. Create an [Auth0](https://auth0.com) account. In your Auth0 account, go to Settings -> Clients -> and then grab your Client ID, Client Secret, and your Auth0 domain (should look like xxx.auth0.com). Add those inside your `.env` file (AUTH0_CLIENT_ID, AUTH0_CLIENT_SECRET, AUTH0_DOMAIN respectively).
+6. Create an [Auth0](https://auth0.com) account. In your Auth0 account, go to Settings -> Clients -> and then grab your Client ID, Client Secret, and your Auth0 domain (should look like xxx.auth0.com). Add those inside your `.env` file (AUTH0_CLIENT_ID, AUTH0_CLIENT_SECRET, AUTH0_DOMAIN respectively).
```
jmcarp, line 30: Can we use markdown-style numbering and number every bullet point as `1.`? The items will render the same way, but we won't have to spend time renumbering if we insert or delete an item early in the list.

schuyler1d: I personally like the numbering in the file, because then it functions as just a text file as well (I generally read the readme file from the checkout, not the github page).

The renumbering here is because we're flipping the order, not messing with ordering accidents.

shakalee14 (approved):

PR #390, opened 2017-11-13
Drop deprecated babel presets.

jmcarp:

PR #414, opened 2017-11-29
Removes babyparse and changes to papaparse

shakalee14:


jmcarp:
This is also happening in https://github.com/MoveOnOrg/Spoke/pull/390/commits/15bad57b343ccecff52746c61ca747591a79f46f, within #390.

PR #418, opened 2017-12-01
[WIP] Update apollo client and react bindings to latest.

jmcarp:
The `apollo-client` and `react-apollo` libraries have both changed significantly since this project was started. To continue getting fixes and improvements, we'll need to upgrade. Upgrading will also help new contributors figure out how the code works--it's confusing to read the `react-apollo` docs and see that their examples don't resemble our code at all.

This is a pretty early work in progress--most queries for the admin interface are working, but mutations and other pieces of the app aren't there yet. I'm submitting in an incomplete state for feedback, and in case anybody has time to jump in.

cc @schuyler1d @shakalee14


schuyler1d:
Wow, super cool spike!
My guess is we're going to need to resolve crazy enough merges as this evolves, that I would make a couple of thoughts:
1. Can you document particularly tricky parts of the code?  Maybe by filename? (see 4. below)
2. Can you describe the more mechanical aspects (listing e.g. "convert client calls like X")
3. It seems in a couple places you rename a variable (organizationData -> organization).  I would suggest in this kind of patchset that you want to avoid that so we stick to just the mechanical parts as much as possible.
4. YMMV, but when we had to rip out rethinkdb => knex, I had a doc that tracked what parts were complete: https://github.com/MoveOnOrg/Spoke/blob/2964cb0a2fcf4a9a637ba63aab9a873a40757c0e/docs/DB_MIGRATION.md (you can look at the history of that file) -- I think that helped me and others track the scope of work, along with places we could document what happened with particular files.

schuyler1d:

schuyler1d:
This was completed with https://github.com/MoveOnOrg/Spoke/pull/1581

PR #424, opened 2017-12-03
Twilio instructions update

shakalee14:


schuyler1d (approved):
:+1: 
Would be good to link this from both heroku instructions (instead of duplilcating?) and also the README

Issue #266, opened 2017-09-28
Allow multimedia images in texts and scripts


schuyler1d:
Create the ability to send MMS messages with media in them.

The easiest way might be to allow a markdown-ish format in script text like `[<img url>]` which gets converted to being sent with the MMS.

More complicated options:
* A more sophisticated interface would allow image upload, etc.  
* UI could also include embedding the message, at least as a thumb or href for the Texter post-sending

shakalee14:
trying easy implementation first.

shakalee14:
Didn't work on this yet but will assign myself if we decided to focus on this issue in upcoming milestone.

lperson:
I'm going to work on this.

shakalee14:
@lperson 💯

ben-pr-p:
Me too!

shakalee14:
added with https://github.com/MoveOnOrg/Spoke/pull/425

PR #425, opened 2017-12-03
Support for multimedia (#266)

ben-pr-p:
Hey!

This adds support for images in texts. The syntax is, in any message, `[https://image-url.com/image-is-here]`.

screen shot 2017-12-03 at 2 19 54 pm

And the result:

img_0122

### DIFF
--- a/src/server/api/lib/twilio.js
+++ b/src/server/api/lib/twilio.js
@@ -111,11 +111,11 @@ async function sendMessage(message) {
     }
 
     // Image extraction
-    const extractor = new RegExp(/\[.*http(.*)\]/)
+    const extractor = new RegExp(/\[\s*(http[^\]\s]*)\s*\]/)
     let mediaUrl
     if (message.text.match(extractor)) {
       const results = extractor.exec(message.text)
-      mediaUrl = `http${results[1].trim()}`
+      mediaUrl = results[1]
     }
 
     const messageParams = Object.assign({
--- a/src/server/api/lib/twilio.js
+++ b/src/server/api/lib/twilio.js
@@ -109,15 +109,26 @@ async function sendMessage(message) {
     if (message.service !== 'twilio') {
       log.warn('Message not marked as a twilio message', message.id)
     }
-    const messageParams = {
+
+    // Image extraction
+    const extractor = new RegExp(/\[.*http(.*)\]/)
+    let mediaUrl
+    if (message.text.match(extractor)) {
+      const results = extractor.exec(message.text)
+      mediaUrl = `http${results[1].trim()}`
+    }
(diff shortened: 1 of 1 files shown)


src/server/api/lib/twilio.js
```diff
@@ -109,15 +109,26 @@ async function sendMessage(message) {
     if (message.service !== 'twilio') {
       log.warn('Message not marked as a twilio message', message.id)
     }
-    const messageParams = {
+
+    // Image extraction
+    const extractor = new RegExp(/\[.*http(.*)\]/)
```
schuyler1d: I think the first `.*` should be `\s*` and the second you probably want `([^\]]*)` to be a little safer

schuyler1d (changes requested):
:+1: 
after we tweak the regex.  This is great!

schuyler1d (approved):

src/server/api/lib/twilio.js
```diff
@@ -109,15 +109,26 @@ async function sendMessage(message) {
     if (message.service !== 'twilio') {
       log.warn('Message not marked as a twilio message', message.id)
     }
-    const messageParams = {
+
+    // Image extraction
+    const extractor = new RegExp(/\[\s*(http[^\]\s]*)\s*\]/)
```
jmcarp, line 114: WDYT about moving this to a separate function and adding tests?

shakalee14:
this is ready to be merged and was qa'd on stage

joemcl:
I'm trying to send a small test campaign with an image - not working so far. 
The number I'm sending is MMS-enabled in Twiio, just checked. 
The Twilio message log  shows an unhelpful error message - (Error:30008)Unknown error. 
The link I'm using is [https://drive.google.com/file/d/1i6X8OmTqCLEdbhTxXHbnaZFvmTZ7StNQ/view?usp=sharing], that's OK, just tried that in an 'incognito window' . 
Checking if there's some Heroku setting I need to change/tweak. 
I'm using our staging Spoke instance, it works fine for test campaigns sending plain text.

joemcl:
Thanks @bchrobot *shouldn't* matter - the link and image is public - but of course it might :) cause  Google. So I'll check out imgur.com. In the test above, is that i.ytimg.com link produced by imgur.com or, like, YouTube? Some other site?

joemcl:
OK @bchrobot thanks, so this link worked, hosted on imgur - https://i.imgur.com/lv66Kzg.jpg - a 'Direct Link'. 
An imgur 'Image Link' like https://imgur.com/lv66Kzg (does not contain file type suffix .jpg) did not. 
Thanks!

Issue #434, opened 2017-12-14
Spoke buttons not appearing on iPhone 6/6s and Galaxy devices


ErikKirkstein:
Post deploy to prod on 12/14, buttons on bottom of the screen are no longer appearing

ErikKirkstein:
It appears some are able to see the buttons by minimizing their screen

schuyler1d:
![image uploaded from ios 6](https://user-images.githubusercontent.com/52117/34013953-69173f0c-e0e7-11e7-9070-a4da7c105815.jpg)

shakalee14:
fixed and confirmed on production with commit: https://github.com/MoveOnOrg/Spoke/commit/62ed4a3c68604a0ad11a4f844445c9e1417ada7f

PR #436, opened 2017-12-14
Decreases rowmax size and margin top for picture to fix button pushin…

shakalee14:
…g down issue on mobile views

for #434

### DIFF
--- a/src/components/Empty.jsx
+++ b/src/components/Empty.jsx
@@ -11,9 +11,11 @@ const inlineStyles = {
   }
 }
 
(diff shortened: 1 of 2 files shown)


shakalee14:
testing with same script in issue screen shot, is see this locally (with a scrollable text field)

screen shot 2017-12-14 at 5 09 11 pm

hiemanshu (approved):
Changes look good to me. I tested it with a few devices (using the chrome simulator), with super long messages, and it looked good.

PR #437, opened 2017-12-15
[WIP] Lint redux

jmcarp:
Down to <100 errors so far 🎉. @schuyler1d and @shakalee14: feel free to push to this branch and help get errors to zero. I'll try to finish whatever's left over the weekend.

### DIFF
--- a/src/lib/zip-format.js
+++ b/src/lib/zip-format.js
@@ -65,3 +65,24 @@ export const zipToTimeZone = (zip) => {
   }
   return null
 }
+
+// export const findZipRanges = function (r) {
+//   var zipchanges = []
+//   return r.knex('zip_code').select('zip', 'timezone_offset', 'has_dst')
+//     .orderBy('zip').then(function (zips) {
+//       var front = -1
+//       var curTz = -4
+//       var curHasDst = -1
+//       zips.forEach((zipRec) => {
+//         if (zipRec.timezone_offset != curTz || zipRec.has_dst != curHasDst) {
+//           zipchanges.push([front, parseInt(zipRec.zip), curTz, curHasDst, parseInt(zipRec.zip) - front])
+//           curTz = zipRec.timezone_offset
+//           curHasDst = zipRec.has_dst
+//           front = parseInt(zipRec.zip)
+//         }
+//       })
+//       zipchanges.sort(function (a, b) { return b[4] - a[4] })
+//       console.log(zipchanges)
+//     })
+//   return zipchanges
+// }
--- a/src/lib/zip-format.js
+++ b/src/lib/zip-format.js
@@ -56,11 +56,12 @@ const commonZipRanges = [
 
 commonZipRanges.sort((a, b) => (a[0] - b[0]))
 
-export const zipToTimeZone = (zip, r) => {
+export const zipToTimeZone = (zip) => {
   // will search common zip ranges -- won't necessarily find something
   // so fallback on looking it up in db
   if (typeof zip === 'number' || zip.length >= 5) {
-    zip = parseInt(zip)
-    return commonZipRanges.find((g) => (zip >= g[0] && zip < g[1]))
+    const parsed = parseInt(zip, 10)
+    return commonZipRanges.find((g) => (parsed >= g[0] && parsed < g[1]))
   }
+  return null
 }
--- a/src/lib/zip-format.js
+++ b/src/lib/zip-format.js
@@ -4,12 +4,11 @@ export const getFormattedZip = (zip, country = 'US') => {
     const [, first5] = zip.match(regex) || []
 
     return first5
-  } else {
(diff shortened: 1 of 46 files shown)


ben-pr-p:
I highly recommend you all use `prettier`, optionally with eslint on top: https://github.com/prettier/prettier-eslint.

Basically, it parses your JS (and JSX) and then spits back out JS from the AST. It not only makes keeping up linter compliance super easy, but personally it reduces my development time as well, because I just get to mash things and then auto format after

jmcarp:
This is ready for review @schuyler1d @shakalee14.

@ben-pr-p: agreed that `prettier` would be useful here, but I tried it out and found that it didn't fix many of the linter errors. Let's add `prettier` in a separate patch.

src/lib/zip-format.js
```diff
@@ -57,32 +56,12 @@ var commonZipRanges = [
...
-  if (typeof zip == 'number' || zip.length >= 5) {
-    zip = parseInt(zip)
-    return commonZipRanges.find((g) => (zip >= g[0] && zip < g[1]))
+  if (typeof zip === 'number' || zip.length >= 5) {
+    const parsed = parseInt(zip, 10)
+    return commonZipRanges.find((g) => (parsed >= g[0] && parsed < g[1]))
   }
-}
-
-export const findZipRanges = function (r) {
```
schuyler1d: This isn't used in code, but it's how to generate the data higher up in `commonZipRanges` so let's not delete, please.

src/server/action_handlers/test-action.js
```diff
@@ -16,7 +15,7 @@ export const instructions = () => (
 // either in environment variables or organization.features json data
 // Besides this returning true, "test-action" will also need to be added to
 // process.env.ACTION_HANDLERS
-export async function available(organizationId) {
+export async function available() {
```
schuyler1d: The purpose of this file is to expose the arguments sent -- it doesn't matter that it's not used.  please add back.

src/server/api/lib/fakeservice.js
```diff
@@ -28,15 +27,15 @@ async function convertMessagePartsToMessage(messageParts) {
...
+  const serviceId = (firstPart.service_id
                       || `fakeservice_${Math.random().toString(36).replace(/[^a-zA-Z1-9]+/g, '')}`)
   return new Message({
     contact_number: contactNumber,
     user_number: userNumber,
     is_from_contact: true,
     text,
     service_response: JSON.stringify(messageParts),
-    service_id,
+    serviceId,
```
schuyler1d: You can't change the key of the value sent or it won't be saved to the right field.

schuyler1d (changes requested):
This is mostly ok.

My big concern beyond the couple of nits I found is about interactionStep ids.  It's possible we've moved past this, but there's possibly a real discrepancy between string/number types needed for the backend vs what's sent, so anything where you changed `==` => `===` around interaction step ids, we should either test carefully or guard with `parseInt` wrappers.

src/lib/zip-format.js
jmcarp: I'm not sure this function works as intended--the final `return` is unreachable, so `zipchanges` gets populated and then discarded. I'm restoring this function but commenting it out--we can uncomment and fix it separately.

jmcarp:
Updated. I think I'm missing some context around interaction step IDs--why would some be strings and others numbers? Anyway, it's probably better to get to the bottom of the issue, or explicitly parse to numbers, than to use `==`.

schuyler1d:
the interaction step id thing is that during the rethinkdb period all ids were strings.  Migrating to knex/postgres made ids integers, and a good part of the interaction step processing on the *backend* was rewritten, but just adapting the frontend string ids, because we were leaving the frontend implementation alone.

Then last month we merged what was more-or-less a rewrite on the frontend from GetUp.  They *might* have changed ids to be properly integers, but maybe not, and I haven't dug in.

I'm all for getting to the bottom of the issue -- the point is until we *do* get to the bottom of it (or do qa/testing), we can't merge this PR.

jmcarp:
Sounds good--I restored the `==` comparisons for interaction steps in https://github.com/MoveOnOrg/Spoke/pull/437/commits/6c5f3eeb54e628393b3d0aa34a7d941bd4508728. Let me know if you want anything else reverted, or feel free to commit to this branch.

schuyler1d:
We QA'd this morning and found some issues.  We haven't had time to trace them yet, but reporting here:
1. When I start a campaign and enter a section (e.g. contacts), I get the spinner right away (rather than when I save it)
2. When I click send on a message, I get an error in the client:
```
extraInfo: undefinedgraphQLErrors: nullmessage: "Network error: Cannot read property 'forEach' of undefined"networkError: TypeError: Cannot read property 'forEach' of undefined
```
though sending at least sometimes works.

schuyler1d:

PR #439, opened 2017-12-15
tweak message sending and add some tests

schuyler1d:


shakalee14:
interesting @schuyler1d - looks like code climate says there is 1 issue to fix?

shakalee14 (commented):
Where is this disabling mms sending? After a round of QAing, I'm still able to send images via text locally.

hiemanshu:
@schuyler1d How about adding a test case for where there are two URLs given, and only the first one is selected? Otherwise this looks good.

schuyler1d:
ok added @hiemanshu 's test.  @shakalee14 I think we should just ignore this code climate 'issue'
I tried to refactor a bit, but it's not really giving me anything actionable to address "Function messageSenderCreator has a Cognitive Complexity of 8 (exceeds 5 allowed). Consider refactoring"

schuyler1d:
Also, this does keep media parsing working -- one commit disables them, but then a latter commit enables them again (with a bit refactor to make it super-work) and adds tests to make sure it stays super-working.

shakalee14 (approved):

PR #440, opened 2017-12-16
Removes image for mobile devices

shakalee14:


shakalee14:
i suggest we remove the image from mobile devices.

hiemanshu:
The size values used [here ](https://github.com/MoveOnOrg/Spoke/blob/7a4500f63c96e6949133aa274c8baab508516f32/src/containers/AssignmentTexterContact.jsx#L596) is `< 450px`. Can we use the same size for this? I'd also prefer if we extracted it out to a variable, and it isn't a magic number in code. Otherwise looks like a good idea to me!

schuyler1d:
If we do this or something else can we get our 6 lines back?

Also consider this as alternative:
https://github.com/MoveOnOrg/Spoke/commit/73f4787781182d6c38e18cb03ebce6fc4f6b7bee

Issue #444, opened 2017-12-19
"You have nothing to do!" message no longer appears on mobile devices


ErikKirkstein:
The "You have nothing to do!" message no longer appears when texting/replies are complete on mobile. Instead texters just see a blank screen. Image below for iOS device.
![image uploaded from ios 1](https://user-images.githubusercontent.com/32142717/34165504-790bdf00-e4aa-11e7-8ad7-bf489a073575.png)

PR #445, opened 2017-12-19
444 restore nothing to do message

shakalee14:
- For #444 
- When we removed the pencil from mobile, we also removed the 'nothing to do message' on the texter dashboard. This refactors the Empty.jsx component to acct for both cases.


hiemanshu:
@shakalee14 because of other cases where this might happen again, I'd do the reverse. I'd check if it has to be hidden, otherwise display it. So its hidden in only one case, and shown in the others.

hiemanshu (approved):

Issue #452, opened 2018-01-02
Spoke skipping texts; sending every other initial text instead of each


ErikKirkstein:
After assigning a texter X amount of texts, when clicking send to send the initial texts, Spoke is skipping over odd numbered texts.

Ex: You have 50 texts, you enter the user screen. You click "Send" and according to the counter, instead of sending 1/50. 2/50. 3/50, 4/50, 5/50 etc, Spoke is sending 2/50, 4/50, 6/50 etc.

The skipped over texts then appear as new texts to send in the to-do page.

shakalee14:
This bug isn't necessarily even or odd - i believe it is just skipping. I can reproduce this issue locally.

PR #453, opened 2018-01-04
Separates logic out between navigating next after skipping and sendin…

shakalee14:
…g message

- not sure why this logic can't stay within the same method. happy for ideas

### DIFF
--- a/src/components/AssignmentTexter.jsx
+++ b/src/components/AssignmentTexter.jsx
@@ -71,14 +71,23 @@ class AssignmentTexter extends React.Component {
 
   handleFinishContact = () => {
     if (this.hasNext()) {
-      this.handleNavigateNext()
+      this.handleNavigateNextforSend()
     } else {
       // Will look async and then redirect to todo page if not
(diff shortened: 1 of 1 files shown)


schuyler1d:
My guess is there's a race condition somewhere with the contact index.
If you need to merge this before getting to the bottom of it, please add some comments with a link to the github issue and explain what you're guarding against in the code.

shakalee14:
Issue described here: https://github.com/MoveOnOrg/Spoke/issues/452
Ex: You have 50 texts, you enter the user screen. You click "Send" and according to the counter, instead of sending 1/50. 2/50. 3/50, 4/50, 5/50 etc, Spoke is sending 2/50, 4/50, 6/50 etc.

This pr separates the handleNavigateNext logic into 2 methods. If you try to use the same method, skipping will break or sending messages in an incremental order will break.

schuyler1d (approved):
please include a link IN the code to `https://github.com/MoveOnOrg/Spoke/issues/452` so we can trace it back when we DO try to get to the bottom of it.

PR #472, opened 2018-01-22
User data editing

### LINKED ISSUE #320: Allow texters to change their name
Currently changing a texter name requires a change in the database. Would be nice if texters could do this themselves.

At least, via loading/refreshing info from Auth0

schuyler1d:
This is take2 on #334 just to make the diff simpler.

This fixes #320 and also provides the same interface for an organization admin to edit the user info, as well.

Ancillary to that change, this also starts moving towards a second goal -- to make Auth0 optional, butthis needs a SignUp form, but that will also be the same form (with some additional password fields, and interaction with an invite link)


shakalee14 (approved):
Let's QA this branch

PR #473, opened 2018-01-22
Contacts section won't expand when the campaign is started ...

lperson:
Contacts section won't expand when campaign is started; if open when the campaign starts it will close.


lperson:
Closes #469 
@schuyler1d please review, thanks!

schuyler1d (approved):
looks good to me.
The next step is for us to queue it for QA

shakalee14:
Merged with #483

Issue #474, opened 2018-01-22
Prevent addition of new contacts once a campaign starts -- server side


lperson:
Related to #469 
@schuyler1d please assign to me

PR #475, opened 2018-01-23
prevent addition of contacts after campaign starts

lperson:


lperson:
closes #474 
for review @schuyler1d

shakalee14:
This is great - we should qa this on stage but it works locally. I'm wondering if we can add a message in the UI too (vs just server side error log). Let me know what you think about that @lperson

shakalee14 (commented):
It would be nice to also send a message to client/ux on why new upload is 'failing'

lperson:
@shakalee14 Thanks for the review!

The changes in the UI in PR #473 blocks the change on the client side by closing the contacts panel if it's open when the campaign starts, and now allowing it to open if the campaign is already started.  This change is a backstop, and the attempt to upload contacts won't get to the server in most cases.

That having been said, I agree the error should be passed back to the client.  Should we iterate on this PR and update the code?  Or maybe merge this PR and open an issue to address passing the message back?

shakalee14 (approved):

shakalee14:
Merged with #483

PR #477, opened 2018-01-25
Fix bind error

### LINKED ISSUE #464: js error on adding a response in campaign script/interaction creation
When I click on 'Add Response' in interactions, i see a javascript error:
```
bundle.e5cab88a454f1e15acca.js:92 Uncaught TypeError: Cannot read property 'bind' of undefined
    at Object.onTouchTap (bundle.e5cab88a454f1e15acca.js:92)
    at r.handleTouchTap (bundle.e5cab88a454f1e15acca.js:15)
    at Object.r (bundle.e5cab88a454f1e15acca.js:23)
    at a (bundle.e5cab88a454f1e15acca.js:23)
    at Object.s [as executeDispatchesInOrder] (bundle.e5cab88a454f1e15acca.js:23)
    at p (bundle.e5cab88a454f1e15acca.js:23)
    at m (bundle.e5cab88a454f1e15acca.js:23)
    at Array.forEach (<anonymous>)
    at r (bundle.e5cab88a454f1e15acca.js:44)
    at Object.processEventQueue (bundle.e5cab88a454f1e15acca.js:23)
```
( this doesn't seem to affect save behavior, though it concerns me)

jparkrr:
Closes #464

My first contribution! Spoke seems cool, nice job.

`bind()` was attempted to be run on the return value of the method rather than the method itself.

I changed `addStep` and `deleteStep` to return a function that the tap handler can execute instead. This returned function can then be binded/bound/bindeded.

There are multiple ways of doing this. Let me know if you would prefer another style.

### DIFF
--- a/src/components/CampaignInteractionStepsForm.jsx
+++ b/src/components/CampaignInteractionStepsForm.jsx
@@ -72,27 +72,31 @@ export default class CampaignInteractionStepsForm extends React.Component {
   }
 
   addStep(parentInteractionId) {
-    const newId = 'new' + Math.random().toString(36).replace(/[^a-zA-Z1-9]+/g, '')
-    this.setState({
-      interactionSteps: [
-        ...this.state.interactionSteps,
-        { id: newId, parentInteractionId, questionText: '', script: '', answerOption: '', answerActions: '', isDeleted: false }
-      ]
-    })
+    return () => {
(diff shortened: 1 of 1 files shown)


schuyler1d (approved):
Nice catch.
I think we could also do (keeping the previous function):
`this.deleteStep.bind(this, interactionStep.id)` which is probably too clever by half, but shorter.
I think your way will probably be more readable over the long term.

Issue #292, opened 2017-10-04
Integrate React to testing framework


schuyler1d:
In theory we just need to load in `enzyme` and load up some components. I tried a naive version and we might be in version hell with react. We need 15.5 or 16 and we are on 15.6 :-(

schuyler1d:
Some good ones to start with (non-complex) will be:

* src/components/TexterStats.jsx (no state)
* src/components/TopNav.jsx

and some more complex ones that would be great to wire up (because we should have a bunch of tests for it) would be:
* src/components/AssignmentSummary.jsx
* src/components/AssignmentTexter.jsx
* src/components/CampaignBasicsForm.jsx
* src/components/CampaignContactsForm.jsx

schuyler1d:
I'm going to close this since we now have a framework.  We could still use a lot more tests though, so will open a new issue to track that.

PR #478, opened 2018-01-26
wip: Ability to test frontend

jparkrr:
WIP for #292

Everything seems working so far.

Question: I'm not completely happy with the way the `text()` tests look. Should we use [snapshot testing](https://facebook.github.io/jest/docs/en/snapshot-testing.html)? Or do you feel these tests are too brittle overall?

Alternatively I could change the component and make it easier to test by adding html classes. 

@schuyler1d  mentioned some good ones to test in #292. Does it make sense to do all those here? It might take me a few weeks. I don't mind keeping the branch up to date with `master`.

- [x] src/components/TexterStats.jsx
- [x] src/components/TopNav.jsx
- [x] src/components/AssignmentSummary.jsx (in #486)
- [ ] src/components/AssignmentTexter.jsx
- [ ] src/components/CampaignBasicsForm.jsx
- [ ] src/components/CampaignContactsForm.jsx


__test__/TexterStats.test.js
```diff
@@ -0,0 +1,88 @@
...
+    })
+  })
+})
+
+
+describe('TexterStats (Dynamic campaign)', () => {
+  it('contains the right text', () => {
+    const stats = shallow(<TexterStats campaign={campaignDynamic} />)
+    expect(stats.text()).toEqual(
+      'Test45 initial messages sentSomeone541 initial messages sent'
```
jparkrr, line 85: This is the kind of thing that's weird

schuyler1d: I suppose you could do stats.find(...).text() and evaluate the sub elements but until it gets more sophisticated, I think it's fine

schuyler1d:
We might merge this before you get all the way through, but while it's open, feel free to keep adding here.

Thanks so much, this will help enormously over time as we build up the suite.

__test__/TexterStats.test.js
shakalee14: Agree with  @schuyler1d

jparkrr: Thanks for the feedback!

jparkrr:
I found this which talks about ways to test Apollo components https://medium.com/@ferdingler/testing-apollo-react-components-a7618d8a9e7a

Basically they mostly test the unwrapped component (so the component file will have to change slightly), which sounds good to me.

Any thoughts?

schuyler1d:
The pattern that we have for [another project](https://github.com/MoveOnOrg/mop-frontend/blob/master/src/components/signature-add-form.js#L453) is putting at the bottom:
`export const WrappedComponent = Banana` at the bottom (and then in the test do `import { WrappedComponent as Banana } from 'lib/banana'`) (to use the article's example).  If that works, then let's do that.  Otherwise, do what works, and it's totally good and proper to update a component to make it more testable.

Issue #479, opened 2018-01-26
Remove dependences that use React.PropTypes


jparkrr:
I was trying to see what I could do to fix the warning:
```
Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs
```

Spoke's code was all updated, so the issue was dependences that are behind:

- Material-ui (I've sent a PR to update spoke https://github.com/MoveOnOrg/Spoke/pull/480)
- [material-ui-color-picker](https://github.com/LoicMahieu/material-ui-color-picker)
  - hasn't been updated, but it would be easy enough to send a PR or fork
- React-router (I've sent a PR to update spoke https://github.com/MoveOnOrg/Spoke/pull/480)
- React-formal needs to be updated to at least to 25.4
  - This package is for defining a schema and validating forms
  - They don't have a good changelog, so could be tricky to verify https://github.com/jquense/react-formal/releases
- React-apollo needs to be updated to at least 1.1.1
  - Looks like there was some work in https://github.com/MoveOnOrg/Spoke/pull/418
  - But the work is also long to complete and tricky to verify

So I won't be able to fix the warning 😢. I just wanted to document what I found.

jparkrr:
Found a few more that might need to be updated:

- redux
- [react-async-script](https://github.com/dozoisch/react-async-script)

I ran these commands to find all (hopefully) React.PropTypes deps. You'll also see dependencies of dependencies, so you have to check what depends on it if it's not a direct dependency.
```
grep -Rl "PropTypes.*react" node_modules/
grep -Rl "React.PropTypes" node_modules/
```

Frydafly:
@schuyler1d checking in case this is relevant given a potential React upgrade coming soon?

schuyler1d:
i think this is mostly gone already, but will definitely be gone if/when we merge the React upgrade (AFAIK)

Arique1104:
Hello,

My name is Arique (they/them) and I am the Community Manager for Spoke. First of all, thank you so much for creating this issue and for everyone’s contributions to this ticket.

I am going through all of our issues and beginning to close all issues created in 2018 or older. Your issue qualifies.

I am giving everyone one week to answer the following question: 

- **Has this issue been resolved during the subsequent Spoke Updates?**

On November 16th I will be closing this issue if:
1. There is no response
2. You respond yes

[...]

PR #480, opened 2018-01-26
Update deps: material-ui and react-router

jparkrr:
See https://github.com/MoveOnOrg/Spoke/issues/479 for more details. But the short story is these are a few of the updates needed to be able to upgrade react in the future (they use `propTypes` from the main react package, which is deprecated).

Updating these two deps won't have any meaningful effect right away, so feel free to close this if you don't want to risk it. It's nice to get some work out of the way and stay semi-current though.

Overall I feel pretty confident that there won't be big issues, but not 100%.

# Material-ui 0.15.2 -> 0.18.7
- Material-ui keeps a good changelog of breaking changes, https://github.com/mui-org/material-ui/releases.
- I don't upgrade to the 1.x-beta branch because it might require more changes.
- v0.19 removes `react-tap-event-plugin` so we would need to replace all onTouchTaps with onClicks, so I don't upgrade to that. https://github.com/mui-org/material-ui/releases/tag/v0.19.0
- The only breaking change that seems applicable to Spoke is  in `CircularProgress`, which I've updated. See https://github.com/mui-org/material-ui/releases/tag/v0.16.0-rc1

 # React-router 2.5.2 -> 3.2.0
- react-router v3 seems like they only removed things that were already deprecated in v2, so we should be safe
- we use react-router-redux which is compatible with v3
- the latest version of react-router is v4, which is a complete rewrite and might need a lot of changes (v3 is still maintained though)


shakalee14:
Hey this is awesome. We def need to update! - after doing a quick round of QA, i see the following styling regressions (maybe from material ui update?):

Admin Campaign Menu, + button to create a new campaign overlays awkwardly:

screen shot 2018-01-26 at 2 28 01 pm


Texter Campaign Menu, # of todos for texts hides behind 'SEND FIRST TEXT' and 'SEND REPLIES' buttons: 

screen shot 2018-01-26 at 2 27 52 pm

shakalee14 (changes requested):
See comment :)

jparkrr:
@shakalee14 thanks for checking it! I fixed those issues

schuyler1d:

Frydafly:
hey @schuyler1d & @stefanhayden -- checking in whether this PR is still relevant after 2 years? or if it's at all related to current planned updates?

schuyler1d:
let's close this.  This was great work but I don't think it informs the current update roadmap.

PR #467, opened 2018-01-16
Upgrade Auth0 lock.js to fix login problems

hiemanshu:
Also update documentation to add the new required setting.

The endpoints, and other changes made by Auth0 were not backward compatible with
the version of lock.js we were running.

PR #482, opened 2018-01-29
User data editing merge auth0 login

schuyler1d:
This resolves conflicts of the two PRs for qa:
https://github.com/MoveOnOrg/Spoke/pull/472
https://github.com/MoveOnOrg/Spoke/pull/467
Rather than merging either of those, please merge this one (which will mark those as merged), but stop conflicts from occurring.


src/client/auth-service.js
```diff
@@ -15,7 +15,8 @@ export function login(nextUrl) {
       redirectUrl: window.AUTH0_LOGIN_CALLBACK,
       responseType: 'code',
       params: {
-        state: nextUrl
+        state: nextUrl || '/',
```
shakalee14, line 18: ok this is what fixed it - state being updated with the nextUrl was returning a weird letter string in some cases

shakalee14 (approved):

shakalee14:
We can make a separate issue for this. But we should add a 'home button/link' that the texter can click to get back to their dashboard. So after they update their name in acct, they can go to home without hitting the back button.

PR #486, opened 2018-01-31
Test component AssignmentSummary

jparkrr:
Here's a doozy of a test for AssignmentSummary, also related to #292 and requiring #478 

It's not perfect, but it's better than nothing.

Note the `/* @jest-environment jsdom */` at the top of the file in order to allow the component to be fully rendered and mounted in the test. There are some different ways of doing this, but I just wanted to get it working.

I had to eyeball stuff about `NOT_IN_USA`, etc. So please look at it and see if the descriptions and combinations can be made clearer somehow.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,5 @@
 {
-  "extends" : "airbnb",
+  "extends": "airbnb",
   "parser": "babel-eslint",
   "rules": {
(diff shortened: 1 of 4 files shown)


schuyler1d (approved):
Yei! more front-end testing

shakalee14:
Merged with #500

PR #489, opened 2018-02-01
fix people add button, which doesnt have bindings in enough places

schuyler1d:
On the main branch, the `[+]` button on the People admin page doesn't work -- we need some bindings on the methods or it doesn't retain the `this` variable -- a bit silly that they didn't fix this somehow in es6 `#js-gotchas`


shakalee14 (approved):

shakalee14:
Merged with #500

PR #495, opened 2018-02-06
Adds completed convos filter, new toolbar and ability to revisit mess…

### LINKED ISSUE #182: Ability for texters to revisit completed text conversations
Currently after hitting send, no way to go back when:
* the user 'opt-ed' out.  Use-case: 1. opt-out was a mistake, Use-case: want to refer back to it.
* If they hit 'skip' by mistake or changed their minds.
* If they want to 'complete' a reply with another message (in exceptional cases)?

shakalee14:
…aged contacts

Fixes #182 
TODO: 
- [x] fix tests to account for new button, filter


src/components/AssignmentSummary.jsx
```diff
@@ -130,6 +130,15 @@ class AssignmentSummary extends Component {
               contactsFilter: 'reply',
               hideIfZero: true
             })}
+            {(window.NOT_IN_USA && window.ALLOW_SEND_ALL) ? '' : this.renderBadgedButton({
```
schuyler1d: I don't think we need to filter for NOT_IN_USA -- even non-US folks might want to revisit :-)

schuyler1d (approved):
Nice (and fast) work!
This has been a longed-for feature for a long time!

shakalee14:
Adding more filtering to this feature

PR #509, opened 2018-02-14
Display full name of texters in assignment list

### LINKED ISSUE #454: Request: display last name and user ID info in Spoke texter screen
In a live campaign, Spoke currently only displays the first name of the texter to the admin. In cases where the an org has two or ore users with the same first name, should a user request more texts, this limited amount of information prevents the admin of knowing which user is which. See the two "Naomi" users in the screen shot as an example.

screen shot 2018-01-08 at 12 38 28 pm

By adding the last name and user ID to the texter screen, a campaign admin will be able to assign the correct user texts each time they are requested.

harpojaeger:
Fixes #454.

### DIFF
--- a/__test__/TexterStats.test.js
+++ b/__test__/TexterStats.test.js
@@ -59,7 +59,7 @@ describe('TexterStats (Non-dynamic campaign)', () => {
   it('contains the right text', () => {
     const stats = shallow(<TexterStats campaign={campaign} />)
     expect(stats.text()).toEqual(
-      'Test19%<LinearProgress />Someone99%<LinearProgress />'
+      'Test Tester19%<LinearProgress />Someone Else99%<LinearProgress />'
     )
   })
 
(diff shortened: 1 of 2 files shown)


schuyler1d:
This looks good @harpojaeger but can you update the test so it passes with your improved value.

```
TexterStats.test.js (24.569s)
  ● Console
    console.warn node_modules/react/lib/lowPriorityWarning.js:38
      Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs
  ● TexterStats (Non-dynamic campaign) › contains the right text
    expect(received).toEqual(expected)
    
    Expected value to equal:
      "Test19%<LinearProgress />Someone99%<LinearProgress />"
    Received:
      "Test Tester19%<LinearProgress />Someone Else99%<LinearProgress />"
```

harpojaeger:
Oops, my bad. I'll fix that today or tomorrow.

harpojaeger:
Woot woot, it works now.

shakalee14 (approved):

PR #513, opened 2018-02-16
21 copy campaigns

### LINKED ISSUE #21: Reusable Scripts
As an admin, I want to be able to reuse scripts from previous campaigns so that I can text a new group without re-creating the conversation flow each time. 

Copying a campaign involves creating a new campaign with title info (from campaign table), script data and canned response data (texters and campaign contacts should not carry over).

shakalee14:
Currently - copies a live campaign with the same title (with 'COPY'), description, canned responses and the first interaction step

Going to fix #21 

TODO: 
- [x] Make sure all interaction steps and q responses copy (update parent interaction id with interaction step id that has the same campaign id and a null parent interaction id - that's the 'root')

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -637,9 +637,10 @@ const rootMutations = {
       })
       const newCampaign = await campaignInstance.save()
       const newCampaignId = newCampaign.id
+      const oldCampaignId = campaign.id
 
       let interactions = await r.knex('interaction_step')
-        .where({campaign_id: id })
+        .where({campaign_id: oldCampaignId })
 
       const interactionsArr = []
       interactions.forEach((interaction, index) => {
@@ -674,17 +675,21 @@ const rootMutations = {
 
       await createSteps
 
-      let cannedResponses = await r.knex('canned_response')
-        .where({campaign_id: id })
+      let createCannedResponses = r.knex('canned_response')
+        .where({campaign_id: oldCampaignId })
+        .then(function(res){
+          res.forEach((response, index) => {
+            const copiedCannedResponse =
+              new CannedResponse({
+                campaign_id: newCampaignId,
+                title: response.title,
+                text: response.text
+              }).save()
+            }
+          )
+        })
 
-      cannedResponses.forEach((response, index) => {
-        const copiedCannedResponse =
-          new CannedResponse({
-            campaign_id: newCampaignId,
-            title: response.title,
-            text: response.text
-          }).save()
-      })
+      await createCannedResponses
 
       return newCampaign
 
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -637,8 +637,6 @@ const rootMutations = {
       })
       const newCampaign = await campaignInstance.save()
 
-      // TODO: get interaction steps associated with campaign id, and make new ones with other campaign id and other interaction ids
-
       let interactions = await r.knex('interaction_step')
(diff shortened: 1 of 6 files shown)


src/server/api/schema.js
```diff
@@ -622,6 +623,59 @@ const rootMutations = {
...
+      interactions.forEach((interaction, index) => {
+        const copiedInteraction =
+          new InteractionStep({
+            question: interaction.question,
+            script: interaction.script,
+            answer_option: interaction.answer_option,
+            answer_actions: interaction.answer_actions,
+            is_deleted: interaction.is_deleted,
+            campaign_id: newCampaign.id,
+            parent_interaction_id: interaction.parent_interaction_id
```
schuyler1d: This PR mostly looks good, except the parent_interaction_id needs to get redone (the old ones point to the old campaign).  One possibility is to re-use [updateInteractionSteps](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/schema.js#L361) though that seems to assume an hierarchical structure so morphing it into that form might be more annoying than just rolling your own.

The other way would be to basically re-implement some of its pieces -- basically making an `idMap` forEach interactionStep you save with the key the old id and the value being the new id, and then for all the parent_interaction_ids you'll need to update them based on the idMap.

src/server/api/schema.js
```diff
@@ -622,6 +623,77 @@ const rootMutations = {
...
+            answerActions: interaction.answer_actions,
+            isDeleted: interaction.is_deleted,
+            campaign_id: newCampaignId,
+            parentInteractionId: interaction.parent_interaction_id
+          }
+          interactionsArr.push(is)
+        }
+      })
+
+      let createSteps = updateInteractionSteps(newCampaignId, [makeTree(interactionsArr, id = null)], campaign, {})
```
schuyler1d, line 674: Can we write this `const createSteps = await updateInteractionSteps(....)` and remove the line below?

src/server/api/schema.js
```diff
@@ -622,6 +623,77 @@ const rootMutations = {
...
+        is_archived: false
+      })
+      const newCampaign = await campaignInstance.save()
+      const newCampaignId = newCampaign.id
+      const oldCampaignId = campaign.id
+
+      let interactions = await r.knex('interaction_step')
+        .where({campaign_id: oldCampaignId })
+
+      const interactionsArr = []
```
schuyler1d, line 645: This looks correct.  It might be more concise in the form:
```
const interactionsArr = interactions.map( (interaction, index) => ({
  id: 'new'+interaction.id,
  questionText: interaction.question,
  ...
  parentInteractionId: ((interaction.parent_interaction_id) 
                                   ? 'new'+interaction.parent_interaction_id)
                                   : undefined)
}))
```

src/server/api/schema.js
```diff
@@ -622,6 +623,77 @@ const rootMutations = {
...
+          }
+          interactionsArr.push(is)
+        }
+      })
+
+      let createSteps = updateInteractionSteps(newCampaignId, [makeTree(interactionsArr, id = null)], campaign, {})
+
+      await createSteps
+
+      let createCannedResponses = r.knex('canned_response')
```
schuyler1d, line 678: maybe move the `await` up to this line too

schuyler1d (approved):
:+1: 
some (optional) style suggestions, but this was some great work!

PR #515, opened 2018-02-19
Reduce queries for needsMessage and other texter moments

schuyler1d:
This reduces some queries which are unnecessary in certain states.

1. Whenever we have a messageStatus=='needsMessage' we shortcut responses for messages and response values since there won't be any.
2. We don't need to lookup zip information if we already have it in the contact record and even if we don't we should try getting it programmatically first.

This PR is against the decoupling-prep branch, since it builds on that work.  After that branch is merged, we should either merge this one next, or edit the PR to go against `main`

### DIFF
--- a/src/containers/TexterTodo.jsx
+++ b/src/containers/TexterTodo.jsx
@@ -38,7 +38,6 @@ class TexterTodo extends React.Component {
 
   render() {
     const { assignment } = this.props.data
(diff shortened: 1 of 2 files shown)


shakalee14:
this in the browser
```

shakalee14 (approved):

shakalee14:
Merged with #520

Issue #487, opened 2018-02-01
Scaling meta-issue


schuyler1d:
Scaling goals:
* 2000+ simultaneous texters
* 1million+ messages/day
* Bigger lists loaded in contacts - without getting stuck/slow

This issue is meant to track/discuss ideas around scaling. Please refer to it in relevant pull-requests and issues.

Some categories to consider:
* Redis/Caching intermediate layer for texter-bottlenecked apis (#254)
* Finding performance bottlenecks, in general
* Client-side performance wins
* Performance testing

shakalee14:
We think this can happen in 9 steps --> "smaller" PRs vs 1 big PR
- [ ] Add/install `connect-redis` and `fakeredis` for production and dev  
- [ ]  Implementing Campaign ORM structure. Implement it in easy spots - so for example, accessRequired / access control. This way we can prove that this concept works and is self contained.
- [ ]  Then stub out ORM 
- [ ] Move database calls into the ORM
- [ ] Implementing what happens in the start campaign (write)--> when you create a campaign it'll make a bunch of key value pairs to represent that campaign
- [ ] Mutation writes - instead of db, write to redis
- [ ] Create background process that syncs redis writes to db
- [ ] Fill out reads
- [ ] Monitor sql when debugger is on - using some knex debugging features to make sure sql calls happen - this way we make sure we aren't actually touching the DB

Issue #254, opened 2017-09-21
Keep texter todo list updated without significant load on database at high texter volume


sreynen:
We were previously using Apollo polling on the texter todo list data query to keep it refreshed, but this caused too much database load with a high number of texters all repeatedly running the queries, so https://github.com/MoveOnOrg/Spoke/pull/253 removed that. So currently texters need to manually refresh their browser to get the latest data. Ideally we can find a better way to refresh the data without too much database load.

sandramchung:
More context in https://github.com/MoveOnOrg/Spoke/issues/255. Some ideas: 

- implement auto-refresh for the ToDo list that doesn't hit the db every time
- restore polling, but boot texters out of the ToDo list if they leave it open for too long
- make manual refresh the obvious only choice with e.g. a manual refresh button on the ToDo list screen

schuyler1d:
Some thoughts:
* If we have a redis/memcache layer, the generic version might be to insert between knex and the db, the same way django-cachalot works.  If knex doesn't allow that, then maybe we can "conveniently" use the rethink-knex-adapter.  My concern here is that it might be 'too' global for a bunch of queries that really need to be super-up-to-date, while focusing on the sms message queue really might be the only place in our application that exerts this much db pressure and need of freshness.
* So then maybe use redis, e.g. as an (optional) queue for 'updates'.  e.g. for each (?logged in) user, we send messages related to their assignments into the redis queue, and then the query that is polled only looks at the redis queue for updates/new messages.

ben-pr-p:
We can do, via Redis pubsub, a new message on a new reply -> push a message to the client over websockets telling them to check for new updates?

This avoids
a) Ever checking for an update when there won't be one (polling the DB)
b) Having to store reply state in Redis / anywhere outside of the DB

Downside is maintaining a bunch of open socket connections, but this shouldn't each up too much CPU resources if these connections are mostly idle, and should still solve the problem if the bottleneck is the DB.

If this approach makes sense to people, I'll probably tackle this Monday!

schuyler1d:
Hi @ben-pr-p I've been thinking through the same problem for a bit.  I'd like to coordinate on planning and come to consensus on a full plan that will work over-all before diving into an implementation

For that purpose, I wrote up a planning document as a PR here: https://github.com/MoveOnOrg/Spoke/pull/524

[...]

PR #524, opened 2018-02-24
draft plan for scaling message-response cycle with redis

schuyler1d:
This is a draft plan and request-for-comment for scaling addressing issue #254 and meta-issue
#487

cc: @shakalee14 @lperson @ben-pr-p 

[rendered proposal](https://github.com/MoveOnOrg/Spoke/blob/redis-plan/docs/HOWTO-scale-spoke-plan.md)


ben-pr-p:
Is the idea that each of the queues `newassignments-<texter_id>` and `conversation-<contact_cell>` would be independently updated alongisde DB updates, or would be cached DB queries?

schuyler1d:
@ben-pr-p I should probably be more explicit about this in the doc, but I'm imagining a process where through a combination of 'preloading' and caching results when the DB needed to be hit, the Redis cache is updated.  The DB would be updated from a separate Redis queue, `message-write-queue`, so original writes are just to redis, and then the db catches up with syncs, rather than being the system of record.

docs/HOWTO-scale-spoke-plan.md
```diff
@@ -0,0 +1,172 @@
...
+#### Workflows using Data-structures
+
+* Access Control
+
+  1. Load from `texterinfo-<texter-id>`
+
+* getNew (needsMessage)
+
+  1. Load (and LPOP) from `newassignments-<texter_id>-<campaign_id>`
+  2. If newassignments is empty, we check `unsentassigned-<texter_id>-<campaign_id>` for content.  If it does, we ?resend that data with an additional setting in the API getNew so the client knows that the 'original queue' is empty.
```
shakalee14, line 134: I dont understand the concept of unsentassigned-texter_id-campaign_id. When is this created? When we initially create the campaign (like in updateAssignments?).  Is the idea that newassignments is first created, and once the person starts texting, we create a new thing called unsent.. which keeps track of message status? 

Does newassignments get transformed to unsentassigned?

schuyler1d: So the use-case that `unsentassigned-` helps us with is imagine the following situation:
1. Some things are in newassignments
2. A texter's browser sends the query to get some new assignments.  So the system sends them some items from newassignments and removes them from newassignments (so that future queries won't re-send the same ones)
3. The texter leaves, a network failure occurs, or the texter's browser crashes.
4.  Finally later the texter logs in again to resume.  Their browser does more queries to newassignments, and they finish their task

The problem here is that those middle contacts got dropped.  By storing unsentassigned, we track which ones are 'in the queue' -- i.e. sent to the texter, but the texter has not yet actually sent them a message/response.  Thus we only want to fully remove them after the texter *sends* them a response.  Until then, they're in 'purgatory'--i.e. we'll skip them until the texter thinks they're finished.

package.json
```diff
@@ -84,6 +84,7 @@
     "body-parser": "^1.15.2",
     "camelcase-keys": "^4.1.0",
     "chart.js": "^1.1.1",
+    "connect-redis": "^3.3.0",
```
shakalee14: what do we need connect-redis for? is this for your login work? @schuyler1d

shakalee14 (approved):

PR #525, opened 2018-02-25
Database test stability

### LINKED ISSUE #518: Tests sometimes fail due to race-condition
Occasionally, the db getting created and seeded doesn't happen before the tests start to run.
e.g.
```
 FAIL  __test__/workers/jobs.test.js (10.253s)
  ● Console

    console.log node_modules/minilog/lib/node/console.js:8
      backend warn NO TWILIO CONNECTION
    console.log node_modules/minilog/lib/node/console.js:8
      backend warn Twilio will not be able to send without TWILIO_MESSAGE_SERVICE_SID set

  ● test getTimezoneByZip › does not find a zipcode in the database!

    error: select * from "zip_code" where "zip_code"."zip" = $1 - relation "zip_code" does not exist
```
This is because the zip_code table wasn't created yet.  Somehow we should make sure the db is created (and maybe re-created each test? or each file?) -- see __test__/test_helpers.js -- the setupTest() function is empty -- maybe this is where this should go?

How should it work?

lperson:
This addresses and should resolve #518.

Note that the solution is has two parts.  The code changes here, and a requirement to add a function call to beforeAll() and afterAll() in every test that uses the database.  The second part is documented in docs/EXPLANATION-development-guidelines.md.

@schuyler1d and @shakalee14 please review.  Thanks.

### DIFF
--- a/__test__/backend.test.js
+++ b/__test__/backend.test.js
@@ -1,7 +1,6 @@
 import { schema, resolvers } from '../src/server/api/schema'
 import { graphql } from 'graphql'
 import { User, Organization, Campaign, CampaignContact, Assignment, r } from '../src/server/models/'
-import { sleep } from '../src/workers/lib'
 import { resolvers as campaignResolvers } from '../src/server/api/campaign'
 import { getContext,
   setupTest,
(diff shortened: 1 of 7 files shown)


lperson:
WIP

lperson:
Still a work in progress :-(

PR #527, opened 2018-02-26
fix #395 - upload contacts does not work in firefox

### LINKED ISSUE #395: "Upload contacts" button doesn't work in Firefox
Steps to reproduce:

1) In Firefox, log in as admin and create a new campaign
2) In "Contacts" section, click "Upload Contacts"

In other browsers, this opens a file choose dialog. In Firefox, nothing happens. Specifically, I tested this with Firefox 57.0 on MacOS.

ben-pr-p:
Description of the reason for the problem can be found here: https://github.com/mui-org/material-ui/issues/647

Inputs inside buttons is apparently not part of the W3C spec, but Chrome does it anyways.

This fixes it and still users the same button component, and works in all other browsers too

### DIFF
--- a/src/components/CampaignContactsForm.jsx
+++ b/src/components/CampaignContactsForm.jsx
@@ -176,18 +176,21 @@ export default class CampaignContactsForm extends React.Component {
   renderUploadButton() {
     const { uploading } = this.state
     return (
-      <RaisedButton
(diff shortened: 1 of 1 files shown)


shakalee14:
Fixes #395

shakalee14 (approved):

PR #528, opened 2018-02-26
Add/mailgun

ben-pr-p:
Hey! I saw you all removed mailgun a while back (https://github.com/MoveOnOrg/Spoke/pull/143). 

We want to use it, and it works nicely with the heroku button set up, and so this adds it back optionally and wraps it in the same api as `nodemailer`. If process.env.MAILGUN_API_KEY and process.env.MAILGUN_DOMAIN exist, it will use the mailgun adapter instead.

I also added free tier mailgun to the heroku deploy button, and instructions for how to set up a domain with mailgun in the heroku deploy instructions

### DIFF
--- a/app.json
+++ b/app.json
@@ -195,6 +195,7 @@
   },
 
   "addons": [
-    "heroku-postgresql:hobby-dev"
+    "heroku-postgresql:hobby-dev",
+    "mailgun:starter"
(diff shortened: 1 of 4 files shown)


shakalee14 (approved):
LGTM

PR #536, opened 2018-03-11
linting main

schuyler1d:
ran `npm run lint` on main.


shakalee14 (approved):
LGTM - tested a bit locally but let's look at this on stage

PR #540, opened 2018-03-14
Invite page is broken

gracecarey:
After creating an account, I was sent to the homepage with a message that I do not have a company, when I should have been directed to a view inviting me to create a company.


shakalee14:
I think this has been fixed - thanks for the note.

PR #543, opened 2018-03-14
rollback jest version so tests run, tweak test script so it runs

lperson:


lperson:
#543 tracks a more permanent change to ensure that tests run with the last dependencies.

PR #547, opened 2018-03-19
Send arrow fix

shakalee14:
fix more mobile device bug on texting screen


shakalee14:
this was qa'd on stage by @ErikKirkstein and I before merging

Issue #539, opened 2018-03-13
Remove env setting AUTH0_LOGIN_CALLBACK (and logout)


schuyler1d:
env settings AUTH0_LOGIN_CALLBACK and AUTH0_LOGOUT_CALLBACK are extraneous and can use BASE_URL.

 - [x] Make them unnecessary by changing the context they are used in.
 - [x] Remove in .env.example
 - [x] Remove from Heroku
 - [x] update the documentation

schuyler1d:
fixed with 20ead811087d999b7b40c2025163b0ca38d78dc7

PR #564, opened 2018-03-31
removed AUTH0_LOGIN_CALLBACK and AUTH0_LOGOUT_CALLBACK env vars

dtoakley:
Related issue: #539


.gitignore
```diff
@@ -13,3 +13,4 @@ test.sqlite
 *~
 *#
 coverage/
+.idea/
```
shakalee14, line 16: hey - what's this `.idea/` directory?

shakalee14 (approved):
LGTM - will test this on stage this week

.gitignore
dtoakley: it's a local dev directory automagically created for projects when you're using IntelliJ IDEA. I might be the only person rolling this IDE but we don't want the dir committed.

PR #572, opened 2018-04-17
Adds cacheable query for logging in

shakalee14:


schuyler1d (approved):
The code looks good.  It would be nice to get us back to tests running/passing before going too far down this branch.

PR #575, opened 2018-04-26
heroku knex ssl config / exports error handling

codygordon:
1. We found that using 'standard'+ tier Heroku PG plans require the essentially undocumented `ssl` knex config option. See: http://cek.io/blog/2016/10/24/knex-configuration-heroku/

>Note: obviously you could just go with the env variable option for this, so it's not super critical.

2. I noticed that if an export failed it would simply result in an unhandled promise rejection, and the user wouldn't be notified or even really be able to tell what happened via the logs. I wrapped the export block in a try/catch and send an email + log the error message upon failure. 

3. If the job deletion fails for some reason you have to manually delete the job from the db to update the UI (the export button will just remain disabled if the job still exists), so I added retries to the job deletion attempt.

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -540,6 +540,10 @@ export async function exportCampaign(job) {
         text: `Your Spoke exports are ready! These URLs will be valid for 24 hours.
         Campaign export: ${campaignExportUrl}
         Message export: ${campaignMessagesExportUrl}`
+      }).catch((err) => {
+        log.error(err)
+        log.info(`Campaign Export URL - ${campaignExportUrl}`)
+        log.info(`Campaign Messages Export URL - ${campaignMessagesExportUrl}`)
       })
       log.info(`Successfully exported ${id}`)
     } catch (err) {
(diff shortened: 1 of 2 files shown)


src/workers/jobs.js
```diff
@@ -521,37 +521,57 @@ export async function exportCampaign(job) {
...
+      const key = `${campaignTitle}-${moment().format('YYYY-MM-DD-HH-mm-ss')}.csv`
+      const messageKey = `${key}-messages.csv`
+      let params = { Key: key, Body: campaignCsv }
+      await s3bucket.putObject(params).promise()
+      params = { Key: key, Expires: 86400 }
+      const campaignExportUrl = await s3bucket.getSignedUrl('getObject', params)
+      params = { Key: messageKey, Body: messageCsv }
+      await s3bucket.putObject(params).promise()
+      params = { Key: messageKey, Expires: 86400 }
+      const campaignMessagesExportUrl = await s3bucket.getSignedUrl('getObject', params)
```
schuyler1d, line 536: in case email gets borked, and since you're changing this code, can you drop a `log.info` log here with the `campaignMessagesExportUrl` -- that way, if the email fails, at least the heroku admin can see the url from the logs.

schuyler1d (approved):
:+1:  Thanks for submitting!  I have one suggestion, but we can also merge as-is.

src/workers/jobs.js
codygordon: No problem, I think directly catching errors on the email and logging there should do the trick since the try/catch would catch an error at any step and the URLs may not be created (or in scope unless defined before the block). The alternative would be using `.catch` on every `await` function, I guess.

See: https://github.com/MoveOnOrg/Spoke/pull/575/files#diff-ee93a980f871b21055cf3e181d6583beR543

shakalee14:
merged with #583

PR #581, opened 2018-05-07
improve db race condition scenarios in tests

schuyler1d:
Hopefully this paves a way to removing the race conditions on startup -- starting at least with tests.

This also fixes some warnings that show up in tests

### DIFF
--- a/src/components/SendButton.jsx
+++ b/src/components/SendButton.jsx
@@ -47,9 +47,9 @@ class SendButton extends Component {
 }
 
 SendButton.propTypes = {
-  threeClickEnabled: PropTypes.boolean,
-  onFinalTouchTap: PropTypes.function,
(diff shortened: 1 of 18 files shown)


src/components/SendButton.jsx
```diff
@@ -47,9 +47,9 @@ class SendButton extends Component {
 }
 
 SendButton.propTypes = {
-  threeClickEnabled: PropTypes.boolean,
-  onFinalTouchTap: PropTypes.function,
-  disabled: PropTypes.boolean
+  threeClickEnabled: PropTypes.bool,
```
shakalee14, line 50: are some of these diffs because of a linting pass?

shakalee14 (approved):
Great! Ran it locally - just with caveat that some of the changes here I believe are from the linting pr

PR #589, opened 2018-05-11
Adding two new 'league table' sample queries

joemcl:
Added 'count of all texts sent by all users in all campaigns' and  'count of all texts sent by all users in one campaign' for use in a 'leader board'/'league table' Dashboard in redash.


shakalee14 (approved):

joemcl:
Thanks @shakalee14 !

PR #600, opened 2018-05-23
added null default value to parent_interaction_id

mkc1:
this fixes an error when running with sqlite when creating interaction steps on a campaign.


shakalee14:
Merged with #621

PR #624, opened 2018-06-05
Allows user to delete message and press send which doesnt send a mess…

### LINKED ISSUE #577: Want option of opting out without a message
Occasionally opting out without a message is called for. 

It would be good to allow opt-out without having to send a message.
This issue is probably entirely client-side.
In src/containers/AssignmentTexterContact:handleOptOut you'll notice that it sends a message and then createOptOut, etc.

We should have an additional button "Opt out without sending message" or something similar in the interface which createOptOut but without sending a message. Alternatively we could document that if no message content is included, then we don't send a message, but still opt them out.

shakalee14:
…age and opts user out

Fixes #577

### DIFF
--- a/src/containers/AssignmentTexterContact.jsx
+++ b/src/containers/AssignmentTexterContact.jsx
@@ -211,7 +211,7 @@ export class AssignmentTexterContact extends React.Component {
       snackbarError,
       snackbarActionTitle,
       snackbarOnTouchTap,
-      optOutMessageText: "I'm opting you out of texts immediately. Have a great day." || "",
+      optOutMessageText: "I'm opting you out of texts immediately. Have a great day.",
       responsePopoverOpen: false,
       messageText: this.getStartingMessageText(),
       optOutDialogOpen: false,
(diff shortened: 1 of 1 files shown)


src/containers/AssignmentTexterContact.jsx
```diff
@@ -211,7 +211,7 @@ export class AssignmentTexterContact extends React.Component {
       snackbarError,
       snackbarActionTitle,
       snackbarOnTouchTap,
-      optOutMessageText: "I'm opting you out of texts immediately. Have a great day.",
+      optOutMessageText: "I'm opting you out of texts immediately. Have a great day." || "",
```
schuyler1d: What does this change do?  the hard-coded string will never evaluate to false....

src/containers/AssignmentTexterContact.jsx
```diff
@@ -440,18 +440,29 @@ export class AssignmentTexterContact extends React.Component {
...
-    try {
-      await this.props.mutations.sendMessage(message, contact.id)
-      const optOut = {
-        cell: contact.cell,
-        assignmentId: assignment.id
-      }
 
+    if(optOutMessageText.length > 0){
+      try {
+        await this.props.mutations.sendMessage(message, contact.id)
```
schuyler1d: I think this would be more readable if we just remove the top condition and then wrap the `> 0` condition around the sendMessage.

schuyler1d (approved):
suggested tweaks, but I think this would work as-is.

shakalee14:
apologies @schuyler1d  - i forgot to tag it as `wip`

I need to do a few more things, like, I want to change the button message from `send` to `opt out without text` and clean up the code

PR #651, opened 2018-06-15
Expose skipped messages

bchrobot:
Expose skipped messages via a new assignment summary button. Addresses #560 

screen shot 2018-06-15 at 12 58 59 pm


shakalee14:
Hey @bchrobot - this functionally looks great.

I checked out how this would look on mobile (found this on android and iphone models), and I believe there is a little bit more work to do. As you can see, the opt out and reopen button are covering some of the input box.

Also, since we will have 4 buttons, it would probably be helpful to update this in the texter FAQ (this of course could be another issue). For example:
- Send First Texts means...
- Past Messages means...
- Skipped Messages means... (which would include that you need to `reopen` it in order for it to appear in send replies)



screen shot 2018-06-18 at 10 13 44 am

shakalee14:
hey @bchrobot - i can work on updating the FAQ! thanks for your work on this

shakalee14:
I'm not getting consistent behavior on this feature - will try to write notes on why today after finishing stage testing.

shakalee14:
Ok @bchrobot - sorry for the delay.  

In order to produce what I'm seeing. Create a campaign, assign yourself to those texts, and then send texts, receive replies (or use send-replies endpoint in admin). Skip replies, which should mark them as 'closed' in the db. I then see the new button for 'skipped messages'. I reopen those messages, and navigate back to home and then see the button disappear (because they've filtered back into send replies). If I hit skip reply again, the message is 'lost' (and I confirmed that I had messages marked as 'closed' that were'nt exposed in the UI). If this is intended behavior (which gives the texter 1 more chance to review skipped replies), then we can document that.

shakalee14:
@bchrobot - i did refresh to no avail but i can keep investigating - i know this feature is already being used in production on some instances - has anyone reported this behavior?

shakalee14:
Great - this seems to be fixed now - tested locally - I'm able to see it go through the expected cycle

shakalee14:
Merged with #682

Issue #626, opened 2018-06-06
Make Auth0 optional -- regular user/pass login


schuyler1d:
Auth0 does some decent user/pw management, but the API keeps changing and it's a large burden and stumbling block for new developers getting the project up and running.

Instead there should be some option that sets up a login for the super-user at the beginning on start, and then allows users to set a user/pass just like updating their other account vars.

However, we should also maintain backwards compatibility and support Auth0 by default.  There would be an env variable that changes auth to use the new setup.

 - [x] Setup env variable AUTH_BACKEND='local' (if not set or if it's 'auth0' it falls back to Auth0)
 - [x] Modify [src/server/auth-passport.js](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/auth-passport.js) to support both and do the right password testing.
 - [x] Load/config the above appropriately in src/server/index.js
 - [ ] Setup/switch the registration screen -- probably replacing the auth0 'booting' here (https://github.com/MoveOnOrg/Spoke/blob/main/src/server/middleware/render-index.js#L27) conditional on AUTH_BACKEND value
 - [ ] Probably adapt the [UserEdit.jsx container](https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/UserEdit.jsx) for a registration view when appropriate.
 - [ ] Do something on first run, when there's no accounts -- sort of a default registration screen
 - [ ] Document the other backend and setup, and possibly make it the default suggested setup for development environments.
 - [ ] Move Auth0 docs to someplace more appropriate than polluting the README file.

schuyler1d:
see PR for progress.

ibrand:
I believe this is fully implemented since I've used local login before! If I'm wrong and this should be re-opened go ahead

PR #658, opened 2018-06-19
wip: Local login (w/o Auth0)

schuyler1d:
tackling https://github.com/MoveOnOrg/Spoke/issues/626


src/server/auth-passport.js
```diff
@@ -19,90 +22,113 @@ export function setupAuth0Passport() {
...
-          done(null, user)
-        }
-      })
-    })
-  }
-  )
+      AuthHasher.verify(
+        password, hashed,
+        (err, verified) => (verified
+                            ? done('Validation failed', null)
```
jparkrr: I was trying to add onto your work on this PR, but didn't get too far. I'll try to get back to it soon, but the main thing I noticed is that this conditional needs to be flipped around I think.

jparkrr:
Also do you think it makes sense to implement non-auth0 signup as you have it in the verifier, so every failed login is a signup? Maybe that was just for testing? Should it be maybe a new graphql modifier `createUser()`, hooked up to a form?

src/server/auth-passport.js
schuyler1d: dur, thanks.

I'll need to verify that what I thought was working actually is if this was really flipped then maybe I wasn't even as far as I thought....

PR #661, opened 2018-06-20
Sort question responses

bchrobot:
Texters develop muscle memory for the script which is disrupted when the response order changes. Responses to questions should have a consistent sort order.

### DIFF
--- a/src/server/api/question.js
+++ b/src/server/api/question.js
@@ -25,6 +25,7 @@ export const resolvers = {
       r.table('interaction_step')
         .filter({ parent_interaction_id: interactionStep.id })
         .filter({ is_deleted: false })
+        .orderBy('answer_option')
         .map({
           value: r.row('answer_option'),
           action: r.row('answer_actions'),


src/server/api/question.js
```diff
@@ -25,6 +25,7 @@ export const resolvers = {
       r.table('interaction_step')
         .filter({ parent_interaction_id: interactionStep.id })
         .filter({ is_deleted: false })
+        .orderBy('answer_option')
```
shakalee14, line 28: @bchrobot what if this was `.orderBy('answer_option', 'desc')` in order to make sure 'yes' or yes answers are first.

bchrobot: The way that the campaign I'm working with uses this is that they have numerically prefixed answers (ex. 1. Lean D Unspecified, 2. Strong D. Unspec, ... 6. Refused/Angry). There may be multiple answer options for the same number (this is being used somehow by their data team but I'm not sure exactly how). So for that ascending sort is definitely better. And if you want Yes to be first you could use '1. Yes'.

The more important thing is that the ordering is consistent.

Eventually there could be a `position: Int` column that stores the ordering for each group of sibling answer options. This would probably want draggable reordering (and possibly indenting) as well as an expand/collapse option for the responses (similar to WordPress's menu editor) to make it easier to drag smaller things around.

shakalee14 (approved):

PR #662, opened 2018-06-20
[react-v16] Part I: Basics

bchrobot:
This has just the bare bones, no complications dependency updates. This breaks up PR #627 into 3 sub-PRs. Because of various breaking changes, I believe Spoke will not run until they are all updated in lockstep, but at least the functionality changes are broken up into more easily reviewed segments:

* `material-ui` - see bchrobot/Spoke#2
* `react-router` - see bchrobot/Spoke#4
* `react-apollo` - see bchrobot/Spoke#5
* `webpack` + `react-hot-loader` - see bchrobot/Spoke#6

The above PRs are against the branch for this PR.


schuyler1d (approved):
this is fine to go into react-v16 branch.

Have you tested that the onClick works on mobile devices?  I believe we moved away from click specifically because of an issue with some mobile devices.

PR #663, opened 2018-06-21
[WIP] Handle MMS responses

bchrobot:
Texters had been getting blank messages and were confused by what that meant. This PR checks to see if the blank message contains multimedia attachments and renders links to those preceded by a warning (in case of lewd responses to campaign messaging).

Screen shot of from live Twilio messaging:
screen shot 2018-06-21 at 2 25 23 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,6 +126,7 @@
     "react-chartjs": "^0.7.3",
     "react-dom": "^15.6.1",
     "react-formal": "^0.18.5",
+    "react-linkify": "^0.2.2",
     "react-router": "^2.5.2",
     "react-router-redux": "^4.0.5",
     "react-tap-event-plugin": "^2.0.0",
--- a/src/components/MessageList.jsx
+++ b/src/components/MessageList.jsx
@@ -39,7 +39,6 @@ class MessageList extends React.Component {
 
     this.handleOpenLink = this.handleOpenLink.bind(this)
     this.handleCloseLink = this.handleCloseLink.bind(this)
-    this.textDecorator = this.textDecorator.bind(this)
     this.componentDecorator = this.componentDecorator.bind(this)
   }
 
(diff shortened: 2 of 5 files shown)


schuyler1d:
Hi, so @bchrobot -- I like the feature.

I agree that we should restructure this a bit for security and extensibility.
1. I agree we should move this into twilio.js since the MediaUrl... keys are specific to twilio
2. I think as it is Linkify() is insecure -- I don't want to make links to any url.  I suggest that we maybe do the same in reverse that we do for sending images here: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/lib/twilio.js#L97-L110  i.e. for MediaUrls, we append to the text [http.....].
3. However, the above caveat means that we should UNmatch any texts that have that exact structure `[<url>]` and for anything that matches that with incoming messages, we maybe insert an 'X' right after the `[` or something.  (this is a bit hacky, but happy to go with different tweaks)
4. Finally, instead of Linkify() on the frontend and the warning being added to the message itself, let's linkify only things that match `[<url>]` in the message and then the client will render the message just below or above the message text.
5. It might also be worth when we are 'linkify'ing it we shorten the link text -- maybe just to the domain and a ....
6. This should all be wrapped by a config variable.  It might also be cool to have it also configurable BY-campaign.  Our experience is that campaigns that are like "Hey, we won!!!" get really awesome responses that we want to share with our texters.  OTOH, we have other campaigns that are reaching out to ..... possibly angry individuals and it would be good to keep this off for those.

Finally -- thanks for this and your other contributions!  Your work is really exciting!

shakalee14:
@bchrobot - i like the idea of making a component

joemcl:
Hi @bchrobot. This is for handling incoming images only? @schuyler1d you wrote above 'we maybe do the same in reverse that we do for sending images here: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/lib/twilio.js#L97-L110 ' - so we *can* send images with Spoke?

Frydafly:
@bchrobot thanks for your work! not exactly sure how MMS responses are handled now -- is the feature still relevant? @schuyler1d

schuyler1d:
This was (re)implemented by Jeff Mann in 9.x period.
https://github.com/MoveOnOrg/Spoke/pull/1866

Issue #554, opened 2018-03-20
Implement "start campaign"


shakalee14:
When a campaign starts, we want to push all the required assignment/contact info into the redis info 

Implement an `updateAssignments` method in src/server/models/cacheable-queries.js
as described here:
https://github.com/MoveOnOrg/Spoke/blob/redis-plan/docs/HOWTO-scale-spoke-plan.md#workflows-using-data-structures
(both for dynamic and regular campaigns)

The tricky part here is that we are trying to upload the full contact datastructure that redis will need for serving getContact -- which is essentially #553 -- there is a little bit of a chicken-egg aspect here -- it helps to have data to try to load, but we need to know what form the data should be in before we write the loader.  Doing this task first, we should be ok with imperfect/incomplete contact info and plan to iterate on this step after we go between this and #553.

## Some notes
0. The contact info that is saved to redis will need to include all the information pulled for and used by [AssignmentTexterContact::getContact](https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/AssignmentTexterContact.jsx#L879-L922)
1. Unlike many of the redis/queries methods, this one will be synchronous to database changes.  So unlike the auth stuff, where we pull in the db calls to cacheable queries and don't do them otherwise, this one will 'just' write to redis -- probably by loading the data from the database
2. We should be careful to accomodate changes in assignments -- deletion might be a little tricky and might require an additional redis hash data structure mapping all 'current' assignments for a particular campaign.  (which will allow us to iterate over it and delete any queues that have been, e.g. re-assigned/changed)
3. Note that Larry's WIP PR will allow re-assignment after the campaign has started, too -- when that is merged, we'll need to call this method as well -- or write a similar method.

shakalee14:
Hey @schuyler1d - is it true that there's more work to do on this branch https://github.com/MoveOnOrg/Spoke/tree/decoupling-contactinfo before workingon this?

schuyler1d:
this was completed.

PR #673, opened 2018-06-28
Adds updateAssignments method to cache queries and starts lpush assig…

shakalee14:
…nments

**WIP**
For #554 

TODOS / AND BIG QUESTIONS
- [x] Decide on structure to push assignment data into redis -- this should make it easy to pull(Pop) data out of redis in order to satisfy getContact method (so there's not extra processing when it comes out)
- [x] When should we expire data and should we support worker retries
- [ ] FUTURE todo: Figure out how to pull data out with contact Filters --> currently getContacts gets an assignment with a knex query and then adds where statements (via helper functions) to return the correct group of contacts for a user. If contacts are now retrieved in an array, going to have to figure out a consistent way to filter the array by established filters

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -74,6 +74,7 @@ const uuidv4 = require('uuid').v4
 import GraphQLDate from 'graphql-date'
 import GraphQLJSON from 'graphql-type-json'
 import { GraphQLError } from 'graphql/error'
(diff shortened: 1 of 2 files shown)


shakalee14:
WIP / closing until we figure out more about `getContacts` - we'll probably need it to be structured differently.

PR #678, opened 2018-07-02
Incorporating feedback from https://github.com/MoveOnOrg/Spoke/pull/579

lperson:


schuyler1d (approved):
looks good.  Thanks for following up. :+1:

PR #680, opened 2018-07-02
Fix zip column bug

bchrobot:
Like `external_id`, `zip` is an `optionalString` in `src/server/models/campaign-contact.js`. The way `optionalString` is defined requires an empty string rather than `null`.

This addresses #648

### DIFF
--- a/src/containers/AdminCampaignEdit.jsx
+++ b/src/containers/AdminCampaignEdit.jsx
@@ -187,7 +187,7 @@ class AdminCampaignEdit extends React.Component {
             cell: contact.cell,
             firstName: contact.firstName,
             lastName: contact.lastName,
-            zip: contact.zip,
+            zip: contact.zip || '',
             external_id: contact.external_id || ''
           }
           Object.keys(contact).forEach((key) => {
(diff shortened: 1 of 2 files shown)


shakalee14 (approved):

joemcl:
Hey @codygordon this is the issue that you fixed on our repo yesterday right?

codygordon:
Yeah, but see `optionalString` should configure your column to have a default of an empty string, so this is only a problem for databases that were created before the zip field was changed from `requiredString` to `optionalString`: https://github.com/MoveOnOrg/Spoke/blob/1a89127d1a707b21ec01290bac77f9837fc45663/src/server/models/custom-types.js#L19

My solution was just to do a one-off `alter table campaign_contact alter column zip set default '';` to fix the problem.

codygordon:
@bchrobot check your table schema, I'd be willing to bet it doesn't have the empty string default parameter. The issue here was that the model was changed from `requiredString` to `optionalString` in a recent release, thus requiring a manual alteration of an existing database's table to match the new model. Alternatively, your fix in the code works as well.

Edit: Well, I'm wrong about the change from `requiredString` to `optionalString`, but our database didn't have the default set for the zip column for some reason, and adding that fixed the issue. Not sure what in the last release borked that, but it is what it is. Your fix here is obviously good.

Issue #628, opened 2018-06-08
"Send First Texts" button remains green after all texts sent when using "Dynamically assign contacts" tool


ErikKirkstein:
Currently, when a Spoke campaign is using the "Dynamically assign contacts" feature, the "Send first texts" button in a texter's To-do menu remains green, implying a texter still has texts they can send even if all have been sent within the campaign.
![screen shot 2018-06-08 at 10 10 37 am](https://user-images.githubusercontent.com/32142717/41162552-3cf5a6b4-6b04-11e8-9cc1-a029d67dc88d.png)

To limit confusion, it would be helpful if this button is greyed out after all initial texts have been sent

dmitry-vereykin:
Investigating the issue.

shakalee14:
Handled with #686

PR #686, opened 2018-07-06
#628: "Send First Texts" button disables properly

dmitry-vereykin:
The "Send first messages" button will be disabled if all texters are assigned and unmessaged count is zero. Additionally, onTouchTap replaced with onClick because of Matrial UI and React updates.


shakalee14 (approved):
LGTM! - tested locally

Issue #688, opened 2018-07-07
In CampaignTextersForm, typing a number greater than the number of unassigned contacts causes all contacts to be unassigned


lperson:
> @bchrobot [6:14 PM]
> There is a bug where if you are trying to assign people and you accidentally type a number greater than the available unassigned texts, everyone gets unassigned

shakalee14:
Handled with #689

PR #689, opened 2018-07-07
Don't wipe out all assignments when entering a number of contacts > unassigned contacts

lperson:
This is a proposed fix for #688 

@bchrobot thoughts?

### DIFF
--- a/src/components/CampaignTextersForm.jsx
+++ b/src/components/CampaignTextersForm.jsx
@@ -4,6 +4,7 @@ import Slider from './Slider'
 import AutoComplete from 'material-ui/AutoComplete'
 import IconButton from 'material-ui/IconButton'
 import RaisedButton from 'material-ui/RaisedButton'
+import Snackbar from 'material-ui/Snackbar'
 import GSForm from '../components/forms/GSForm'
 import yup from 'yup'
 import Form from 'react-formal'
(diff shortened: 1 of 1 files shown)


src/components/CampaignTextersForm.jsx
```diff
@@ -165,23 +175,17 @@ export default class CampaignTextersForm extends React.Component {
     })
 
     let extraTexterCapacity = totalNeedsMessage + totalMessaged - this.formValues().contactsCount
-    if (extraTexterCapacity > 0) {
-      // 2. With extraTexterCapacity beyond contacts, remove contact counts from the top texters
```
bchrobot: I think the step 2 instructions should stay, but just get updated. It's a little confusing otherwise as there are still comments about step 1 and step 3

src/components/CampaignTextersForm.jsx
```diff
@@ -165,23 +175,17 @@ export default class CampaignTextersForm extends React.Component {
...
 
-      newFormValues.texters = newFormValues.texters.map((texter) => {
-        const newTexter = texter
-        const messagedCount = texter.assignment.contactsCount - texter.assignment.needsMessageCount
-        if (texter.id === theTexter.id) {
-          newTexter.assignment.needsMessageCount = this.formValues().contactsCount - totalMessaged
-        } else {
-          newTexter.assignment.needsMessageCount = 0
+    if (extraTexterCapacity > 0) {
+      this.setState({ snackbarOpen: true, snackbarMessage: 'All remaining contacts assigned to this texter' })
```
bchrobot: Maybe move this after the `newFormValues.texters.map` so that `changedTexter`'s name can be grabbed and included in the Snackbar?

Side effects like that within a `map` aren't great so maybe renaming `chagedTexter` --> `changedTexterId` (`state.focusedTexter` should also really be `state.focusedTexterId` as well in my opinion) and then finding the actual `changedTexter` object with an `Array.find()`?

shakalee14 (approved):

Issue #664, opened 2018-06-21
Per-organization OSDI settings


harpojaeger:
A forthcoming PR will enable setting OSDI API URL and token on a per-organization business. Decisions that need to be made before this is ready:

1. Who can modify OSDI settings?
2. Who can see the current OSDI settings?
3. Should an environment variable be required to even expose the OSDI settings GUI in the first place?
4. Should sensitive OSDI information be displayed in the settings GUI, or should it disappear after being sent to the server?

harpojaeger:
My proposed answers:

1. Admins and owneres
2. Admins and owners
3. No. The "enable OSDI integrations" boolean can default to FALSE. Since only admins and owners can see or modify these settings, requiring them to jump through the additional hoop of setting an environment variable seems like a waste of time (and it defeats the point of having a GUI for this, which is to _prevent_ users from needing to get into the details of the deploy to modify settings.
4. I'm agnostic about this. Masking the API token is a good idea from a data security perspective, but it adds technological and interface complexity.

schuyler1d:
I'd like to counter propose:

1. Who can modify OSDI settings (assuming 3 is enabled)?
 - Owners only: If someone who is an admin should have this access, they can be upgraded to an owner.  Generally this is a one-time setup
2. Who can see the current OSDI settings?
 - No one -- it should be visible whether they are set.  If OSDI settings include auth credentials, the server should never leak them back to the client -- it exposes too much risk
3. Should an environment variable be required to even expose the OSDI settings GUI in the first place?
 - Yes. environment variables are set by the person hosting the system.  They know who their owners and admins are -- generally they are campaigners and know very little about systems.  Exposing this feature will only be appropriate for certain host systems.  It is not an 'extra' hoop -- whoever sets the system up makes a decision, and then there's no hoop for the owners/admins themselves.  I'm even fine with it being default off, but our default app.json for heroku setting it to ON.  I just don't want to expose something that existing systems may want/need to control.
4. Should sensitive OSDI information be displayed in the settings GUI, or should it disappear after being sent to the server?
 - I think my preference would be that it converts to a password field, to indicate that it will be secret from now on. But if you're talking about right after save on the screen of the person that entered the setting, I don't mind if it 'stays' on the screen before they navigate away.

harpojaeger:
For reference, this is what the settings currently look like.
[...]

Issue #597, opened 2018-05-23
OSDI record_canvass_helper action handler


harpojaeger:
An OSDI `record_canvass` handler would connect conversations in Spoke to, e.g., survey questions in VAN.

Ultimately, this will entail updates to the database model for questions to accommodate an `external_id` for both questions and responses (see, e.g. https://developers.ngpvan.com/osdi#people-post-people--personid--record-canvass-helper).

harpojaeger:
Also generated from a discussion at Prog. Hack Night. This is a more involved update than #596. To be production-ready, the questions UI would need to expose the `external_id` fields for editing, or, even better, use the OSDI questions collection GET endpoint (https://opensupporter.github.io/osdi-docs/questions.html#scenario-retrieving-a-collection-of-question-resources-get) to let the user graphically map the desired question into the script somehow.

joemcl:
@harpojaeger really cool, cc: @deasterdaywfp @codygordon @j-ro

harpojaeger:
I mostly created this issue to keep notes for myself on what this feature will ential, but I do want to build it (I think #596 would come first).

harpojaeger:
When creating a new campaign, present the user with an option to import survey questions from an OSDI instance, and then automatically create questions on the Spoke script that map to OSDI questions.

harpojaeger:
The quick-and-dirty version of how this works: 1) at campaign creation, give user a button that says something like "fetch OSDI questions" or whatever. 2) clicking the button fetches the questions. 3) probably provide some search/filter functionality to make it easier to find the desired question. 4) user can select a question and it's added into the script, along with the various question and response IDs necessary for an action handler to send responses back to the OSDI system.

harpojaeger:
In order to map questions to OSDI survey questions, I need to store the question ID and response ID somewhere in the interaction steps. I can see three possible ways of doing this:

1. Use the question and response IDs as the interaction step IDs in Spoke. This is a bit hacky. It entails creating a `mapQuestion` function that, given an OSDI question object, creates new interaction steps and sets their IDs accordingly (probably entailing slight modifications to the current `addStep` function [here](https://github.com/MoveOnOrg/Spoke/blob/main/src/components/CampaignInteractionStepsForm.jsx#L64-L74), perhaps to add a default `newStepID` param or the like – it would be backwards-compatible).
**Pros of this approach**: Little changes to existing architecture, no DB schema updates.
[...]

Issue #526, opened 2018-02-26
Use `kue` instead of jobs table


ben-pr-p:
My understanding of the current setup is that right now, asynchronous jobs are accomplished via a jobs table, and there is a separate process (or separate Lambda event) to go and process them every 5 minutes (or how often one configures their lambda events).

Advantages of dropping the Postgres solution and moving to Redis (`kue`)[http://automattic.github.io/kue/]:
* Jobs processed instantly / based on throughput via Redis pubsub, meaning under low load, there is ms delay, and under high load one can configure the processing time to increase without increasing the load in the DB (which is what would potentially degrade performance in the unaffected parts of the app), solving any outstanding problems with #255.
* That nice GUI!
* Better developer experience for adding jobs, via their callback registering system. Right now adding a job requires updating several different data structures
* Identical processing interface for single and multi-core runtime configurations

schuyler1d:
When JOBS_SAME_PROCESS=1 most of the 'jobs' don't need to be a job at all.  Maybe that should be the first step?

If we use Redis, I would want to make sure it could be optional (i.e. ok to stay in the DB, by default) and independent of the intended Redis caching server intended for the [scaling work](https://github.com/MoveOnOrg/Spoke/pull/524), since the long-running jobs (csv processing) deal with large storage and I wouldn't want that to make the caching size unstable.

From MoveOn's perspective the most important job to track is actually the datawarehouse version of the contact upload -- everything else could just be synchronous (though the regular csv contact upload might be usefully async).

Are you planning on expanding background tasks and want a proper system for it, or are you focusing on the current ones?

ben-pr-p:
Both – I think developing / modifying the current jobs, understanding what the current jobs do, and adding new jobs would be easier with a external job management queue library than the in-house solution that exists right now (we would also get a bunch of features for free that we'll probably want in the future).

Some things we can improve:
* Clarity in code around where jobs are and which function does what job
* Number of files / places that must be changed in order to add a job
* Number of if statements / options switched by JOBS_SAME_PROCESS. Ideally, there should be only one or 0 if possible switches. 

Some solutions / options – (kue)[https://github.com/Automattic/kue] and (bull)[https://github.com/OptimalBits/bull] are the most feature rich, but I do agree with keeping Redis optional.

[...]

PR #690, opened 2018-07-08
OSDI survey question mapping & syncing

harpojaeger:
This is a work in progress pull request to enable mapping OSDI questions into a script and syncing responses back.


harpojaeger:
Column naming issues solved, so this is ready to merge. Closes #597.

harpojaeger:
Also closes #664

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "src/server",
   "engines": {
     "node": "6.10.3",
-    "npm": "3.10.10"
+    "npm": "5.6.0"
```
bchrobot: ❤️

src/components/CampaignOSDIQuestionFetcher.jsx
```diff
@@ -0,0 +1,101 @@
...
+    this.props.mapQuestion(options)
+  }
+
+  render() {
+    const { osdiQuestions } = this.props.osdiQuestions.organization
+    const questionsAvailable = osdiQuestions.length > 0
+    const { description, responses } = this.state.selectedQuestion
+    return (
+      <div>
+        {questionsAvailable ?
```
bchrobot, line 65: Perform this check outside of the return block and have two returns instead? Removes a level of nesting/conditional within the JSX which I think improves readability

bchrobot (commented):
Looks pretty freakin sweet to me! Looking forward to actually playing around with it with an API key eventually. Hopefully. Maybe...

Most of my comments are nit-picking style and indentation and can probably be ignored.

There are lots of `console.log`s still present which can be removed it seems like?

harpojaeger:
BTW thanks for the code review @bchrobot. I've been holding off on making substantial changes until Salazar tests this out in the wild. Just starting to get some feedback today, so I'll return to it soon.

harpojaeger:
@bchrobot and I are working on getting this ready for merge after the various GraphQL changes that were made in the last week or so.

harpojaeger (commented):
Think I've addressed all these now.

harpojaeger:
Back home tonight and will take a look then.

harpojaeger:
@shakalee14 It's ready.

harpojaeger:
Yeah this needs some work to be ready to merge, and I want to add tests for it too.

harpojaeger:
Okay @shakalee14 & @schuyler1d, this is tested and ready for review!

bchrobot (commented):
This review was on the diff since my last review. I realized halfway through that this includes changes made in `main` that are not part of your PR, so preemptive apologies for off-topic comments.

I think I have mentioned this elsewhere, but copying here for posterity/completeness. A campaign used this in August against VAN. They ran into trouble trying to map multiple questions and mixing/matching Spoke survey answers with OSDI sync actions across the two OSDI survey questions. This resulted in response `InteractionStep` rows with a value for `external_response`, but no `external_question` value in the parent `InteractionStep`.

A better approach may be to store both `external_question` and `external_response` in response `InteractionStep`s. The "map question" action would then be a helper to create the appropriate Spoke `InteractionStep` tree with OSDI mappings, but wouldn't require that parent/child split of information. This would require some UI changes for each InteractionStep to both an OSDI survey question and and response. Organizations don't always get their survey's correct the first time but get in too deep to switch horses mid-stream so this seems to be a viable use case.

Last comment from me: some OSDI providers (VAN) rate limit and charge per request so this may need so language warning users about those risks.

joshco:
I'm trying to test this on my dev machine in a docker container pointing to another OSDI implementation.  I fill in the AEP URI and API token, and I see the Interactions part of the campaign UI. However, I don't see any network traffic to the OSDI endpoint, and and menu for the questions don't show.

Is there something specific I need to do in my environment?  Do I need lambda or anything else?

When is the retrieval of questions supposed to happen? On save of settings? On campaign edit load?

In the action handler, at line 60...
````
    // interactionStep has the response ID, but we need to retrieve the question ID
    const question = await getexternalQuestionByInteractionStepId(interactionStep.parent_interaction_id)
    if (!question) throw new Error('Could not retrieve the mapped OSDI question id.')
    const body = {
      canvass: {
        action_date: (new Date()).toISOString(),
        contact_type: 'spoke',
        success: true
      },
      add_answers: [
        {
          question,
          responses: [interactionStep.external_response]
        }
      ]
    }
````
It looks like VAN the VAN implementation used the "add_answers" attribute, whereas the OSDI spec says it should be "add_questions_responses_uri" (http://opensupporter.github.io/osdi-docs/record_canvass.html#question-response-object-fields )

Not sure if others have implemented this, but we should either change the spec or change the VAN implementation.  The spec on the helper is also vague about sending the key vs the title of the response, which we should fix as well.

harpojaeger:
There's an undocumented environment variable called `ACTION_HANDLERS` – do you have that set? I believe it would need to be set to `osdi-survey-question`, but it's been a while since I ran this.

Sounds like the VAN API needs to be updated to match the OSDI spec.

Question retrieval happens during the script creation process.

More generally, this PR is pretty stale at this point. It contains a lot of usable parts, but I don't think I want to merge it into `main` until it has intelligent error handling & network retries. Currently, it fails pretty silently; using it in production could result in lost data if the OSDI server is unavailable at times. I also think it should wait until #526 is done; the network requests should really be dispatched by some central job handler.

joshco:
Ah, that did it. With that set, i can map OSDI questions.  For whoever continues this at some point, the code should handles the choice based questions, but chokes on paragraph. (free form where the possible responses array is not present).

Is there an in progress PR on OSDI list exchange?

harpojaeger:
Huh, I didn't know that was in the OSDI spec. Could be interesting to implement it. FYI, @bchrobot and I are probably the only people who've run this on a real campaign.

My goal is to keep this branch up to date and mergeable, but not to do further work until some if the underlying infrastructure is in place. It also needs testing. Do you have any existing code that stubs or mocks an OSDI API for testing purposes?

joshco:
I thought I already responded, but must have forgotten.  You can check out the OSDI sandbox, which is a working OSDI front end attached to a mock CRM.  it supports most resources and helpers, and on the admin dashboard UI side let's you edit resources at the OSDI JSON level.  It autogenerates a sample data set, has automation to stir the data (updates, new records) so a client can have a realistic experience.

Via the OSDI endpoint, you can save the data and responses into files for mocking, or use node-vcr to record and replay them.

create an account here: http://public.demo.osdi.io

schuyler1d:
The action-handler extensions from @lperson 's recent work are similar to some of this.  If someone updates this functionality to for our current repo, we'll take it.  Closing this because it's too far from `main`

PR #691, opened 2018-07-08
Fix dev-nowrap

### LINKED ISSUE #671: Should node-foreman hard-wrap console output?
I find this kind of annoying; since every hard-wrapped line is prepended with a timestamp, logging is harder to understand. This is easy to disable (I have `"dev": "nf start --trim 500 --procfile ./dev-tools/Procfile.dev",` in my `package.json`). Curious if others who are regularly doing dev work agree. If so, happy to make a PR.

harpojaeger:
Without `-w` or `--trim` specified, `nf` defaults to trimming logs to the value of `tput cols`. This is undesirable. `--trim 500` is an effective workaround.
Closes #671.


schuyler1d (approved):
:+1: thanks!

PR #692, opened 2018-07-08
Implement a knexfile

### LINKED ISSUE #683: Implement a knexfile
Currently, migrations are written by hand in `/src/migrations/index.js'. Implementing a [Knexfile](https://knexjs.org/#knexfile) would have several advantages:

- ability to run `knex migrate:create` to get a migration stub file with all the necessary scaffolding,
- head off a lot of merge conflicts, since migration stub files are timestamped to the second,
- define production, staging and development databases centrally,
- possibly allow phasing out of `r.getCount()` (I haven't investigated this deeply, but it's a possibility)

This would be a big step towards completely removing the legacy DB system, which is a larger project.

Ultimately, I suspect the right way to do this is to write a single Knex "initial" migration that recreates the entire structure of the database as it currently exists. That would allow complete removal of the old migration system, and all future migrations would be done in Knex.

If there are reasons for preserving the existing migration system of which I'm not aware, I'd like to hear about them here! I think this is another issue under the "improve developer experience" heading.

harpojaeger:
Will close #683 when complete.


knexfile.js
```diff
@@ -0,0 +1,63 @@
+const { DB_USE_SSL, DB_TYPE, DB_JSON = global.DB_JSON, DB_HOST, DB_PORT, DB_NAME, DB_PASSWORD, DB_USER, DB_MIN_POOL = 2, DB_MAX_POOL = 10, DATABASE_URL } = process.env
+
+const useSSL = DB_USE_SSL && (DB_USE_SSL.toLowerCase() === 'true' || DB_USE_SSL === '1')
+
+// TODO if useSSL then pg.defaults.ssl = true (figure out whether pg should be imported here)
```
bchrobot: Maybe include a link to the Knex bug tracking this?
https://github.com/tgriesser/knex/issues/852

harpojaeger: Yeah good idea. I've used this workaround in previous projects without even knowing it was a workaround for a bug – I thought it was just how you configured SSL with Knex!

harpojaeger:
I'd appreciate a sanity check from, well, anyone who knows the database really well. In particular, I'd like at least one more pair of eyes on my [custom `jest` matcher](https://github.com/harpojaeger/Spoke/blob/0514a1ebf5f556774b63163aa86b8a004bfa9161/__test__/server/db/knex.test.js#L26-L70). It performs a series of checks on JSON generated by `knex('tablename').columnInfo()` to ensure that the schema generated by the new knex `init` migration is the same as the schema that already exists. I have already used it to diagnose and fix problems with the migration that creates the `log` table.

Here's how to use it:

1. Make sure your development DB is up to date with the most recent migration.
2. From `/__test__/server/db/schemas`, run `export.js` (probably like this: `npx babel-node export.js` [[npx](https://www.npmjs.com/package/npx) is awesome]). You will get a lot of JSON files representing the schema of all the tables currently in the DB.
3. Make sure you've set up the test DB as described in the docs.
4. `npm run test-knex` should result in a single passed test for the `log` table.

The reason you have to run `export.js` separately is that it needs to happen in the development environment, so it can access the development DB. I tried doing this programmatically inside the test and gave up after banging my head against it for a while. This seems good enough to me.

You could uncomment [these two lines](https://github.com/harpojaeger/Spoke/blob/0514a1ebf5f556774b63163aa86b8a004bfa9161/__test__/server/db/knex.test.js#L16-L17) if you want to see the actual JSON representations of the schema that are being compared.

I think these instructions cover it all. I'm particularly interested in feedback on the custom matcher I wrote. If there's a logical flaw, I want to catch it _before_ I start rebuilding the rest of the schemas in Knex. If we do this right, it's an immediate, seamless replacement for `r.knex` and the old migration system, and ultimately will let us phase out the entire `rethink-knex-adapter` if we want to.

harpojaeger:
Also, before anyone asks: yes, using `_.isEqual(oldSchema, newSchema)` would have been a faster way to do this, but I wanted explicit, verbose feedback on how the schemas differed.

schuyler1d:
so i'm intrigued.  I wonder if some of db/schemas/export.js can/should go into the rethink-knex-adapter, though I guess from your perspective, that might be a dependency we eventually want to slough off?

Should I worry that knexfile.js looks hairy or is that standard boilerplate?

I think before we merge this direction, I'd want to see it actually making the tests use it and behave more consistently.  If you can connect those dots, I think it's a slam dunk.

harpojaeger:
> I wonder if some of db/schemas/export.js can/should go into the rethink-knex-adapter, though I guess from your perspective, that might be a dependency we eventually want to slough off?

I'm agnostic about whether the `export.js` functionality should live somewhere else. An advantage to moving it is that there's one less step needed to obtain an up-to-date-schema upon which to base tests. Disadvantage: clogging up runtime code with unnecessary `fs.writeFile()`s, and the general smelliness of doing non-application-runtime-related-things at runtime.

To clarify (because from your above-quoted comment I think I didn't really say this, which is on me): I see this as essentially a one-time test. After we have verified sufficiently that the new, Knex-generated schema matches, we can deprecate this test.

At least, that's how I've _been_ conceiving of it. If we're taking control of the schema away from `thinky` and giving it to `knex.migrate.latest()`, there will no longer be a reason to ask `thinky` to give us its version of the schema. The Knex migrations will be the canonical representation of the schema.

The proposed change does not obviate the need to keep the `thinky` models up to date. Ultimately, I think I would need to know a lot more about how ORMs work in general (this is the first one I've worked with at any scale) before forming a strong opinion about whether we should replace it in this case. If we *did* decide to replace it, having a strong knex foundation won't hurt (see, e.g. [this](https://github.com/kripod/knex-orm)).

harpojaeger:
Now that I'm building the schema for some more complex tables, I've noticed that knex's `columnInfo()` doesn't capture index data. In fact, it seems there is no way for Knex to check for the existence of a particular index except with raw SQL.

I'm also noticing a disparity between the Thinky model and the actual database schema, on the `message` table (and I think this is going to pop up elsewhere too. Thinky defines `user_number: optionalString()`, and later runs `ensureIndex(user_number)`, but the exported schema JSON has that column as `NOT NULL` with a default of `''::text`, which I think is PSQL for 'a blank string'.

I can replicate this schema with `t.text('user_number').notNullable().defaultTo('')`, so that's not the issue. But I'm a bit concerned about this (apparent) discrepancy. Does this strike anyone as a problem?

As far as the indices, I can't really think of a good way to test them in a database-agnostic way. In pg, we can do `select indexname from pg_indexes;`. The assumption would be that a Knex schema that creates the right indices in postgres will also create the right indices in SQLite. Does that seem like a fair assumption?

schuyler1d:
Well, optionalString is unintuitively NOT NULL -- see:
https://github.com/MoveOnOrg/Spoke/blob/main/src/server/models/custom-types.js#L19-L25
That's mostly to appease the apollo gods that require things to be strings and don't always accept null (not a string) in their place (and also is a legacy spec from pre-MoveOn era with rethinkDB)

I'm actually less worried about getting the indexes -- because our models already declare them -- that's why i was thinking these exports might be better created by rethink-knex-adapter?

harpojaeger:
That makes sense about `optionalString` – seems like I'm at least doing the schema right.

Does rethink-knex-adapter have some method for dumping the database schema that I can test against? I used knex's `columnInfo()` because JSON is convenient to work with in `jest`, but I'm open to rewriting the test if there's a more thorough way.

harpojaeger:
Yeah, good point – although we don't require `DB_TYPE` in all cases, so that wouldn't necessarily be reliable. I'll take a look at `object.schema` and see what it gives me.

harpojaeger:
Kinda stuck on how to actually access `object.schema` in the wild, so if anyone wants to throw me a tip on that I'd appreciate it.

schuyler1d:
The schema IS the model that spoke has, e.g.
https://github.com/MoveOnOrg/Spoke/blob/main/src/server/models/campaign.js#L7
so you can just e.g. `import {Campaign} from './server/models'` or whatever and then iterate over `(new Campaign).fields` and `(new Campaign).indexes`

harpojaeger:
This is throwing `TypeError: Cannot set property '__proto__' of undefined` for me. Oy.

harpojaeger:
Am I understanding `model.ensureIndex()` correctly in that when I see `CampaignContact.ensureIndex('campaign_assignment', (doc) => [doc('campaign_id'), doc('assignment_id')])` [here](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/models/campaign-contact.js#L40), that's creating a multicolumn index called `campaign_assignment` on the `campaign_id` and `assignment_id` columns?

harpojaeger:
Notes from a call with @lperson, @bchrobot & @schuyler1d:

- Continue working on this migration so Knex can take over schema management (MoveOnOrg/rethink-knex-adapter#15 will enable disabling table creation)
- In `test_helpers.js`, see if `setupTest()` and `cleanupTest()` can be updated to use Knex migrations (possibly fixing some race conditions)

Not discussed on the call, but needs doing anyway:
- update `export.js` script to also dump info about foreign keys (it'll probably have to be PSQL-specific, but that's good enough for now)

schuyler1d:
@harpojaeger I made a branch of rethink-knex-adapter that can have the auto-creation of the tables disabled
https://github.com/MoveOnOrg/rethink-knex-adapter/pull/16
This might be useful for this branch.

harpojaeger:
Made good progress today. The migration now correctly creates every table schema! I've started writing a test to check that all the indexes are set up properly. If you're running these tests yourself, you'll need to rerun `npx babel-node export.js` from `/__test__/server/db/schemas` to get the new `indexes.json` file upon which the index tests depend.

This is a couple days from completion.

harpojaeger:
Could end up closing #518 too.

harpojaeger:
FYI, I'm essentially offline until Sunday or Monday.

harpojaeger:
My schema and index tests pass. Has anyone else managed to run my janky schema-exporter thing? Would love to have confirmation from at least a couple others before I declare this ready for merge.

harpojaeger:
Closes #683. Added [comments](https://github.com/MoveOnOrg/Spoke/pull/692/commits/20b813b6bd935f70546876e3856957b44d2f96e7#diff-4e2d2a12f67e4ab6f22993be3acb1affR137) to clarify how to merge this in a way that's compatible with #690.

harpojaeger:
Oh wait, @schuyler1d, do I also need to bump the `rethink-knex-adapter` version and use your new don't-autocreate-tables flag? How does that work?

schuyler1d:
Tracking issues I encountered while checking this out:
* [instructions on how/what to do](https://github.com/MoveOnOrg/Spoke/pull/692#issuecomment-403981502) need to be in a doc integrated into install instructions, etc -- not just a PR comment
* babel-node export.js does not terminate -- it just stays at the end
* index tests seemed to fail because of an assumption about what db backend was used:
```
   Expected "canned_response.canned_response_user_id_index.indexdef" to be 
"CREATE INDEX canned_response_user_id_index ON canned_response USING btree (user_id)", but received 
"CREATE INDEX canned_response_user_id_index ON public.canned_response USING btree (user_id)".
```
* If we use .env explicitly and we run things like `knex.destroy()` it feels like we need some safeguards to make sure people don't accidentally run this on an important DB backend . (not sure answer here) -- 
* related to above, `.env` should not be forced -- it should use the environment vars, and depending on how the user runs things, .env should be loaded

I'll post my full output of errors from `test-knex` in another comment

schuyler1d:
This gist has the full errors list.
https://gist.github.com/schuyler1d/b2b96f44f49b21c6b908384450c0d995

I also wonder whether we could get rethink-knex-adapter to generate the json output

harpojaeger:
Agreed. I need to look back at this and see if the knexfile/Knex migrations system can be implemented separately from the bigger-picture ORM changes we discussed at the end of last year. If so, I'll start cleaning this up and getting it ready for merge.

harpojaeger:
Status update: merged upstream/main and trimmed off some fat, as well as making a config change that makes CI pass (excluding the by-hand Knex schema equality tests from automatic Jest runs, since they can't be run in that environment yet). @schuyler1d has a PR in against the source branch for _this_ PR which bumps the `rethink-knex-adapter` version. Once that one is updated from `MoveOnOrg:main`, I'll take a look and see if I can merge it in. My instinct is that it might be better to do them separately, but I'm not sure yet.

In the meantime, if anyone else can run the "by-hand" Knex schema test I wrote (instructions for doing so are [further up this thread](https://github.com/MoveOnOrg/Spoke/pull/692#issuecomment-403981502)), I'd be interested to know what happens. It's not the most elegant system ever, but this is a tricky thing to test, and I think this approach may be the best we have under the circumstances.

joemcl:
hey @codygordon you've done some stuff in the past re Spoke and knex stuff right? Any bandwidth to help @harpojaeger test the schema?

harpojaeger:
Good stuff, thanks. You can open a PR with your changes to the export script and I'll take a look tonight.

harpojaeger:
I've updated the initial migration so it's caught up with the DB changes that have been made since I started work on this PR. Haven't addressed the index mismatches yet; I'll take a look at that this weekend.

harpojaeger:
Note from my work today: it looks to me like the index test is failing because the foreign keys aren't getting dumped out of the existing DB schema. This is weird; I remember this test passing when I first wrote the migration, and I actually spent quite a bit of time getting the foreign keys right, because they required a fair amount of manual tuning. I suspect that this is a problem with my exporting script – maybe something changed in the way PG stores indices (I've upgrade to PG 11 locally, and I might even have been on 9 last time I ran this) – not a truly failing test.

harpojaeger:
Actually, on even closer inspection, it's not that the foreign keys exist and aren't getting exported; they don't exist _at all_ in my postgres. @bchrobot, you said you also had 30 missing indices when you ran this test. Could be that Thinky isn't creating the indices properly at DB initialization.

harpojaeger:
Mystery solved: there _is_ a problem with the export script. I don't know why it was working before, but the way I was fetching indexes didn't include foreign keys. [This query](https://dba.stackexchange.com/a/37068) does the trick. Refining the export script and test to use it now.

harpojaeger:
This has been updated, and all tests now pass! I've also made a change to the knex setup; `knexfile.js` now lives in `src/server`, and instead of just exporting a knex config that other files can pass to `knex()`, it exports an instance of `knex()`. This seems like a better way of doing things. Interested in feedback on this.

Remaining work to be done on this PR:
1. Switch thinky over to using this new, centrally-defined `knex` instance,
2. Replace any other instances of direct calls to `knex` in the codebase with references to the new `knex` instance, and
2. Merge @schuyler1d's PR (after it's been rebased), or make the `rethink-knex-adapter` changes (disabling table auto-creation) myself.

I think that's everything, but if it looks to anyone like I'm leaving something out, please say so here.

harpojaeger:
Oh, and 4): remove the existing migrations code and update docs on how to generate new ones.

harpojaeger:
Updates on the action items above:
1 and 2) It turns out that the rethink-knex-adapter can only accept a knex config object, not an existing instance of knex. So I rejiggered it such that `/src/server/knex.js` only exports a config object for now. `dumbThinky` is instantiated with that object, and the rest of the application continues to access `r.knex` as before.
3) is pending a rebase of the PR.
4) is done.

As a bonus, I refactored my knex schema equivalency tests to make them less fragile. The schemas for comparison are now in version control, which essentially means they explicitly declare what the database should look like, instead of relying on a developer to generate the schema themself. This seems like a better practice to me. Additional eyes on the schema files are necessary to make sure we're testing for the right things.

harpojaeger:
Quick update: the Travis tests are failing because of discrepancy between `CURRENT_TIMESTAMP` and `now()`. I suspect this is a Postgres version mismatch; I'm running 11.1, and Travis defaults to 9.2. I tried to bump the Travis version, but there are some complications (see https://github.com/travis-ci/travis-ci/issues/5737). I have a comment in on that issue; if I can figure out how to make it work, I may put in a separate PR for incrementing our Travis Postgres version, unless anyone has any flags on that.

Downgrading my Postgres version is kind of a pain. If anyone here is running 9.2 and can run the `export.js` script to recreate the JSON schemas in `__test__/server/db/init_schemas` (which will result in your local working copy of them looking the way they look in Postgres), and then run `npm run test-knex` to verify, that'd be helpful. As a last resort, I can replace `CURRENT_TIMESTAMP` with `now()` in the JSON schemas, but modifying test data to match the computed results is a bad pattern and I'd like to be very sure this is the issue before doing that/avoid it in the first place if at all possible.

schuyler1d:
We should revisit this PR.  If we (or Harpo ^_^) can get it up-to-date, let's move to this. ( @shakalee14 )

schuyler1d:
@bchrobot This looks similar to this: https://github.com/MoveOnOrg/Spoke/pull/1154
(e,g, using knex.env.js) -- but not sure if there are other pieces in your changes?

Note one thing we needed to do is include I think one or two of your knex migration files in the init one.  It would be good if we could align as much as possible.

PR #695, opened 2018-07-09
Don't need organization.assignments ...

lperson:
... in the initial query to populate filter dropdowns.

@schuyler1d Do you want to give this a spin?


schuyler1d (approved):
This makes it work for MoveOn! :+1: :heart:

PR #696, opened 2018-07-10
Message review message fetch optimization

lperson:
@bchrobot @schuyler1d 

This change is an optimization to (hopefully) reduce latency while paging in Message Review.  The intent was to reduce the number of queries.  Instead of doing a separate query to get the messages 
in each conversation, this code gets the messages and conversation details in a single query.  It is now 3 queries for each page, rather than 2 + (number of conversations returned) queries.

Please let me know if this improves performance at scale.

I merged the changes in #678 into main before making these changes.  You may want to do the same before reviewing.


lperson:
@bchrobot is this ok to merge the way it is?  In our slack discussion of this, I made the point that it would be hard to consolidate the 3 queries, and it seemed query consolidation could be deferred to a new future PR.

PR #700, opened 2018-07-11
Iterate warehouse contact loads by STEPs for groups and can delegate to a new invoked lambda for each call

### LINKED ISSUE #672: Datawarehouse contact upload can max-out on short-lived backends (like AWS Lambda)
In https://github.com/MoveOnOrg/Spoke/blob/main/src/workers/jobs.js#L148
The full query is done at once, and contacts are updated in a single go.

For very large queries, this will timeout (or run out of memory or both) in constrained environments like AWS Lambda.

On conditions:
* `SELECT COUNT(*) FROM (<QUERY>)` and yields more than 10-20k contacts
* Somehow we detect having a 'dispatcher' available, (see https://github.com/MoveOnOrg/Spoke/blob/main/lambda.js#L43-L58)

1. Queries should be batched into groups of 10-20k with LIMIT OFFSET appends -- at least when a pre-query runs with 
2. Dispatch to a lambda sub-task one of the OFFSET/LIMIT batches (the SQL query itself will also need to be sent, of course)
3. As it continues, the Job status should be updated.

If that works in parallel, then great -- otherwise, we may need to do it 'serially' -- where a lambda job is sent the first, iterates through the offsets until it's run out of time and then dispatches another one to continue the work.  We should do what's most reliable and will also work predictably and dependably.

schuyler1d:
fixes #672


src/workers/jobs.js
```diff
@@ -183,55 +224,137 @@ export async function loadContactsFromDataWarehouse(job) {
...
-  const optOutCellCount = await r.knex('campaign_contact')
-    .whereIn('cell', function () {
-      this.select('cell').from('opt_out').where('organization_id', campaign.organization_id)
-    })
-    .where('campaign_id', job.campaign_id)
-    .delete()
+  const totalParts = Math.ceil(knexCount / STEP)
 
-  console.log('OPTOUT CELL COUNT', optOutCellCount)
+  if (totalParts > 1 && /LIMIT/.test(sqlQuery)) {
```
mathemagica: Let's add a comment here explaining why: "We're instrumenting this query with limit / offset to implement batching results back from this query, to get back currently scaling limitations. It'd be way more complicated to implement batching with a query that already does some batching itself, so we do the easier thing here and check whether queries contain LIMIT, and if so, give the user an error so they can rewrite this query to not preemptively batch."

lambda.js
```diff
@@ -11,7 +11,7 @@ const jobs = require('./build/server/workers/job-processes')
 // See: http://docs.aws.amazon.com/lambda/latest/dg/best-practices.html#function-code
 // "Separate the Lambda handler (entry point) from your core logic"
 
-exports.handler = (event, context) => {
```
mathemagica, line 14: lets put a PSA comment here that lambda doesnt know its actually done (and may retry on perceived failure) unless handleCallback is called (eek)

src/workers/jobs.js
```diff
@@ -183,55 +224,137 @@ export async function loadContactsFromDataWarehouse(job) {
...
+        .whereIn('cell', function optouts() {
+          this.select('cell').from('opt_out').where('organization_id', jobEvent.organizationId)
+        })
+        .where('campaign_id', jobEvent.campaignId)
+        .delete()
+      console.log('OPTOUT CELL COUNT', optOutCellCount)
+    }
+    await r.table('job_request').get(jobEvent.jobId).delete()
+    return { 'completed': 1 }
+  } else if (jobEvent.part < (jobEvent.totalParts - 1)) {
```
mathemagica, line 276: lots of complexity in here, and when we have a reliable unit test framework that can setup and tear down dbs without race conditions, i'd ask that you create unit tests for cases: single batch < 10K in size, a batch > 10K, a batch of exactly 30K (to check for off by one errors). For now, could you create unit test stubs with descriptive names so we dont forget to unit test this later?

mathemagica (changes requested):

mathemagica (approved):
👍

azuzunaga:
## QA feedback:
### Instructions
1) Hit the + button to create a new campaign
2) Go to the contacts tab, copy paste query (may have to add AS statements for validation.. for example columns should be `first_name`, `last_name`, `cell`) run query
3) Wait for it to load - make sure it completes & try to keep up with time it takes to complete
4) Archive campaign

### Results:
- It takes +1 min to load contacts from any query.
- Can’t tell when contacts are loading, how long until all contacts finish loading, or whether there is a query error.
  - Silent fail on query error (introduced a typo)
- After contacts have been added they don’t show up under *Already texted* or *Available to assign*
  - UPDATE: I actually didn't know how to assign contacts at first, but there is another bug: In campaigns with a large number of contacts (tested at 4.7 million) only one texted can be assigned at a time.
    - Also tested with a campaign with ~50k users and wasn't able to reproduce the bug.
- In the SQL input text field, `SELECT` form validation is case-sensitive.
- Query `select distinct(cell), first_name, last_name from spoke.campaign_contact where is_opted_out=false limit 150000` didn’t load after 15 mins
  -  Query `select distinct(cell), first_name, last_name from spoke.campaign_contact where is_opted_out=false limit 5` loads 5 contacts without issue
- Preliminary: Query `select distinct(cell), first_name, last_name from spoke.campaign_contact where is_opted_out=false` loaded up to 340,000 contacts before the app crashed.

PR #702, opened 2018-07-12
Removes opt outs from initial assignments query - not sure we ever ha…

shakalee14:
…ve isOptedOut filter set to true


schuyler1d (approved):
:+1:

Issue #432, opened 2017-12-14
With no TZ variable set and no timezone_offset, time display defaults to GMT


schuyler1d:
This is confusing behavior and should either hide it in the unknown circumstance or have a more sane default.

bchrobot:
The time display in the toolbar [`src/Components/ContactToolbar.jsx`](https://github.com/MoveOnOrg/Spoke/blob/aae35f862ae2097434afcd3225e0620df864aa33/src/components/ContactToolbar.jsx#L26-L50) is missing any reference to `process.env.TZ` (via `window.TZ`).

Seems like the logic should be about the same as in `isBetweenTextingHours` in [`src/lib/timezones.js`](https://github.com/MoveOnOrg/Spoke/blob/aae35f862ae2097434afcd3225e0620df864aa33/src/lib/timezones.js#L23-L28).

Perhaps it makes sense to extract the functionality of checking `process.env.TZ`, the contact's zip code, and then reverting to a default to a utility method like `timezoneForContact(contact)`. This could then be used in `ContactToolbar.jsx` as well as [within `AssignmentTexterContact.jsx`](https://github.com/MoveOnOrg/Spoke/blob/aae35f862ae2097434afcd3225e0620df864aa33/src/containers/AssignmentTexterContact.jsx#L505-L509).

ibrand:
This appears to have been fixed in #704 so I'm closing it

PR #704, opened 2018-07-14
Handle TZ defaults in case of no TZ info (432)

bdatkins:
Fix for https://github.com/MoveOnOrg/Spoke/issues/432

- Introduced new helper method getContactTimezone (in /lib/timezones.js), which determines if there is timezone info available. If so, just return it. If not, try using the process.env.TZ info.  If that's not available, use the fallback timezone config.  Between these two options, the timezone info will always be populated.
- Called getContactTimezone from ContactToolbar and AssignmentTexterContact to handle possible missing timezone info

### DIFF
--- a/src/containers/AssignmentTexterContact.jsx
+++ b/src/containers/AssignmentTexterContact.jsx
@@ -30,6 +30,8 @@ import { withRouter } from 'react-router'
 import wrapMutations from './hoc/wrap-mutations'
 import Empty from '../components/Empty'
 import CreateIcon from 'material-ui/svg-icons/content/create'
+import { getContactTimezone } from '../lib/timezones'
+
 
 const styles = StyleSheet.create({
   mobile: {
@@ -511,7 +513,11 @@ export class AssignmentTexterContact extends React.Component {
       const { hasDST, offset } = contact.location.timezone
 
       timezoneData = { hasDST, offset }
+     } else {
+        const { hasDST, offset } = getContactTimezone(location).timezone
+        timezoneData = { hasDST, offset }
     }
+
     const { textingHoursStart, textingHoursEnd, textingHoursEnforced } = campaign.organization
     const config = {
       textingHoursStart,
--- a/src/containers/AssignmentTexterContact.jsx
+++ b/src/containers/AssignmentTexterContact.jsx
@@ -32,7 +32,6 @@ import Empty from '../components/Empty'
 import CreateIcon from 'material-ui/svg-icons/content/create'
 import { getContactTimezone } from '../lib/timezones'
 
-
 const styles = StyleSheet.create({
   mobile: {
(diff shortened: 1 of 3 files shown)


src/lib/timezones.js
```diff
@@ -12,6 +12,28 @@ const TIMEZONE_CONFIG = {
...
+export const getContactTimezone = (location) => {
+
+  if (location.timezone == null || location.timezone.offset == null) {
+    let timezoneData = null;
+    let offset;
+    let hasDST;
+
+    if (getProcessEnvTz()) {
+      offset = moment().tz(getProcessEnvTz()).format('Z');
+      isDst =  moment().isDST();
```
bchrobot: s/isDst/hasDst

src/containers/AssignmentTexterContact.jsx
```diff
@@ -511,7 +513,17 @@ export class AssignmentTexterContact extends React.Component {
       const { hasDST, offset } = contact.location.timezone
 
       timezoneData = { hasDST, offset }
+     } else {
+        let location = getContactTimezone(contact.location)
+        if (location) {
+          let timezone = location.timezone
+          if (timezone) {
+            const { hasDST, offset } = timezone
```
bchrobot: Any reason not to just have `timezoneData = timezone`? (I don't think either value is mutated later on)

src/containers/AssignmentTexterContact.jsx
```diff
@@ -30,6 +30,8 @@ import { withRouter } from 'react-router'
 import wrapMutations from './hoc/wrap-mutations'
 import Empty from '../components/Empty'
 import CreateIcon from 'material-ui/svg-icons/content/create'
+import { getContactTimezone } from '../lib/timezones'
+
```
bchrobot: While it's not in our `.eslintrc` and the base AirBnB config allows two empty lines, I believe Spoke convention is for a max of one empty line

bchrobot (changes requested):
Overall, looks good!

src/lib/timezones.js
bdatkins: Can you provide some more info? I don't see a hasDst in the moment() docs.

bchrobot: I mean the variable you're assigning it to -- sorry, that wasn't very clear haha

bdatkins: Der, of course, silly me.  Coding while tired, sorry!

src/containers/AssignmentTexterContact.jsx
bdatkins: Good catch, simplified and removed the extra assignment

shakalee14 (approved):

bdatkins:
Please ignore the various commits from this evening, user error.

PR #705, opened 2018-07-14
Lock pending messages while sending

bchrobot:
This is to address concurrency problems where multiple workers each end up sending all `PENDING`/`QUEUED` messages.

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -606,7 +606,7 @@ export async function sendMessages(queryFunc, defaultStatus) {
           const service = serviceMap[message.service]
           log.info(`Sending (${message.service}): ${message.user_number} -> ${message.contact_number}\nMessage: ${message.text}`)
           // TODO: pass transaction to service.sendMessage
-          await service.sendMessage(message)
+          await service.sendMessage(message, trx)
           pastMessages.push(message.id)
           pastMessages = pastMessages.slice(-100) // keep the last 100
         }
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -592,7 +592,8 @@ export async function sendMessages(queryFunc, defaultStatus) {
     if (queryFunc) {
       messageQuery = queryFunc(messageQuery)
     }
-    await messageQuery.orderBy('created_at')
+
+    messageQuery.orderBy('created_at')
       .then((messages) => {
         for (let index = 0; index < messages.length; index++) {
           let message = messages[index]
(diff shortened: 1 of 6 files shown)


joemcl:
Hey @schuyler1d @shakalee14 @mathemagica can we possibly fast track review of this? Might be the fix for current dupes issue. cc: @deasterdaywfp @ErikKirkstein

src/workers/jobs.js
```diff
@@ -583,26 +583,52 @@ export async function exportCampaign(job) {
...
+        // sending them. We will exit gracefully in that case.
+        trx.rollback()
+        return
+      }
+
+      try {
+        for (let index = 0; index < messages.length; index++) {
+          let message = messages[index]
+          if (pastMessages.indexOf(message.id) !== -1) {
+            throw new Error('Encountered send message request of the same message.'
```
schuyler1d, line 611: Have you hit this error?

schuyler1d (approved):
This looks ok, fwiw.

@deasterdaywfp 
If WFP and other groups are using multiple 'messageSending' threads at once, then I don't think this would be sufficient -- I think you would want to just break up the message sending with the workers:
message-sender-234.js
message-sender-56.js
message-sender-789.js
(and not send on the 'main' worker thread)

MoveOn might be seeing this and doesn't use multiple workers, so not sure this will actually solve the 'root' issue #873

src/workers/jobs.js
bchrobot: No, but that error was not added in this PR

joemcl:
thanks @schuyler1d for taking a look. I've merged into our staging Spoke and testing it now. We have 6 web dynos running and 1 messagesender dyno - do you mean running a messagesender dyno might be a bad idea? cc: @deasterdaywfp

joemcl:
and @schuyler1d see this - we did not get that merge conflict - 
https://github.com/WorkingFamilies/Spoke/pull/56

schuyler1d:
I'm closing this, because we determined that the root cause was that multiple incomingMessage jobs were running at once. while JOBS_SAME_PROCESS was *also* enabled -- don't do that :-)
This code is also undergoing enough changes in the near future that it will cause merge conflicts (see branch `scaling-messages`.
I think we should maybe better document the jobs modes we have.  It looks like a lot of our code assumes JOBS_SAME_PROCESS anyway and thus we should maybe just force that mode going forward -- at least for incomingMessages.

PR #708, opened 2018-07-16
Refactoring -- no functional changes

lperson:


lperson:
@bchrobot I don't think infinite scrolling is what we're looking for.  Wouldn't autocomplete be more suitable for the use case?  Consider: 10,000 texters in a large campaign.  I did an iteration of a naive autocomplete that repeatedly goes to the database and it was really slow.  Multiple seconds after each letter was typed.

It's easy to paginate a large dataset -- I did an iteration of that, too -- but it's still slower than bringing back the whole thing as a string.

I also thought about server-side rendering for large lists.

PR #710, opened 2018-07-16
Attribute messages to the texter who hits send

bchrobot:
Now that we have introduced campaign contact re-assignment, the `user_id` of the assignment is no longer a valid way to determine which user actually hit send. This is important for quality control, especially in largely volunteer-driven campaigns where bad actors may sign up to volunteer with little to no vetting.

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -1083,6 +1083,7 @@ const rootMutations = {
       const messageInstance = new Message({
         text: replaceCurlyApostrophes(text),
         contact_number: contactNumber,
+        user_id: campaignContactId,
         user_number: '',
         assignment_id: message.assignmentId,
         send_status: JOBS_SAME_PROCESS ? 'SENDING' : 'QUEUED',
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -1023,7 +1023,7 @@ const rootMutations = {
 
       return []
     },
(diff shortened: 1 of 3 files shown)


src/server/models/message.js
```diff
@@ -2,10 +2,14 @@ import thinky from './thinky'
...
 import { requiredString, optionalString, timestamp } from './custom-types'
 
+import User from './user'
 import Assignment from './assignment'
 
 const Message = thinky.createModel('message', type.object().schema({
   id: type.string(),
+  // Assignments may change, so attribute the message to the specific
+  // texter account that sent it
+  user_id: type.string().allowNull(true),
```
bchrobot, line 12: Seems like this should work for an optional foreign key based on [this part of `toKnex`](https://github.com/MoveOnOrg/rethink-knex-adapter/blob/37bb13346f13f1971b39e0ecf05fcc7e5430c3a3/models.js#L347-L358).

src/server/api/schema.js
```diff
@@ -1083,6 +1083,7 @@ const rootMutations = {
       const messageInstance = new Message({
         text: replaceCurlyApostrophes(text),
         contact_number: contactNumber,
+        user_id: user.id,
```
bchrobot, line 1086: I believe this is the only place it needs to be added?

bdatkins (approved):

Issue #709, opened 2018-07-16
Texter Assignment Bug


shakalee14:
Introduced with last stage-main merge

1. Text assignments do not always equal amount which appears in a texter's to-do menu
2. A couple texters reporting getting batches that were not assigned to them
3. Campaigns are indicating they are closed (all texts assigned) but there are still texts available when you open the texter screen, and when you try to assign them, they do not appear in the texter's menu

joemcl:
Thanks @shakalee14 . So, for example, the screenshot below reports a count of 9 contacts still to be sent a first message ( 2,869 of 2,878 contacts completed) but only 2 that could be assigned. So is the count being reported incorrectly, or are there actually 9 contacts still to be sent a first message? 
![20180716spoketextcounts](https://user-images.githubusercontent.com/3912434/42782962-52e0438a-88ff-11e8-8352-017a4e4c1570.jpg)

ErikKirkstein:
Currently in MoveOn’s Spoke, the amount to texts assigned to a texter does not always correlate with the amount they see in their to-do menu.

EX:

Text Team Admin assigned Texter A 500 texts.

Texter A opens their to-do screen and see “481” texts, or “462” texts.


This issue becomes more stark in campaigns where more text texts have already been assigned. So you’ll notice it more in a campaign that has, say, 6362 of 9981 texts assigned, as opposed to 575 of 9981 texts assigned

shakalee14:
Fixed with https://github.com/MoveOnOrg/Spoke/pull/731

PR #715, opened 2018-07-17
Fix issue 709

lperson:


lperson:
To reproduce: 

1. Prerequisite: fully assigned campaign.  (This may not be necessary in the minimum steps to reproduce.)
2. Change the needs message number for a texter who has already sent texts.  
3. Note that the number of texts sent by the texter changes to 0.

This may or may not be a fix for #709.  I couldn't reproduce it.

PR #716, opened 2018-07-17
default to country=US (even if process.env.PHONE_NUMBER_COUNTRY is not set)

schuyler1d:
otherwise known as the 'MERICA patch :-P


shakalee14 (approved):

shakalee14:
Tried unsetting this on stage and client still received `Upload at least one contact`

schuyler1d:
meh

PR #718, opened 2018-07-18
Cleans up logic for skip reply button - should only show if it needs …

shakalee14:
…a response


bchrobot (approved):
👍

bdatkins (approved):

shakalee14:
TODO:
- [ ] Mobile UX Update

PR #722, opened 2018-07-18
Backwards compatibility for roles query

bchrobot:
As new roles are introduced, the [initial set of roles](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/schema.js#L847-L853) created for a user/organization may become out of date. This currently requires explicit calls to the `editOrganizationRoles` mutation to correct the set of roles. Instead, check for all roles as high or higher than the requested role.

Specifically, this was causing problems when visiting the site index. Admins were getting a message that they didn't belong to any organizations and had to manually fix the URL.

Also includes query refactor to use knex.

### DIFF
--- a/src/server/api/user.js
+++ b/src/server/api/user.js
@@ -1,5 +1,6 @@
 import { mapFieldsToModel } from './lib/utils'
 import { r, User } from '../models'
+import { rolesAtLeast } from '../../lib/permissions'
 
 export function buildUserOrganizationQuery(queryParam, organizationId, role) {
   const roleFilter = role ? { role } : {}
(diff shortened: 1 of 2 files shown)


shakalee14 (approved):

src/server/api/user.js
```diff
@@ -33,12 +34,15 @@ export const resolvers = {
       if (!user || !user.id) {
         return []
       }
-      let orgs = r.table('user_organization')
-        .getAll(user.id, { index: 'user_id' })
+      let orgs = r.knex.select('organization.*')
```
schuyler1d, line 37: Can you add a method to get this info in:
src/server/models/cacheable-queries.js
which leverages cache data `texterinfo-${userId}` when available (and skips the db)?

shakalee14: hey @schuyler1d - you want the info in this query to be cached (when possible and looked at if available)?

```
let orgs = r.knex.select('organization.*')
        .from('organization')
        .join('user_organization', 'organization.id', 'user_organization.organization_id')
        .where('user_organization.user_id', user.id)
```

schuyler1d: The information for this query will already be in the cache if it exists.  The query here is basically `userOrgsWithRole(role)`  We should implement this funciton in cacheable-queries/ so that
```
async function userOrgsWithRole(role) {
  if (r.redis) {
     const userKey = `texterinfo-${userId}`
     const userOrgs = await r.redis.hgetallAsync(userKey)
     if (userOrgs) {
        // below we return all keys with values equal to or higher than the requested role
       // turn the dict into an array and filter 
       return orgs
    }
  }
  let orgs = r.knex.select('id')
        .from('organization')
        .join('user_organization', 'organization.id', 'user_organization.organization_id')
        .where('user_organization.user_id', user.id)
  return orgs
}
```
Also note we only call that resolver asking for organization id, so I don't think we should bother loading any of the other data until we need it.

shakalee14: I think we need the name as well @schuyler1d for the resolver.

PR #727, opened 2018-07-19
Fix join compatibility for SQLite

bchrobot:


schuyler1d (approved):
yaas!  thanks!

bdatkins:
Do we know why a right join was used originally?  Wouldn't want to lose any functionality by making this change.

bdatkins (approved):

PR #728, opened 2018-07-19
[WIP] Convert SQLite's unix timestamp to date/time format

bchrobot:
SQLite bug due to differences in how `date` vs `timestamp` are stored in Postgres vs SQLite:
```js
// https://github.com/MoveOnOrg/Spoke/blob/stage-main/src/server/models/campaign.js#L15-L19

  due_by: type
    .date()
    .required()
    .default(null),
  created_at: timestamp(),
```

<details open>
<summary>Logging the `campaigns` resolver for an organization returns:</summary>

```
[ { id: 2,
    organization_id: 2,
    title: 'Date Format Errors',
    description: 'Does this error with SQLite?',
    is_started: 0,
    due_by: 1532995200000,
    created_at: '2018-07-19 15:36:46',
    is_archived: 0,
    use_dynamic_assignment: null,
    logo_image_url: '',
    intro_html: null,
    primary_color: null } ]
TypeError: Field error: value is not an instance of Date
```
</details>

<details open>
<summary>Whereas Postgres returns:</summary>

```
anonymous {
  id: 79,
  organization_id: 1,
  title: '[bchrobot test] Custom Fields',
  description: 'There are custom fields',
  is_started: false,
  due_by: 2018-06-30T00:00:00.000Z,
  created_at: 2018-06-25T16:35:30.139Z,
  is_archived: false,
  use_dynamic_assignment: null,
  logo_image_url: '',
  intro_html: null,
  primary_color: null } ]
```
</details>

### DIFF
--- a/src/server/api/organization.js
+++ b/src/server/api/organization.js
@@ -1,3 +1,4 @@
+import moment from 'moment'
 import { mapFieldsToModel } from './lib/utils'
 import { r, Organization } from '../models'
 import { accessRequired } from './errors'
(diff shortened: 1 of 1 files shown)


schuyler1d:
I'm a little scared of `moment` -- we had a lot of bugs in some places where we had to remove it because it was very unpredictable.

Rethink-knex-adapter solved this here:
https://github.com/MoveOnOrg/rethink-knex-adapter/blob/master/models.js#L256
(and here: https://github.com/MoveOnOrg/rethink-knex-adapter/blob/master/query.js#L200-L209)

I suppose we're not getting this because we're using knex-straight up here (which is fine).  Would this also work if we just converted it with `new Date(...)` like in rethink-knex-adapter? rather than using moment?

PR #732, opened 2018-07-20
Allow users to, at upload time, exclude contacts from existing campaigns

bchrobot:
At upload time, allow users to exclude contacts from existing campaigns. This exclusion is done server-side using the same mechanism as for filtering opt-outs.

screen shot 2018-07-20 at 3 58 53 pm

### DIFF
--- a/src/api/campaign.js
+++ b/src/api/campaign.js
@@ -49,5 +49,6 @@ export const schema = `
     textingHoursStart: Int
     textingHoursEnd: Int
     timezone: String
+    createdAt: Date
   }
 `
--- a/src/api/schema.js
+++ b/src/api/schema.js
@@ -87,6 +87,7 @@ const rootSchema = `
     introHtml: String
     useDynamicAssignment: Boolean
     contacts: [CampaignContactInput]
+    excludeCampaignIds: [Int]
     contactSql: String
     organizationId: String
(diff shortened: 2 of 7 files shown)


harpojaeger:
Trivial note: the language on here ("exclude contacts from existing Spoke campaigns") is ambiguous. It could mean "ignore and do not upload contacts that are already in other Spoke campaigns" or "remove uploaded contacts from any existing Spoke campaigns they are already in". I would suggest something like "You can also filter out contacts from this upload that are already in an existing Spoke campaign."

It would also be worth making clear to the user whether this is any contact that has _ever_ been uploaded, any contact in an active (non-archived) campaign, or any contact in an active (non-archived) campaign that has not yet been texted.

schuyler1d:
Thanks!  This looks like it will do what you want it to do (so code-review is approved)

I'm wondering if it would be simpler to have the feature just be a checkbox for excluding non-archived non-overdue campaigns, rather than choosing them.  The plus here is the customizability, the minus is that if that's what I 'really want' then I have to hunt down and make sure i have all the active campaigns.

Thoughts?

schuyler1d:
useful use-case, but if we did this, it would need reimplementation within contact-loaders framework anyway.

PR #735, opened 2018-07-23
Fix for End to End tests

cp4r3z:
## Summary
This fix sets the admin assignments to zero before assigning all contacts to the texter.

## Root Cause
The logic for the inputs for assigning contacts to texters changed slightly. Now you aren't allowed to over-assign texters beyond the total number of contacts. Previously, if you did so the contacts would be unassigned from other texters.

## Run
`npm run test-e2e`


shakalee14:
Merged with #738

Issue #614, opened 2018-06-04
End to End Test: Basic Text Manager Workflow


shakalee14:
1. Login as an admin user
2. Create a new campaign in an existing organization
3. Create a csv with columns firstName, lastName, cell and add a few sample people to text (or query sql db)
with optOuts
without optOut
4. Assign a test texter, all the contacts
5. Create a script with a question
6. Define a few answers, and test saving with no answers, blank answers, and good answers.
7. Link at least one question to a next step
8. Add a canned response
9. Click "Start this campaign!"

PR #738, opened 2018-07-24
Modified e2e basic texter

shakalee14:
#735 #614 #734


schuyler1d (approved):
Didn't test if it runs but I like the doc!

PR #697, opened 2018-07-10
End to End Tests: Edit User, Edit/Copy Campaign

cp4r3z:
## Summary

Addresses the following Project tasks:

#612 Add Documentation for running Selenium Tests Locally
#613 Add Documentation for adding Selenium Tests
#615 End to End Test: Texter workflow
#616 End to End Test: Basic text manager workflow, for an organization with no existing texters
#617 End to End Test: Editing names and email as an admin or a texter
#619 End to End Test: Editing Campaigns
#620 End to End Test: Copying Campaigns

## Developer notes

To run tests:

Check node version
`npm run test-e2e`

PR #740, opened 2018-07-25
Stage main

shakalee14:
#697 #722 #700 #710


shakalee14:
This pr also includes : https://github.com/MoveOnOrg/Spoke/pull/740/commits/764f90edd28ec988728a18e3cd564100b320bb9b

Which is an update to fix the following jest error:
`SecurityError: localStorage is not available for opaque origins`

As suggested by here:
https://stackoverflow.com/questions/51554366/npm-test-fails-with-jest

schuyler1d (approved):
tests passing!!?!!

PR #743, opened 2018-07-26
[react-v16] Phase 2-b: webpack + react-hot-loader

bchrobot:
`react-hot-loader` upgrade is a mix of instructions from:

- https://github.com/gaearon/react-hot-loader#getting-started
- http://gaearon.github.io/react-hot-loader/getstarted/

Ex. the use of the `hot` HOC is listed as preferred in the repo's readme but not mentioned at all in the official documentation.


ibrand:
Closing after consulting with @bchrobot. This PR never got fully finished but has a lot of useful work in it so it's now referenced in: https://github.com/MoveOnOrg/Spoke/projects/10

PR #744, opened 2018-07-26
[react-v16] Phase 2-a: material-ui

bchrobot:
v1.2.0 adds longer term support, better layout support, and more functionality to `Table` components, which would be helpful for MoveOnOrg/Spoke#579 and addresses MoveOnOrg/Spoke#480.

There are still a few ToDos and lots of manual regression testing (global search for `TODO: material-ui`). The main things being:
- [x] Functionality for `AutoComplete` has been removed in favor of one of three 3rd party choices (98f6484)
- [x] `DatePicker` does not have a working implementation in v1.2.0 (40604ae)
- [x] Replace `onTouchTap` with `onClick` and drop the now-deprecated `react-tap-event-plugin`(b837243)
- [x] Remove `react-addons-css-transition-group` (it's not being used anywhere) (70113cd)
- [x] Update Toggle --> Switch (401da20)
- [x] Update DropDownMenu --> List + Menu (d294314)
- [ ] Confirm that the [`material-ui-color-picker` package](https://www.npmjs.com/package/material-ui-color-picker) works with v1.2.0


ibrand:
@bchrobot That rocks!!

PR #746, opened 2018-07-26
[react-v16] Phase 2-d: react-apollo

bchrobot:
Update `react-apollo` from v0.3 to v2.1. Inspired by MoveOnOrg/Spoke#418.

* v2.x [drops `redux`](https://www.apollographql.com/docs/react/recipes/2.0-migration.html#redux) for its store


ibrand:
Closing after consulting with @bchrobot. This PR never got fully finished but has a lot of useful work in it so it's now referenced in: https://github.com/MoveOnOrg/Spoke/projects/10

PR #755, opened 2018-08-06
Stage main

shakalee14:
#710 #700 
and bug fixes:
#758 #760 #757


shakalee14:
i can successfully use this branch to deploy to heroku, update a current heroku instance, deploy to lambda and run locally

PR #770, opened 2018-08-10
Added tools for developing in Docker

amancevice:
- Add `Dockerfile` to create Docker image with Spoke installed
- Add `docker-compose` configuration to test locally with a PostgreSQL container
- Add README section for getting started

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,10 +13,7 @@ RUN wget https://github.com/MoveOnOrg/Spoke/archive/v$SPOKE_VERSION.tar.gz && \
     npm install && \
     npm install -g foreman
 
-# Build Spoke
-WORKDIR /Spoke
-RUN npm run prod-build
-
 # Spoke Runtime
+WORKDIR /Spoke
 EXPOSE 3000
 CMD ["nf", "start", "--procfile", "./dev-tools/Procfile.dev"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM node:6.10
+
+# Install Spoke
+ARG SPOKE_VERSION=1.4.1
+ENV OUTPUT_DIR=/Spoke/build \
+    SPOKE_VERSION=$SPOKE_VERSION \
+    SUDO_USER=root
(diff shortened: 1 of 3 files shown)


shakalee14 (approved):

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    npm install && \
+    npm install -g foreman
+
+# Build Spoke
+WORKDIR /Spoke
+RUN npm run prod-build
+
+# Spoke Runtime
+EXPOSE 3000
+CMD ["nf", "start", "--procfile", "./dev-tools/Procfile.dev"]
```
schuyler1d: It's not clear to me if this docker file is to run production or dev. Here it is dev, on 18 it is prod, and the download is the release (so it can't be used for dev on main branch or someone's local clone)

amancevice: It's entirely possible that I've done something silly here. I'm not entirely certain I can explain the differences between "running in prod" vs "running in dev" in Spoke. I got my start on v1.4 of Spoke, which I've come to understand had some bugs relating to deployment so I might have been struggling against those.

I'll try to address the choices I made 1 by 1:

**Line 8: Installing from release**
Installing from a release -- as opposed to using the `COPY` command -- ensures that each build is repeatable and free of local state (`node_modules` dir, `.env` files, etc). A `.dockerignore` file could also be used to prevent cruft from getting into the image, but would not necessarily ensure the build is repeatable.

**Line 18: Running prod-build**
The `RUN npm run prod-build` stage, to my understanding, creates the artifacts under the `./build` directory. I did this so that those files would be available to a user should he/she choose to start the app in prod mode. Again, not sure I fully understand the difference, but that was the reasoning.

**Running on local clone**
You can do dev work on this image by mounting your local clone to `/Spoke`. You can do this from the `docker` CLI or changing `- ./.env:/Spoke/.env` to `- ./:/Spoke` in the `volumes` section of the `app` service in `docker-compose.yml`.

Ex:

```bash
docker run --rm -v $PWD:/Spoke spoke npm run some-local-task
```

amancevice: I removed the `npm prod-build` step from the Dockerfile.

PR #772, opened 2018-08-12
Possible usability improvement for texter reassignment

lperson:
Now using SuperSelectField to choose a texter to reassign to.

PR #754, opened 2018-08-06
First iteration of Message Review display optimizations for small screens

lperson:

PR #773, opened 2018-08-12
add campaign id to campaign filter (WFP request)

lperson:


joemcl:
Thanks @lperson . The idea of this not just to add Campaign ID in the Campaign filter, but to have that as the value that appears in the Campaign column in message search results, to save real estate on smaller screens, rather than full campaign name, which we can see when filtering by Campaign.

lperson:
@joemcl let's discuss after you merge #772  and #754.  Maybe it won't be necessary.

joemcl:
OK cool @lperson yeah let's see how #772 and #754 work in combo.

PR #774, opened 2018-08-12
enabled inline source map -- one line change

lperson:
@shakalee14 @schuyler1d @bchrobot Any thoughts on this change?


lperson:
@shakalee14 this is ready to merge into main

Issue #796, opened 2018-08-20
Texter Role Information Not Expiring in Redis


shakalee14:
- User role information not expiring in Redis cache which causes problems if role information has updated.

  - Can reproduce this issue by going to redis-cli
  - Running `KEYS *texter*`
  - Grabbing a `texterinfo-{id}` key
  - Running `ttl texterinfo-{id}` - this will return `-1` which according to [Redis](https://redis.io/commands/ttl) means `the key exists but has no associated expire.`

- This can be handled in different ways
  1) Figure out how to expire it OR 
  2) Delete texterinfo-{id} on logout so role is updated when user logs out and logs back in

shakalee14:
No expiration set here:
https://github.com/MoveOnOrg/Spoke/blob/63a4f66bd069ae4251a3ffe7789d98a78f74ec6c/src/server/models/cacheable-queries.js#L20

Noting - not possible after reading docs

PR #804, opened 2018-08-22
Delete redis userKey on logout

azuzunaga:
Co-Authored-By: Shaka Lee <shakalee14@users.noreply.github.com>

Addresses #796


shakalee14 (approved):

shakalee14 (changes requested):
Need to figure out another way to update roles.

schuyler1d (dismissed):
Before merging, can you:
1. rename logoutUser to something like clearUserCache
2. call the same function (clearUserCache) in [editUser](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/schema.js#L379)

schuyler1d:
I agree with @shakalee14 's implied hunch that the login-logout-login is actually related to https://github.com/MoveOnOrg/Spoke/pull/722

shakalee14:
I'm wondering if it should expire AND have some form of manual updating. 

@schuyler1d  thoughts?

shakalee14:
I think we can't actually expire it based on this:
https://stackoverflow.com/questions/16545321/how-to-expire-the-hset-child-key-in-redis

Also when attempting to do this, 
```
      await r.redis.multi()
        .hsetAsync(userKey, orgId, highestRole)
        .expire(userKey, 86400)
        .exec()
```
I get an error that it's not a function. I think @schuyler1d 's suggestions are fine.

schuyler1d:
I think this will be folded into the user caching refactor work: https://github.com/MoveOnOrg/Spoke/issues/894
-- but yes, we should have a `clear()` method which is triggered on logout or any user change.

Depending on where @azuzunaga is on this, we might want to close this PR and wait for the refactor one.

PR #805, opened 2018-08-23
Initialize env vars in test and development

### LINKED ISSUE #753: passport-auth0 sometimes breaks the build
Sometimes in dev, build fails with `backend error Error: You must provide the domain configuration value to use passport-auth0.` I think this is because `passport-auth0` gets imported before `process.env` is populated.

A global search of the entire source code shows that `dotenv` isn't being imported anywhere except in `/dev-tools/babel-run-with-env` and `/dev-tools/jest.transform.js`, neither of which is used in the normal `npm run dev` scenario. Unless there's env magic happening somewhere else I'm not aware of, this seems like the problem.

I propose adding this to the very top of `/src/server/index.js`:
```
if (['development', 'test'].includes(process.env.NODE_ENV)) {
  // eslint-disable-next-line global-require
  require('dotenv').config() // configure env vars in dev and test environments
}
```
The problem hasn't recurred since I tried this locally.

harpojaeger:
Closes #753


shakalee14 (dismissed):

shakalee14:
hey @harpojaeger - when attempting to include this branch in a claudia/lambda deploy, I get the following errors - which then breaks the deployment:

```
Error: Cannot find module 'dotenv'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/private/var/folders/sh/lljyq6xx2ddgd40snm01f57m0000gn/T/hvvGDl/spoke-1.4.1-E5lckB/package/webpack/config.js:1:63)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
```

package.json
```diff
@@ -86,7 +86,6 @@
     "chart.js": "^1.1.1",
     "cookie-session": "^2.0.0-alpha.1",
     "dataloader": "^1.2.0",
-    "dotenv": "^2.0.0",
```
jlegrone, line 88: Moving from dependencies to devdependencies can lead to runtime errors if there is potential for usage directly in application code. Using [dotenv with the -r flag](https://www.npmjs.com/package/dotenv#preload) works well in some cases, but here I suggest just keeping dotenv in dependencies since we're using nodemon.

src/server/index.js
```diff
@@ -1,3 +1,8 @@
+// configure env vars in dev and test environments
+if (['development', 'test'].includes(process.env.NODE_ENV)) {
```
jlegrone, line 2: `if (process.env.NODE_ENV !== 'production') {...}`?

package.json
harpojaeger: I haven't looked carefull yet, but AFAICT, dotenv shouldn't ever be used in production. I'd like to figure out why that's happening before deciding to leave it in dependencies.

jlegrone: Dotenv will never override already set environment variables ([docs](https://www.npmjs.com/package/dotenv#what-happens-to-environment-variables-that-were-already-set)) so in general I think it's safe but not preferable to use in production.

If we add code paths that could be hit in deployed applications then we should also make dotenv a production dependency IMO, though this is also avoidable if we use a wrapper script for `src/server/index.js` in dev or the `node --require` flag.

harpojaeger: Ooooh, I didn't realize it had that safeguard. Good to know that it at least won't wipe out already set vars. I still want to figure out where the actual problem is happening as it is (shouldn't be hard, I just haven't worked on it), but I'll keep that in mind.

src/server/index.js
harpojaeger: I generally prefer to specify the positive case, for the sake of making things overall more deterministic, but I'll come back to this after taking a closer look at the existing problems with this PR.

package.json
jlegrone: I suspect that `NODE_ENV` was set to `development` in @shakalee14's deployment, and because claudia only installs production dependencies it caused a runtime exception.

harpojaeger: Sounds highly plausible!

harpojaeger:
Okay, I've gone back over this PR to make sure I knew about all the places `dotenv` is required. Two questions:

- What is `dev-tools/babel-run-with-env.js` used for?
- What is `dev-tools/jest.transform.js` used for? Is L6 of that file supposed to say `module.exports` as opposed to `model.exports`?

These two files both use dotenv, and I'd like to understand why, to see if I need to make changes there too. Does anyone actually use these scripts?

The other loose end was that our webpack config.js required `dotenv` without checking whether we were in prod or not. I've put the same conditional logic in there as we have in `server/index.js`. Doing this in a webpack config struck me as a bit dodgy at first, but when I looked closer, I saw that we're doing good things with our envs in Webpack (inlining `NODE_ENV` and one other, which is also good for production bundle size).

The reason I'm sticking with moving dotenv from `dependencies` to `devDependencies` is:
1. It makes the production bundle smaller. Not by much, but I still think it's a good principle, and
2. Doing `require('dotenv').config();` in production could give users the idea that it's okay to keep a `.env` file in version control. _We definitely do not want anyone to do that._ It's a serious security risk.

I'm open to counterpoints on this, though.

@shakalee14 &/or @schuyler1d, I'm also curious if you can confirm @jlegrone's suspicion that the Claudia error was due to Claudia only installing production dependencies, but `NODE_ENV` being set to `test`.

harpojaeger:
Bump! This is a small change and eliminates a headache for new developers. What can I do to help this get merged?

schuyler1d:
`dotenv` should stay in babel startup, so it doesn't pollute the app.

PR #816, opened 2018-08-28
799 admin campaign list loading

shakalee14:
For #799 - changes initial page load for campaign list to be 25 and then gives options to reveal 10,25,50 and all.


schuyler1d:
Some notes/thoughts.
* Not sure we have this in our graphQL/Apollo version, but fwiw, there is a pagination model built in that we can/should use if it IS available: https://www.apollographql.com/docs/react/features/pagination.html (but only if it's not too hard)
* Assuming we're going to keep this, I would suggest that if we do pagination that we do this with offset/limit args -- i.e. both -- not just limit.  We might want to use those args, assuming we don't do fetchMore with whatever its args
* rather than adding it to the end of all those branches, maybe just have at the very end `if (listSize) {query = query.limit(listSize)}`

shakalee14:
@schuyler1d so is the suggestion to use `fetchMore` and implement pagination on this page - which then would need offset?

shakalee14:
@schuyler1d doing some research, it looks like we're running a version that has the initial implementation of `fetchMore` but people have found some issues while using it - reading this [thread on an issue created in apollo-client](https://github.com/apollographql/apollo-client/issues/2499), a known bug is that the "updateQuery method passed to ObservableQuery.fetchMore is receiving the original query variables instead of the new variables that it used to fetch more data". I'm not sure if that will be a problem for us.

schuyler1d:
so I'm ++ to skipping the whole fetchMore thing -- just wanted to bring it up.

src/server/api/campaign.js
```diff
@@ -3,6 +3,8 @@ import { Campaign, JobRequest, r } from '../models'
 
 export function buildCampaignQuery(queryParam, organizationId, campaignsFilter, addFromClause = true) {
   let query = queryParam
+  const resultSize = (campaignsFilter.listSize ? campaignsFilter.listSize : 0)
+  const pageSize = (campaignsFilter.pageSize ? campaignsFilter.pageSize : 0)
```
schuyler1d, line 7: let's default pagesize to something other than 0 -- maybe 10? 
If `pageSize` is `offset' -- then maybe rename to `pageOffset` -- 'pagesize' sounds like it would mean the same thing as listsize/resultsize

src/server/api/campaign.js
```diff
@@ -17,6 +19,12 @@ export function buildCampaignQuery(queryParam, organizationId, campaignsFilter,
     if ('campaignId' in campaignsFilter) {
       query = query.where('campaign.id', parseInt(campaignsFilter.campaignId, 10))
     }
+    if (resultSize && !pageSize) {
```
schuyler1d, line 22: let's have cleaner conditionals
```
if (resultSize) {
   query = query.limit(resultSize)
}
if (pageSize) {
   query = query.offset(pageSize)
}

schuyler1d (approved):
style nits.  no more approval necessary

PR #820, opened 2018-08-28
786 reverse order past messages

### LINKED ISSUE #786: Please reverse order of Past Messages
As it is now, a texter who needs to correct a user error has to scroll through all Past Messages to find the one that needs correction. If scrolling were ordered most recent first, this would be easier in most cases when the Quality Control Team is able to give feedback to users in a timely manner. 
@codygordon @lperson

azuzunaga:
Resolves #786


shakalee14:
@azuzunaga - can you pull in the latest main?

shakalee14 (approved):
Tested locally & compared results in UI to updated query results in DB.

Issue #779, opened 2018-08-13
Sometimes Assignment Creation in Admin Menu gets Stuck


shakalee14:
- Usually happens on campaigns where assignments are with 10 or more texters
- Current work around is deleting the job request in the db. After deletion, the texter assignments still remain and the Admin can save. Sometimes this has to happen several times & other times, this "resolves itself".

Frydafly:
checking if this is still an issue? if it is, would be helpful to see screenshots or instructions for reproducing the bug! @shakalee14

schuyler1d:
I believe this was resolved, and also largely mitigated with dynamic assignment altogether.  This might also have related somewhat to this issue https://github.com/MoveOnOrg/Spoke/pull/1805

PR #831, opened 2018-08-30
allow admins to kill a job if it has stalled, e.g.

schuyler1d:
This is one approach to solving #779 -- if jobs get stuck, then admins should have the ability to delete the job.

Obviously it would be better if jobs never got stuck....
The plus side of this is that it also provides an interface for displaying job error/success messages in the future.

### DIFF
--- a/src/containers/AdminCampaignEdit.jsx
+++ b/src/containers/AdminCampaignEdit.jsx
@@ -102,7 +102,7 @@ class AdminCampaignEdit extends React.Component {
     // NOTE: Since this does not _deep_ copy the values the
     // expandedKey pointers will remain the same object as before
     // so setState passes on those subsections should1 not refresh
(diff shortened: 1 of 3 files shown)


azuzunaga (approved):

src/containers/AdminCampaignEdit.jsx
```diff
@@ -1,11 +1,14 @@
 import PropTypes from 'prop-types'
 import React from 'react'
+
 import WarningIcon from 'material-ui/svg-icons/alert/warning'
 import DoneIcon from 'material-ui/svg-icons/action/done'
+import CancelIcon from 'material-ui/svg-icons/navigation/cancel'
```
azuzunaga, line 6: Sad we didn't stick with 🚆

PR #832, opened 2018-08-30
Add placeholders for createOptOut method (scaling proposal)

azuzunaga:
* Should it be in contact redis record or overall optouts (or both) 
* Should it be keyed off of cell or contact id


schuyler1d:
closing -- the scaling work was finally merged (though pulled back a bit from the original assignment structure)

PR #835, opened 2018-08-31
670 campaign editor presence

### LINKED ISSUE #670: Spoke admins can overwrite each other if performing admin work in same Spoke campaign
Currently, an admin must refresh their browser to ensure the most recent information (texters assigned; number fo texts sent) are up to date in a Spoke campaign.

In a situation where two admins are simultaneously accessing the same campaign, they will not see any changes made until clicking save, meaning changes made in the same campaign by another admin will not be visible.

A more ideal setup would show any updates to a Spoke campaign in real time as opposed to an admin having to refresh their browser to see if anything has changed in a campaign they are administrating.

azuzunaga:
Resolves #670:

This is what it looks like with editors:
image

And without editors:
image

(Similar for campaigns that haven't started yet)

Let me know if I should change the styling or if this is ok.


mathemagica (dismissed):
👍

shakalee14 (approved):

shakalee14:
Merged with #864

PR #836, opened 2018-08-31
Revert airbnb linter upgrade and package-lock.json tracking

azuzunaga:
Remove package-lock.json and untrack it, downgrade `eslint-config-airbnb` back to `^9.0.1`

Checked by running `npm run lint`.


schuyler1d (approved):
:+1:

PR #838, opened 2018-09-01
texting hours per campaign

lperson:


lperson:
closes #644

shakalee14:
Awesome! @lperson - I'm having trouble getting this working locally / testing locally - let me know if these are the correct steps.

1) Run `npm run dev` - which logs that a migration has occurred on the campaign table
2) Create a campaign and add `texting hours`
screen shot 2018-09-03 at 12 29 53 pm
3) Go to organizational settings and add a different texting hours constraint
screen shot 2018-09-03 at 12 29 41 pm
4) Switch to texter - and it should respect texting hours set on the campaign - currently it's still greyed out: 
screen shot 2018-09-03 at 12 30 00 pm



Currently in the db the texting hours are set to what I set in the campaign and override_organization_texting_hours is set to true - but I'm not able to send texts.

lperson:
@shakalee14 try refreshing the switch to texter view.

lperson:
@shakalee14 what time is the machine set to, and in what timezone?

lperson:
@shakalee14 also, please confirm you have the correct branch -- `new_texting_hours_per_campaign` **NOT** ~~texting_hours_per_campaign~~

shakalee14:
Refreshed a few times with no luck - and my machine? It's in EST.

lperson:
@shakalee14 also, do the contacts have zip codes?

shakalee14:
I have the correct branch and I'm using this csv:
https://github.com/MoveOnOrg/Spoke/blob/main/__test__/test_data/female_scientists.csv

lperson:
@shakalee14  the zip codes are in California.  Spoke is coded to respect the timezone of the zip code.  You set texting to start at noon, but it's about 9:48 in California, so none of those zipcodes are in a timezone where it's before noon.  The timezone in campaign texting hours config is used with contacts that have no zip code and to determine whether it's DST or not.  Try setting campaign texting hours to something before 9am.

src/components/CampaignTextingHoursForm.jsx
```diff
@@ -0,0 +1,234 @@
+import type from 'prop-types'
+import Toggle from 'material-ui/Toggle'
+import React from 'react'
+import Form from 'react-formal'
+import GSForm from './forms/GSForm'
+import CampaignFormSectionHeading from './CampaignFormSectionHeading'
+import yup from 'yup'
+import { css, StyleSheet } from 'aphrodite'
```
shakalee14, line 8: Can we removed this unused import.

src/components/CampaignTextingHoursForm.jsx
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  addAutocompleteFormField(
+    name,
+    stateName,
+    initialValue,
+    label,
+    hint,
+    choices,
+    collection
```
shakalee14, line 60: what is this used for? `collection` ?

lperson: `choices` contains the strings to be displayed

`collection` is the original values, used by a callback created by this function, so when a choice is selected, we can find the value to include when we fire the selected event

src/components/CampaignTextingHoursForm.jsx
lperson: Sure!  ~~... but I think they're all used.~~

lperson: Done.

schuyler1d:
For future, I have resolved conflicts with this PR and `scaling-tip` in branch `new_texting_hours_per_campaign-withcaching`

shakalee14:
Merged with #901  and #900

PR #848, opened 2018-09-03
Add org check to cache

shakalee14:
Includes #722 and caches org info if redis is enabled.


src/server/models/cacheable_queries/user.js
```diff
@@ -32,6 +32,55 @@ export async function userHasRole(userId, orgId, acceptableRoles) {
   }
 }
 
+async function getUserOrgInfo(userId, orgId) {
```
schuyler1d, line 35: I think this does two separate things
1. tests membership
2. gets org name and caches it.

For (1) it is already cached in `texterinfo-<userid>` with key '<orgId>' so we shouldn't recache it again -- and just use that.
For (2), we don't need to save that per-user, because it's at the org level.  and that should be able to use the general org cache in this PR: https://github.com/MoveOnOrg/Spoke/pull/855/files#diff-a384a0e3111ace4e3ecc640befd115e4R11

I know in some cases we just want the org name for the user query -- in that case, maybe worth stuffing it into the already-existing `texterinfo-<userid>` and maybe split role:orgName or something.

src/server/models/cacheable_queries/user.js
```diff
@@ -32,6 +32,55 @@ export async function userHasRole(userId, orgId, acceptableRoles) {
...
+      .distinct()
+    await r.redis.set(userOrgKey, JSON.stringify(orgData[0]))
+    return orgData[0]
+  } else {
+    let parsedOrgInfo = JSON.parse(orgInfo)
+    return parsedOrgInfo
+  }
+}
+
+export async function userOrgsWithRole(role, userId) {
```
schuyler1d, line 55: can we rename this `userOrgs(userId, role)`, and have `role` as the second argument which will be optional (since I believe it is)

schuyler1d (changes requested):
Let's use the cache objects we already have rather than spread ourselves too much.

shakalee14:
Closing this PR for now to try another fix to my initial issue.

Issue #813, opened 2018-08-27
Decouple contact info retrieval (client-side)


schuyler1d:
There are a few things that need to be done for our [scaling plan](https://github.com/MoveOnOrg/Spoke/blob/main/docs/HOWTO-scale-spoke-plan.md#client-side-optimization) on the client-side:
* decoupling getContact in container AssignmentTexterContact.js and moving it to TexterTodo.jsx and be able to load contacts separately from a particular contact screen
* prepare for a point where logic can determine whether the client should try to send contacts that have been loaded from the 'last' pull with contacts that were still in a previous status but might have been purposefully skipped, etc.

WIP/experiment was done in https://github.com/MoveOnOrg/Spoke/pull/783

schuyler1d:
I can think of a few ways to implement this and haven't completely decided which one is best.  Here are The options as I see them:

## 1 `getContacts([id array])` query
Here the initial query in TexterTodo (still) gets ALL the contact ids, and then we add a server call that just makes it easier to query a batch of ids at once.

Pros: simplicity/similarity to current model. 
Cons: dynamic assignment is still a bit of a side channel dance.

## 2 mutation `getNextContacts($contactFilter, $offset)`
Here instead of calling getContact or getting all the ids at all on the first TexterTodo query, we just get counts, etc.  Then a new mutation would sort of combine getContact(s) with dynamic assignment findNewCampaignContact(s) and this will return the 'next' batch of contacts.

Pros: combines the dynamic and non-dynamic code paths for possibly a simpler caching story.  The cache story would be that it would load the contacts in a mutable queue and just give the next 10.

Cons: There's some complexity in what `$offset` would be -- it could be the 'first contact id' or something similar to a sql OFFSET call -- but as we're changing the status as we do it, it wouldn't directly be a real sql offset -- we'd need some state/memory.  One possibility is if this is a mutation, then maybe we just don't need an offset -- and it's just 'next group' -- though this becomes more complex for ordering/skipping -- esp. when navigating replies.

## 3 query `getNextContacts($contactFilter, $offset)`
This could be similar to (2) but we wouldn't mutate anything -- then the offset would be more 'reliable' but we'd have to figure out how to track the offsets.

## Conclusions

[...]

PR #850, opened 2018-09-03
decouple contact getting

schuyler1d:
I believe this should complete #813

In passing, we also fix the root cause of #452 because we need contacts and index in the TexterTodo/AssignmentTexter contexts to be reliable and not change every contact update.

This PR does three things:
*    Move the api call to get contact data up to TexterTodo (and executed by AssignmentTexter) away from AssignmentTexterContact
*    Instead of just gettting one contact at a time, we get 10 contacts at a time and try to load new contacts 5 contacts ahead of when we will run out of cached local data
*    Before on each contact shift we refreshData'd all the assignment info, and now we will update that every 20 seconds.

Note: This is 'reopening' the PR #849 where I accidentally pushed a change I didn't mean to


shakalee14:
After deploying to stage, this PR removes ability for texter to view text contact screen.

shakalee14 (dismissed):

PR #852, opened 2018-09-03
Change button from "replies" to "responses"

JesseWeinstein:
Per suggestion by Dinah -- "replies" are from contacts to texters, "responses" are sent by texters to contacts.

### DIFF
--- a/__test__/AssignmentSummary.test.js
+++ b/__test__/AssignmentSummary.test.js
@@ -189,7 +189,7 @@ describe('contacts filters', () => {
     expect(sendFirstTexts.contactsFilter).toBe('text')
 
     const sendReplies = mockRender.mock.calls[1][0]
-    expect(sendReplies.title).toBe('Send replies')
+    expect(sendReplies.title).toBe('Send responses')
     expect(sendReplies.contactsFilter).toBe('reply')
 
     const sendLater = mockRender.mock.calls[2][0]
--- a/src/components/AssignmentSummary.jsx
+++ b/src/components/AssignmentSummary.jsx
@@ -119,7 +119,7 @@ export class AssignmentSummary extends Component {
             })}
             {(window.NOT_IN_USA && window.ALLOW_SEND_ALL) ? '' : this.renderBadgedButton({
               assignment,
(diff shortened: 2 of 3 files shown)


DinahSanders:
Thanks, Jesse!

Or, as I have it in our internal docs:

> Reply
> A text sent by a contact to a texter; that is, what they say to us. Remember: They _reply_ to our initial text; we _respond_ to their replies.
> 
> Response
> A text sent by a texter to a contact; that is, what we say to them after the initial message. Remember: They _reply_ to our initial text; we _respond_ to their replies.


I recommend these changes specifically:
- change "Send Replies" button on home screen to "Send Responses"
- change "Skip Reply" button on sending screen to "Skip Response"

DinahSanders:
@JesseWeinstein, looks like maybe this missed changing the menu title "Canned Replies" to "Canned Responses"?

This seems to only be the case on some devices (e.g. iPhone).

JesseWeinstein:
Yep, I didn't know to change that. Will get to it when I get a chance. Still unclear when this will get merged. :-)

DinahSanders:
No worries. Makes sense to me not to roll out this change until after midterm elections, so folks have a more relaxed moment in which to adjust training materials and get their texters adapted to the new wording.

schuyler1d:
going to close this -- weirdly, once again we're in another election cycle, so the language needs to stay consistent.  We will re-review language post-election after user testing.

DinahSanders:
Why throw out the work done on finding all the loose ends that need to get cleaned up? Why just cross our fingers that we'll find the problems again?

This feels like saying to a programmer, "well, we didn't roll it in this time, so we're going to delete your code and you can write it again later".

JesseWeinstein:
Er, isn't there a "PRs that had been forgotten for over 2 years" queue that this could get tagged with, rather than just closing it?

schuyler1d:
That's fair.  I'll re-open and we'll get this in post-election.

Frydafly:
Hey folks! Checking in on this language change. Is it part of a roadmap or still on hold? (cc: @schuyler1d )

PR #853, opened 2018-09-04
Prevent twilio from queuing messages past the end of texting hours

lperson:
This is branched from `new_texting_hours_per_campaign`

Intended to either fix or mitigate the problem described starting with [this slack message](https://progcode.slack.com/archives/C717R1DMZ/p1535981668000100).


src/server/api/schema.js
```diff
@@ -909,6 +923,28 @@ const rootMutations = {
...
+        { textingHoursEnd: organization.texting_hours_end, textingHoursEnforced: organization.texting_hours_enforced },
+        {
+          textingHoursEnd: campaign.texting_hours_end,
+          overrideOrganizationTextingHours: campaign.override_organization_texting_hours,
+          textingHoursEnforced: campaign.texting_hours_enforced,
+          timezone: campaign.timezone
+        }
+      )
+
+      const sendBeforeDate = sendBefore ? sendBefore.toDate() : null
```
schuyler1d, line 972: if we are doing sendBefore here, shouldn't we also test whether we're already past sendBefore before creating the message?

-- i.e. if the API call is made past sendBefore we should error, right?

lperson: 👍

PR #855, opened 2018-09-04
Scaling loaders

schuyler1d:
stubs for cacheable_queries files and gesturing to how we might do some things.

Note this is using base: scaling-wip

### DIFF
--- a/src/server/api/assignment.js
+++ b/src/server/api/assignment.js
@@ -95,7 +95,11 @@ export function getContacts(assignment, contactsFilter, organization, campaign, 
 export const resolvers = {
   Assignment: {
     ...mapFieldsToModel(['id', 'maxContacts'], Assignment),
-    texter: async (assignment, _, { loaders }) => loaders.user.load(assignment.user_id),
+    texter: async (assignment, _, { loaders }) => (
+      assignment.texter
+      ? assignment.texter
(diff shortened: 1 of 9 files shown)


shakalee14 (commented):
thanks for stubbing this out!

@schuyler1d is it true that we should probably stick to similar naming conventions because they're the same idea? Specifically, I believe `clear` and `load` in different caching models are the same as `clearQuery` and `query`.

So essentially each model will have a `clear`, `load` and `save`. 

Should there also be an `update`? which then uses `clear` and `save`? 

Also, another pattern to maybe establish is deciding when we should expire - an expire call should happen during the `load`, but that's another arch design decision.

schuyler1d:
@shakalee14 
This is my naming structure reasoning so far.  It might change, but we should document the style/meaning somewhere once we nail it down:

* `clear` and `load` should be reserved for id as only/first argument for purposes of being able to use it here:
https://github.com/MoveOnOrg/Spoke/pull/855/files#diff-72944ac6d65443d67b9ebbf9989a1aecR32
* `query` and `clearQuery` take the same argument object -- hopefully encompasses ALL uses/lookups -- this is for pairs e.g. user-org or cell-org, etc.

schuyler1d:
I'm going to close this since it's an intermediate result.

PR #857, opened 2018-09-05
canned responses, cached

schuyler1d:
need this to take on caching campaigns and getting a feel for the api


schuyler1d:
merged to main

PR #866, opened 2018-09-06
Put cell numbers through formatter

### LINKED ISSUE #863: Data warehouse imports should format cell numbers correctly.
Currently the load from warehouse job processing takes the cell without adjustment here:
https://github.com/MoveOnOrg/Spoke/blob/main/src/workers/jobs.js#L233

However, cell numbers from a db may not include, e.g. `+1` for american contexts -- and we need the full international number in the cell column, as-is, instead of adapting it.

When uploading a csv it's run through:
https://github.com/MoveOnOrg/Spoke/blob/main/src/lib/phone-format.js#L4
getFormattedPhoneNumber(cell, country='US') and maybe these should as well.

shakalee14:
Fixes #863 

Tested on staging.


mathemagica (approved):
👍

PR #867, opened 2018-09-06
first text from campaign

schuyler1d:
This is on top of #850
and removes extraneous api endpoints and then just supplies the first script from campaign's interaction Steps object rather than forcing a useless additional query per-contact.

Also adding some work to #813


schuyler1d:
this was merged into main

PR #877, opened 2018-09-08
Fix behavior for stray opted out contacts

lperson:
This fixes a bug WFP encountered today.

[Full discussion here onward.]( https://progcode.slack.com/archives/C717R1DMZ/p1536424068000100)

### DIFF
--- a/src/components/AssignmentTexter.jsx
+++ b/src/components/AssignmentTexter.jsx
@@ -69,9 +69,13 @@ class AssignmentTexter extends React.Component {
     return this.state.currentContactIndex < this.contactCount() - 1
   }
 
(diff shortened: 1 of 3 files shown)


schuyler1d:
This issue is real, but should be fixed and superceded by this change:
https://github.com/MoveOnOrg/Spoke/pull/850/files#diff-bd86771e517cac3b787b7ee14ea16881R47
which fixes the root issue -- basically stopping the incongruence of advancing the index at the same time as reloading the contact list after doing an action (sending or opting out) which will remove the member from the contact list, and thus the contact list is 'off by 1' from the index increment's expectation.

PR #880, opened 2018-09-10
force display on graphQL errors

mathemagica:
It was driving me nuts that graphQL syntax errors were getting silently eaten- this is particularly hard to debug. Looks like this is because the ApolloClient default error parser was assuming most errors to come in the form of parse-able JSON. When this isn't the case, instead of displaying generic error message on error parsing exception, still display the raw message as a string


schuyler1d (approved):
yaas! :+1:

Issue #741, opened 2018-07-26
Changing the OPT OUT text message for MoveOn


ErikKirkstein:
As we head into election season, I would like to more clearly ID to contacts that we are opting them out of  our texts but that, since we are using the voter file, other groups may still text them. Often, contacts don't distinguish between the groups texting them.

Here's my idea:

"I am opting you out of texts from MoveOn immediately. Have a great day."

joemcl:
@ErikKirkstein this is related - https://github.com/MoveOnOrg/Spoke/issues/632

schuyler1d:
I commented on architecture options here: https://github.com/MoveOnOrg/Spoke/issues/632#issuecomment-409370547

Code points can be seen from @joemcl 's PR:
https://github.com/WorkingFamilies/Spoke/pull/14/files

If you opt to set it per-organization, then you could do that here:
* client-side: This file https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/Settings.jsx#L171  (see above for interface, and that place for the mutation)
* server-side: we would probably add a mutation similar to texting-hours here: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/schema.js#L472-L482 (note accessRequired 'OWNER') in that method and then you would save it in a json blob features: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/models/organization.js#L10
  
* note we'll probably be caching org metadata including anything that can be customized by org, so this would then be cached during our scaling work.
If it's just an env variable, then we wouldn't need to cache it.

PR #882, opened 2018-09-10
741 default instance-wide optout message

mathemagica:
Added a new place to edit org-specific optout message in org settings, and if set, it'll override both the env var instance-wide setting OPT_OUT_MESSAGE and also the hard coded default  "I'm opting you out of texts immediately. Have a great day." Test this by going into your name (in top right) -> your org -> settings and then editing the "Default opt-out message" setting. #741

### DIFF
--- a/src/containers/Settings.jsx
+++ b/src/containers/Settings.jsx
@@ -124,23 +124,29 @@ class Settings extends React.Component {
           <CardHeader
             title='Settings'
           />
-          <GSForm
-            schema={formSchema}
-            onSubmit={this.props.mutations.updateOptOutMessage}
-            defaultValue={{ optOutMessage }}
-          >
+          <CardText>
+            <div className={css(styles.section)}>
 
-            <Form.Field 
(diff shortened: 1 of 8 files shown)


src/containers/Settings.jsx
```diff
@@ -114,12 +113,40 @@ class Settings extends React.Component {
...
+
+            <GSForm
+              schema={formSchema}
+              onSubmit={this.props.mutations.updateOptOutMessage}
+              defaultValue={{ optOutMessage }}
+            >
+
+              <Form.Field 
+                label='Default Opt-Out Message' 
+                name='optOutMessage'
```
shakalee14, line 137: can we add the prop `fullWidth` here so the user can see the full message (versus having to scroll within the form field) ?

shakalee14 (commented):
Thanks! Some changes requested - specifically adding a small improvement to the UI experience

shakalee14 (approved):

PR #869, opened 2018-09-06
Cached opt-outs

schuyler1d:
This implements cached opt-outs per-organization.

it includes a merge of #867 -- the best way to see the diff is just this commit:
https://github.com/MoveOnOrg/Spoke/commit/04a925a24352acecce233cd5e34408dbff8114e6

PR #889, opened 2018-09-10
caching organization/campaign

schuyler1d:
This PR has a number of stubs that include other PRs as well.  Operationally, this creates several caches, but only *uses* campaign and organization.  The rest will need some closer review before being ready.

Client side:
* #850 decoupling client-side contact-getting
* #867 simplify api to get first text from campaign

Server caching
* #857 canned responses
* #869 opt-outs, canned responses

Cache components and readiness
  - [x] campaign
  - [x] organization
  - [x] canned-responses (used in campaign/assignment lookup)

For future PR
  - [ ] campaign-contacts (needs to deal with message_status and when to load)
  - [ ] assignment (just a stub)


schuyler1d:
@shakalee14 CACHE_PREFIX might be set to something like `spokeprod` -- if multiple instance (e.g. stage/prod or just multiple instances, in general) use the same cache server, then it's useful to add a prefix so as not to collide with other 'tenants' of the redis server.

shakalee14:
@schuyler1d 

When attempting to create a campaign an upload the `female_scientists.csv`, I get the following error locally: `TypeError: Field error: value is not an instance of Date`

When I try to create a campaign on staging, I get the same error in the client, which then freezes the page on a loader.

src/server/models/cacheable_queries/opt-out.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+export const optOutCache = {
+  clearQuery: async ({cell, organizationId}) => {
+    // remove cache by organization
+    // (if no cell is present, then clear whole query of organization)
+    if (r.redis) {
+      if (cell) {
+        await r.redis.sdelAsync(orgCacheKey(organizationId), cell)
```
azuzunaga, line 30: Based of the [documentation for redis sets](https://redis.io/commands#set), use `SREM` to delete a member of a set.

PR #891, opened 2018-09-11
884 instance opt outs

### LINKED ISSUE #884: Feature: Environment variable for Opt-Outs to be shared across all organizations on the instance
Sometimes a single real organization will have a spoke instance, but want to divide camaigns and texters for management purposes.

The current blocker is that opt-outs are only per-organization, so an opt-out that occurs from one organization will not get opted-out on another campaign.

Thus we should have an environment variable
OPTOUTS_SHARE_ALL_ORGS
when true:
* Opt-out queries will not filter by organization
* creating an opt-out should mark all current contacts across organizations ([not just across campaigns](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/schema.js#L776), like now)

One possibility while working on this would be to remove/ignore the `campaign_contact.is_opted_out` column and change all queries to JOIN on `opt_out` by cell instead.

Also please be conscious of parallel scaling work that has begun:
https://github.com/MoveOnOrg/Spoke/pull/869

shakalee14:
Branched off of #889 
Fixes #884

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -158,10 +158,12 @@ export async function uploadContacts(job) {
     await CampaignContact.save(savePortion)
   }
 
+  const optOutsByOrgId = function() { this.select('cell').from('opt_out').where('organization_id', campaign.organization_id) }
+  const optOutsByInstance = function() { this.select('cell').from('opt_out') }
+  const getOptOutCount = (!!process.env.OPTOUTS_SHARE_ALL_ORGS ? optOutsByInstance : optOutsByOrgId)
+
   const optOutCellCount = await r.knex('campaign_contact')
-    .whereIn('cell', function optouts() {
-      this.select('cell').from('opt_out').where('organization_id', campaign.organization_id)
-    })
+    .whereIn('cell', getOptOutCount)
     .where('campaign_id', campaignId)
     .delete()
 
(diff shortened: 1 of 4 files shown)


src/workers/jobs.js
```diff
@@ -158,10 +158,12 @@ export async function uploadContacts(job) {
     await CampaignContact.save(savePortion)
   }
 
+  const optOutsByOrgId = function() { this.select('cell').from('opt_out').where('organization_id', campaign.organization_id) }
```
schuyler1d: since this is the same code, as below, can we abstract it out to a global file function or constant?

schuyler1d (changes requested):
So I have a suggestion to make this PR much simpler.
1. Keep what's in src/workers/jobs.js
2. Leave api/campaign-contact.js unchanged
3. Do the following in models/cacheable_queries/opt-out.js:

change 
```
const orgCacheKey = (orgId) => (
!!process.env.OPTOUTS_SHARE_ALL_ORGS
 ? `${process.env.CACHE_PREFIX|""}optouts`
 : `${process.env.CACHE_PREFIX|""}optouts-${orgId}`)
```
And then in loadMany just do
```
let dbQuery = r.knex('opt_out').select('cell')
if (!process.env.OPTOUTS_SHARE_ALL_ORGS) {
  dbQuery = dbQuery.where('organization_id', organizationId)
}
const dbResult = await dbQuery
```
Also keep your `accountingForOrgSharing` in the query method.

And then I think it will all work with just that! -- Basically now that we've abstracted out the opt-out query, we can just decide to ignore organizationId when we don't do something based on it.

Generally, try to avoid repeating yourself in code -- if you find yourself doing a big if ..... else ..... especially if the ....'s are similar, think about restructuring the code.

src/workers/jobs.js
```diff
@@ -14,6 +14,18 @@ import { Notifications, sendUserNotification } from '../server/notifications'
...
 
+function optOutsByOrgId(orgId) {
+  return r.knex.select('cell').from('opt_out').where('organization_id', orgId)
+}
+
+function optOutsByInstance() {
+  return r.knex.select('cell').from('opt_out')
+}
+
+function getOptOutCount(orgId) {
```
schuyler1d: can we rename this something like `getOptOutSubQuery` since it's not the count, but the full list.

schuyler1d (approved):
great! :+1: 
small nit in a function name change.  After that, consider it approved and ready to QA/merge.

Issue #828, opened 2018-08-30
Bug : Max Contacts on Dynamic Assignment Does Not Work


shakalee14:
When an admin attempts to set a max contacts value for a texter a few things happens:
- If the value is not 0, it is set in the DB
- If the value is 0, it sets in the db as an empty string
- The texter can still text above the max contacts value

sandramchung:
Code points: 
[dynamically assigns new contacts in dynamic assignment mode](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/schema.js#L701)
[assignment creation and updating](https://github.com/MoveOnOrg/Spoke/blob/main/src/workers/jobs.js#L364)

sandramchung:
Order of operations is: 
1) make zero mean zero for max contacts. 
2) ensure that max contacts always saves (I couldn't always get max = 0 to save on my local)

We also brought up the issue of simplifying this code, which is a mess.
There's a conversation to be had about whether and how to support switching between dynamic and standard assignment modes in a given campaign - that will inform future work on assignment functionality and probably the assignment UI.

shakalee14:
Conversation with Erik about switching between dynamic and standard assignments - here is when the switching happens:

1) Erik creates a campaign, it's dynamic and texters have no limits on the amount of contacts they can text (so max_contacts is null)
2) Texters start texting, but it's slow (the loading between contacts)
3) Erik switches back to non-dynamic and sets assignments (to fix the loading issue)
4) This potentially improves performance/load times (?)

joemcl:
adding a cross reference here to #1066 - think there's still some stuff to figure out on Dynamic, though I've figured out a solution to it not working for us recently.

joemcl:
I *think* this should be re-opened since it's not actually fixed... see #1066 re the workaround... cc: @ibrand

Frydafly:
@joemcl checking in if this is still an issue?

Issue #629, opened 2018-06-11
Admin cannot remove a texter from Spoke campaign using "dynamically assign contacts" feature


ErikKirkstein:
When a texter is added to a Spoke campaign with "dynamically assign contacts" enabled, and Admin cannot remove them from the campaign.

Normally, a Spoke admin can either set a texter's assignment at "0" OR delete the texter by selecting the trash can icon and click "SAVE" to remove a texter from a Spoke campaign

This process should be enable for Spoke campaigns set to "dynamically assign contacts"

schuyler1d:
Some notes/codepoints

Code points for the user getting more dynamic assignments:
* On the client-side dynamic assignments are pulled/received here: https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/TexterTodo.jsx#L135-L149
* The server side is here: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/schema.js#L704-L760

Code points for assignment:
* server side: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/schema.js#L176-L187
but note that a lot of our code might do 'something different' assumption-wise inside: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/schema.js#L155-L166 and the [following job](https://github.com/MoveOnOrg/Spoke/blob/main/src/workers/jobs.js#L364)
* client-side is here: https://github.com/MoveOnOrg/Spoke/blob/main/src/components/CampaignTextersForm.jsx#L122 (note that onChange method is a beast! -- and has been a source of bugs/subtle-edge-cases in the past)

* This code is very relevant to the scaling work so worth going deeper with familiarity-- and it's also worth noting that the API might need to change slightly because [we want to get assignments/contacts in batches](https://github.com/MoveOnOrg/Spoke/issues/813)

Besides the [max contacts logic](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/schema.js#L723-L736) there there are no real restrictions on the user.
Perhaps setting the max_contacts on the assignment in dynamic mode is buggy (and/or the interface isn't clear) -- maybe it doesn't accept a value of `0`, for example.

If that isn't acceptable or working at all anymore, then maybe some other condition could exist related to the user record that's tested -- e.g. a user blocklist status, etc.

shakalee14:
@sandramchung

The dynamic assignment feature allows you to assign texter(s) to a campaign without giving them a set contact list/count to work with.

[...]

PR #895, opened 2018-09-11
828 fix max contacts

sandramchung:
We can now set max contacts to 0 to stop texters in dynamic assignment mode from sending any more texts in that campaign. Before, we weren't distinguishing between 0 and null, so both meant no limit on contacts assigned to a given texter. This PR addresses #828 and the connected issue #629.

Note that we had to fix rethink-knex and bump it up a version, so ppl will have to update the node module to see changes working.


shakalee14 (approved):
Pulled this locally and tested - was able to successfully remove the ability for a texter from texting more contacts by setting MAX to 0. Was also able to set it to 10 afterwards to put them back in texting mode for the campaign.

PR #900, opened 2018-09-12
New texting hours per campaign withcaching

shakalee14:
Based on a merge of #889 and #838


shakalee14:
@lperson - this looks great - we're testing this on stage/production

@ErikKirkstein made a great point about copied campaigns - when someone copies a campaign with texting hours enforcement, he suggests the custom texting hours for the campaign should copy over too.

PR #905, opened 2018-09-12
Campaign page related fixes

shakalee14:
This PR fixes 2 things: 
1) A bug introduced by https://github.com/MoveOnOrg/Spoke/pull/816 which breaks the ability to view the message review panel because of this error : `"Cannot read property 'listSize' of undefined"`
2) A request to change the default page size to be 50 (vs 25) by @erikkirkstein


mathemagica (approved):
👍

azuzunaga (approved):

PR #912, opened 2018-09-13
tweak csv to have multiple timezones in the file 

schuyler1d:
i want to test having multiple timezones in a file, so changing the test data.  @sandramchung this could change your testing work, so asking for your review.  If you have a similar tweak in-progress, then feel free to close and I'll wait for yours.


sandramchung (approved):
Thanks for flagging. This shouldn't impact my assignment testing work.

PR #915, opened 2018-09-14
fix contactsFilter undefined case in getContacts

schuyler1d:


shakalee14 (approved):
So previously we assumed `contactsFilter` was there, and sometimes it's not, so we should add an explicit check for it.

shakalee14 (changes requested):
After qa'ing on stage I still see the same send 1 text, white screen, refresh behavior. Sent you graphql request data in slack.

Issue #894, opened 2018-09-11
User-org cache/refactor and org-check


schuyler1d:
Building off of #848 and #889 

 - [ ] Add structure to cacheable-queries/user.js to that in https://github.com/MoveOnOrg/Spoke/pull/889/files#diff-64b48d71e43bff45ee02580e057c1851 -- maybe most similar to [opt-out.js](https://github.com/MoveOnOrg/Spoke/pull/889/files#diff-0315c1cb3427dd9041464225afbee43e)
 - [ ] Test org membership using [`texterinfo-${userId}`](https://github.com/MoveOnOrg/Spoke/pull/889/files#diff-bfea22d08d28486e25f685ae0b4a2ba0L8)
 - [ ] use the new structure in api/error.js and elsewhere.
 - [ ] When user or user-membership is modified use the/a new `clearQuery` or `clear(userId)` to clear the cache in all the places.

schuyler1d:
This was merged recently!

PR #916, opened 2018-09-14
WIP - refactor user.js in cacheable queries

azuzunaga:
@schuyler1d this is a WIP PR to make sure I am on the right track. Looking forward to your feedback.

References issue #894


src/server/models/cacheable_queries/user.js
```diff
@@ -1,79 +1,145 @@
...
   return userAuth
 }
 
-export async function currentEditors(redis, campaign, user) {
-  // Add user ID in case of duplicate admin names
-  const displayName = `${user.id}~${user.first_name} ${user.last_name}`
+const userHasRole = async (userId, orgId, acceptableRoles) => {
+  if (r.redis) {
+    // cached approach
+    const userKey = `texterinfo-${userId}`
```
schuyler1d: needs to use `userRoleKey(userId)`

src/server/models/cacheable_queries/user.js
```diff
@@ -1,79 +1,145 @@
...
+    let highestRole = await r.redis.hgetAsync(userKey, orgId)
```
schuyler1d: nit: make this something like `const userRoleResult` or something because it's *not* the highestRole

src/server/models/cacheable_queries/user.js
```diff
@@ -1,79 +1,145 @@
...
+        organization_id: orgId })
+      .whereIn('role', acceptableRoles)
+  )
+  return userHasRoleDb
+}
 
-  await r.redis.hsetAsync(`campaign_editors_${campaign.id}`, displayName, new Date())
-  await r.redis.expire(`campaign_editors_${campaign.id}`, 120)
+const userLoggedIn = async (authId) => {
+  const authKey = `texterauth-${authId}`
```
schuyler1d: same here -- use the userCacheKey or whatever above

src/server/models/cacheable_queries/user.js
```diff
@@ -1,79 +1,145 @@
...
+  const userAuth = await dbLoadUserAuth(authId)
+  return userAuth
+}
 
-  // Return a list of comma-separated names
-  return editors.map(editor => {
-    return editor[0].split('~')[1]
-  }).join(', ')
+const userCache = {
+  loadWithAuthId: async () => {
```
schuyler1d: if we don't need these, then let's remove the stubs.

src/server/api/errors.js
```diff
@@ -2,6 +2,7 @@ import { GraphQLError } from 'graphql/error'
 import { r, cacheableData } from '../models'
 import { userHasRole } from '../models/cacheable_queries'
```
schuyler1d: get rid of the import if we don't need it.

schuyler1d (approved):
fix lints, but then great! :+1:

PR #922, opened 2018-09-17
858 Stop Texting Toggle For Organization

### LINKED ISSUE #858: Need for the owner to stop the campaign in case of emergency.
Need the ability to stop all the campaign with one big RED button in case of emergency.

shakalee14:
Fixes #858 
- Adds toggle in organizational settings available to owner to turn off all texting for organization

**Owner Admin Settings:**
screen shot 2018-09-21 at 9 09 49 am

**Texter Todo List:**
screen shot 2018-09-21 at 9 09 54 am

**Texter Contact Screen:**
screen shot 2018-09-21 at 9 06 18 am

### DIFF
--- a/src/containers/AssignmentTexterContact.jsx
+++ b/src/containers/AssignmentTexterContact.jsx
@@ -971,6 +971,7 @@ const mapMutationsToProps = () => ({
             text
             isFromContact
           }
+          resultStatus
         }
       }
     `,
--- a/src/containers/AssignmentTexterContact.jsx
+++ b/src/containers/AssignmentTexterContact.jsx
@@ -971,7 +971,7 @@ const mapMutationsToProps = () => ({
             text
             isFromContact
           }
-          resultStatus
+          textingTurnedOff
         }
       }
     `,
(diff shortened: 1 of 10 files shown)


schuyler1d (dismissed):
besides replacing with the loaders call :+1:

shakalee14:
After testing on stage/beta - this branch has a bug with message sending. Errors because theres a query on `CampaignContact` for `textingturnedOff`

src/containers/AssignmentTexterContact.jsx
```diff
@@ -971,6 +979,7 @@ const mapMutationsToProps = () => ({
             text
             isFromContact
           }
+          textingTurnedOff
```
schuyler1d, line 982: We might want to add this to TexterTodoList.jsx as well, to move people off the /todos screen which would be consuming DB resources.

src/containers/TexterTodoList.jsx
```diff
@@ -49,23 +50,44 @@ class TexterTodoList extends React.Component {
...
+    
     const empty = (
       <Empty
         title='You have nothing to do!'
         icon={<Check />}
       />
     )
 
+    const textingOff = (
+      <Empty
```
schuyler1d, line 72: We should maybe make this a container with a route so that we redirect to a page where there's no polling.

schuyler1d (commented):
These comments aren't necessary for the feature as-is, but some code points to anticipate using this for a partial-turnoff (some texters but not others) -- and also is setup assuming a possible goal is to move to a page where DB queries will stop (i.e. no polling will occur)

schuyler1d:
This was for an emergency scenario which if we redo would probably do differently, however this was a valuable thing to have at-the-ready

PR #923, opened 2018-09-18
Surface and filter for opt-outs in message review

lperson:


joemcl:
@lperson these are the goals right:
1. Put a symbol on the conversation in message review indicating opt-out
2. Add a message status `Opt Out` to the drop down along with `Needs First Message`, `Needs Texter Response`, etc.
3. Don’t include opt-outs unless explicitly selected from message status drop down.

PR #935, opened 2018-09-21
Organization setting for customizing assignment email

### LINKED ISSUE #719: Texter Assignment Alert Email Should Be Configurable By App
You should be able to use .env to change the email body sent to texters when an assignment is ready.

jparkrr:
closes #719 

Reuses the script editor for editing these assignment emails in a way that hopefully isn't too confusing.

image

Maybe the settings should only have one save button, but I'm not sure because I'd have to change around the way texting hours and opt-out message works.

Happy to make any changes.

### DIFF
--- a/src/server/notifications.js
+++ b/src/server/notifications.js
@@ -16,6 +16,9 @@ async function getOrganizationOwner(organizationId) {
     .limit(1)
     .eqJoin('user_id', r.table('user'))('right')(0)
 }
+
+export const ASSIGNMENT_CREATED_SUBJECT = '[{organizationName}] New assignment: {campaignTitle}'
+export const ASSIGNMENT_CREATED_BODY = 'You just got a new texting assignment from {organizationName}. You can start sending texts right away: \n\n{todoUrl}'
 const sendAssignmentUserNotification = async (assignment, notification) => {
   const campaign = await Campaign.get(assignment.campaign_id)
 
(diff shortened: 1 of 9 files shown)


src/server/notifications.js
```diff
@@ -16,6 +16,9 @@ async function getOrganizationOwner(organizationId) {
     .limit(1)
     .eqJoin('user_id', r.table('user'))('right')(0)
 }
+
+export const ASSIGNMENT_CREATED_SUBJECT = '[{organizationName}] New assignment: {campaignTitle}'
+export const ASSIGNMENT_CREATED_BODY = 'You just got a new texting assignment from {organizationName}. You can start sending texts right away: \n\n{todoUrl}'
```
sandramchung: Wait, you can export a const that references another const variable that hasn't been declared yet?

sandramchung (dismissed):
Just one question about potential JS (ES6) craziness. Otherwise this looks great.

src/server/notifications.js
```diff
@@ -16,6 +17,16 @@ async function getOrganizationOwner(organizationId) {
...
     .eqJoin('user_id', r.table('user'))('right')(0)
 }
+
+const getNotificationFields = (organization, campaign) => ({
+  organizationName: organization.name,
+  campaignTitle: campaign.title,
+  todoUrl: `${process.env.BASE_URL}/app/${campaign.organization_id}/todos`
+})
+
+export const ASSIGNMENT_CREATED_SUBJECT = '[{organizationName}] New assignment: {campaignTitle}'
```
schuyler1d: nit: i get why these are kinda 'big constants', but they're not global, so I think I would vote that they stay normal camel-case variable names.

schuyler1d (dismissed):
:+1:

src/server/notifications.js
jparkrr: Do you mean e.g., `{organizationName}`? Note this isn't a template literal, it will be replaced below by a regex. (The same way texting scripts work)

jparkrr:
Thanks for the reviews. I just updated the variable names.

azuzunaga:
@jparkrr could you merge main into this branch?

jparkrr:
@azuzunaga done

shakalee14 (approved):

PR #937, opened 2018-09-21
Create question-response cache

### LINKED ISSUE #921: cache question-responses
In containers/TexterTodo.jsx from the campaign-contact object, we load:
```
questionResponseValues {
          interactionStepId
          value
        }
```

So per-campaignContact, we want to cache the interactionStepId and value -- note that we do not care about the ordering.  So probably the best redis data structure is a HASH -- i.e. `HSET response-<campaignContactId> <key=interactionStepId> <value=value>`

Creating the abstracted cacheable_query file:
 - [x] create a file src/server/models/cacheable_queries/question-response.js 
 - [x] copy the [questionResponseValues call code](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/campaign-contact.js#L35-L41) into the cacheable_queries file
 - [x] `query(campaignContactId, minimalObj)` should get from cache if cache exists and minimalObj is true with an `hgetallAsync`.  if not all of those things, then it should fallback to the db query.
 - [x] `clearQuery(campaignContactId)` calls `delAsync`
 - [x] `reloadQuery(campaignContactId)` should do the db query for interactionStepId and value, And then a multi() redis command that: 1. `del`s the key, 2. `hmset`s the results back from the db query (all at once), 3. add an `expire` -- say for 24hours

Making it operational:
 - [x] replace the api/campaign-contact.js::questionResponseValues db query with a `cacheableData.questionResponse.query(campaignContactId, minimalObj=justMinimalObj)` (see below about minimalObj)
 - [ ] minimalObj should not always equal true but only when the graphQL query is for just those two values.  You'll want to use a similar reduction/parser as here: https://github.com/MoveOnOrg/Spoke/pull/920/files#diff-20b8c071099fee69600bfaa94f19d844R59 -- maybe abstract out the hierarchy search and the items to search for so both can share code.
 - [x] call reloadQuery() in api/schema.js in updateQuestionResponses and deleteQuestionResponses

azuzunaga:
Resolves #921 

Note: For now, minimalObj is always being invoked as true.

### DIFF
--- a/src/server/models/cacheable_queries/question-response.js
+++ b/src/server/models/cacheable_queries/question-response.js
@@ -1,5 +1,11 @@
 import { r } from '../../models'
 
+/*
+HASH: response-<campaignContactId>
+key = question_response.interaction_step_id
+value = question_response.value
+*/
+
 const responseCacheKey = (campaignContactId) => (
   `${process.env.CACHE_PREFIX || ''}response-${campaignContactId}`
 )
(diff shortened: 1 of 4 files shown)


src/server/models/cacheable_queries/question-response.js
```diff
@@ -0,0 +1,53 @@
...
+    const questionResponseValues = await r.knex('question_response')
+      .where('question_response.campaign_contact_id', campaignContactId)
+      .select('value', 'interaction_step_id')
+
+    const valueInteractionArray = questionResponseValues.reduce((acc, qrv) => {
+      acc.push(qrv.interaction_step_id, qrv.value)
+      return acc
+    }, [])
+
+    if (r.redis) {
```
schuyler1d: This condition should go up to line 33 since if there's no cache, we don't need to do the db queries either.

src/server/models/cacheable_queries/question-response.js
```diff
@@ -0,0 +1,53 @@
+import { r } from '../../models'
+
+const responseCacheKey = (campaignContactId) => (
```
schuyler1d: Can you document at the top the structure i.e. HASH key=..., value=...

schuyler1d (approved):
Two nits, to tweak, but then please merge after those changes.

:+1:

PR #947, opened 2018-09-27
Admin filter people by campaign

jparkrr:
Related to what was proposed in #343 

I added a dropdown to filter the admin person view by campaign. But now I'm wondering if it would be better to instead return the campaigns associated with each texter and offer client-side filtering (on campaign or name).

Or maybe we want both? This is what I originally thought, but perhaps the added complexity isn't worth it. Feel free to close this if so.

I went with a url query parameter for the filter state so the graphql query could obtain access to it without splitting the component.

### DIFF
--- a/src/server/api/campaign.js
+++ b/src/server/api/campaign.js
@@ -14,7 +14,7 @@ export function buildCampaignQuery(queryParam, organizationId, campaignsFilter, 
   if (campaignsFilter) {
     const resultSize = (campaignsFilter.listSize ? campaignsFilter.listSize : 0)
     const pageSize = (campaignsFilter.pageSize ? campaignsFilter.pageSize : 0)
-    
+
     if ('isArchived' in campaignsFilter) {
       query = query.where({ is_archived: campaignsFilter.isArchived })
     }
@@ -111,6 +111,9 @@ export const resolvers = {
       r.table('assignment')
         .getAll(campaign.id, { index: 'campaign_id' })
         .eqJoin('user_id', r.table('user'))('right')
+        .eqJoin('user_id', r.table('user_organization')).zip()
+        .filter({ organization_id: campaign.organization_id })
+
     ),
     assignments: async (campaign, { assignmentsFilter }) => {
       let query = r.table('assignment')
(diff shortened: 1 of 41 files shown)


src/server/api/campaign.js
```diff
@@ -111,6 +111,9 @@ export const resolvers = {
       r.table('assignment')
         .getAll(campaign.id, { index: 'campaign_id' })
         .eqJoin('user_id', r.table('user'))('right')
+        .eqJoin('user_id', r.table('user_organization')).zip()
+        .filter({ organization_id: campaign.organization_id })
```
jparkrr: Do these two lines seem ok? I wanted to be able to access the `role` from `campaign.texters`

jparkrr:
Or, if you have other ideas for simplifying this, let me know.

src/server/api/campaign.js
schuyler1d: Can we convert this to a knex query?  I think it'll be more straightforward there.
Also, not sure it's relevant, but you might want to look at this:
https://github.com/MoveOnOrg/Spoke/blob/scaling-messages/src/server/api/user.js#L4
seems like if we are doing user queries by org with similar aims, maybe worth using here, too.

schuyler1d: Also, the join data won't actually be used, unless you leverage it in this method:
https://github.com/MoveOnOrg/Spoke/blob/scaling-messages/src/server/api/user.js#L39
i.e. something like `user.orgRoles ? getRoles(user.orgRoles[organizationId]) :  ...`

schuyler1d (dismissed):
please tweak sql and make sure it's leveraged in api/user.js.
Beyond that, maybe the organization.texters can take a campaignId and then you can get membership with another join that's exposed from the user object -- otherwise, we're doing the texters query twice on the backend -- not a big deal, but not great either.

jparkrr:
Thanks for the review @schuyler1d. Updated to not touch the campaign query. This way makes a lot more sense.

shakalee14:
@jparkrr - can you dig into why tests are hanging?

shakalee14 (approved):

PR #930, opened 2018-09-20
Admin order archived campaigns by created_at

jparkrr:
I went back and forth on the best way to do this (frontend or backend), but went with this because it opens the door to more sorting options in the future.

This includes a bit of linting and react warning fixes as well.

closes #926

PR #902, opened 2018-09-12
875 add a visual indicator to admin page for campaigns with unsent initial texts

azuzunaga:
Resolves #875.

Add a tag and change the campaign title color when a campaign has unsent initial messages.

image

PR #953, opened 2018-09-27
Stage main 2

shakalee14:
Includes the following PRs and branches:
* #902 from `875-visual-texts-assigned-not-sent`
* #930 from `jparkrr:jp-admin-campaign-order-by`
* #935 from `jparkrr:jp-admin-assignment-notification`


src/server/api/campaign.js
```diff
@@ -149,6 +149,26 @@ export const resolvers = {
...
+        .select('id')
+        .where({
+          campaign_id: campaign.id,
+          message_status: 'needsMessage',
+          is_opted_out: false
+         })
+        .limit(1)
+      return contacts.length > 0
+    },
+    customFields: async (campaign) => {
```
schuyler1d: This is a duplicate key of the function below, which should be preferred since it uses cache when available.  Is it broken?  Then we should fix in dbCustomFields.

My guess though is this is just a merge conflict redundancy

schuyler1d (dismissed):
After the duplicate `customFields` is removed :+1:

src/server/api/campaign.js
shakalee14: good catch - merge conflict issue

schuyler1d (dismissed):
+1

schuyler1d (approved):
Yei on keeping tests green!

PR #954, opened 2018-09-27
default to null for max_contacts assignments

schuyler1d:
null means infinite, 0 means none unless assigned.  this should be null since it's when the user logs in for the first time.


sandramchung (approved):

shakalee14:
This has been merged into `scaling-messages`.

PR #955, opened 2018-09-28
Should only alert for unsent messages for contacts who havent opted out

shakalee14:
Bug fix associated with #953

### DIFF
--- a/src/server/api/campaign.js
+++ b/src/server/api/campaign.js
@@ -158,7 +158,11 @@ export const resolvers = {
     hasUnsentInitialMessages: async (campaign) => {
       const contacts = await r.knex('campaign_contact')
         .select('id')
-        .where({ campaign_id: campaign.id, message_status: 'needsMessage' })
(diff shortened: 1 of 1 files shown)


src/server/api/campaign.js
```diff
@@ -158,7 +158,11 @@ export const resolvers = {
     hasUnsentInitialMessages: async (campaign) => {
       const contacts = await r.knex('campaign_contact')
         .select('id')
-        .where({ campaign_id: campaign.id, message_status: 'needsMessage' })
+        .where({
```
schuyler1d: Actually, can we make this a count instead of getting all the ids?
It should be
`const count = await r.getCount(r.knex.....where(...))`

schuyler1d (approved):
please change to a count call in the db.  otherwise :+1:

shakalee14:
r.getCount times out for some reason - so keeping it as a select with a limit of 1.

PR #963, opened 2018-10-03
Clear loader cache

schuyler1d:
when something is reloaded or cleared we should make sure that application state memoized cache (in DataLoader) is also cleared
See:
https://github.com/facebook/dataloader#disabling-cache

### DIFF
--- a/src/server/models/cacheable_queries/assignment.js
+++ b/src/server/models/cacheable_queries/assignment.js
@@ -1,4 +1,4 @@
-import { r } from '../../models'
+import { r, loaders } from '../../models'
 import campaignCache from './campaign'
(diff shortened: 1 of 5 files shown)


jparkrr:
in my test, I'm still seeing the campaign contact as cached in dataloader when the message status is updated. I'm not sure if that's something that needs to be fixed in the real app though

jparkrr:
hmm, now this branch isn't working for me, but it also doesn't seem to be required for my tests anymore for some reason. do you happen to know what's going on? or is it something I should look into

schuyler1d:
this is belayed by https://github.com/MoveOnOrg/Spoke/pull/1130

PR #965, opened 2018-10-04
WIP: Scaling 1.3 - dynamic assignment

schuyler1d:
This is the final piece of at least the first scaling/caching iteration, which takes on user todos and dynamic assignment optimization.

The goal is for the entire texter workflow in a dynamic assignment (or other) campaign should be completely driven by cached data that stays up-to-date.  Note that we still do DB writes on these requests (If we change that, then that will be a scaling 2.0 phase!).

The main 'new' code is in cacheable_queries/assignment-user.js and cacheable_queries/assignment-dynamic.js however it has wider consequences because caches need to refresh more aggressively and 'in concert'.

Currently there are still a bunch of bugs to flush out.

Current Issues:
 - [x] Make current tests pass
 - [x] Remove some indexes on campaign_contact -- currently too many makes DB write-slowly
 - [ ] Test caching after reassigned texters and texters assigned after campaign start
 - [ ] When to show 'Send First Texts' in dynamic assignment mode 
 - [ ] Some inflight management options for Campaign manager

Past issues (believed to be resoloved) -- would be good to check write tests for these:
 - contact id going to different assignment
 - message sent twice per-contact


shakalee14:
spoke-stage is looking good --> this isn't a regression (because i can reproduce it on prod) but question about behavior:

texter 1 gets assigned to a dynamically assigned campaign, sees the green 'send first texts' button (without a number). sends a few, returns to todos and now sees a number of texts they 'need to send'
*ignore campaign names in the screen shot*

screen shot 2018-10-13 at 1 00 20 pm 

a second texter is added to the campaign, but sees a greyed out button (even though there are texts to be sent)

screen shot 2018-10-13 at 12 58 01 pm

first texter never finishes sending first texts, but they still show up in todos

screen shot 2018-10-13 at 12 56 48 pm

to fix this, the admin should reassign them in the message review panel - or will texts from the first texter eventually be 'unassigned'?

schuyler1d:
@shakalee14 ++
I'm calling the general set of assigned contacts 'inflight' (i.e. which have been assigned but not texted) -- but that includes assigned contacts that the texter will text seconds after.

I'm calling the set of contacts that we'll want to re-assign 'stale assignments' -- we should think about how/when we want to reclaim these assignments to avoid stealing the ones that are 'inflight' -- currently, I'm doing this by ordering texters by their most recent activity in the cache, and then we can clear out the assignments if/when the texters are inactive for X time (an hour or so maybe?)

As you said, this is an issue in production, but since this branch assigns 10-at-a-time rather than just 1 it will become more urgent once we deploy this

This commit:
https://github.com/MoveOnOrg/Spoke/pull/965/commits/842a5352a076bf8fd5d34c795c05ef2473d376af
sets up a stub of an API that can *report* to the campaign admin.  I think there will be some more work to do:
 - [x] report the time of the texter's last activity
 - [ ] create a mutation for clearing them (by user maybe?)

ibrand:
Hey @schuyler1d I'm a little confused rereading these comments... It looks like this PR went to stage but it's still open? Should we close it?

schuyler1d:
Yes, please keep this open -- this represents the longterm scaling work.  It was on stage at one point but was never merged.

schuyler1d:
closing -- the scaling work was finally merged (though pulled back a bit from the original assignment structure)

PR #967, opened 2018-10-07
contacts API

### LINKED ISSUE #949: API to push contacts into a campaign
API would need managed per-organization credentials -- possibly stored in organization.features, and generateable and exposable from the organization admin console.

The (restful) API would have three endpoints:
* see contact count for a campaign
* delete contacts for a non-started, non-archived campaign
* add contacts to a non-started, non-archived campaign.  This would *not* delete existing contacts -- that way an app can push contacts incrementally.

For the last api endpoint, data should be sent in [gzipped base64 JSON](https://github.com/MoveOnOrg/Spoke/blob/main/src/workers/jobs.js#L145) -- i.e. the same data format that is sent in to editCampaign and processed from the job.

It should add them in groups (def NOT a single sql command per contact -- this is to load in millions at once).  It should not create a job -- the api client is responsible for sending in contacts in batches that the Spoke api endpoint can handle in one request.

For security, probably the best would be to have an environment variable for the instance.  In the future, we can have a UI and have a different api key per-organization.

lperson:
Things left to do (as of 2018-10-11)
- [x] Authentication
~~- [ ] Maybe enforce that it's not reentrant?~~
- [ ] Documentation

closes #949


src/server/api/contactsApi.js
```diff
@@ -0,0 +1,165 @@
...
+    return
+  }
+
+  const campaign = campaignArray[0]
+
+  let resp = {}
+
+  switch (req.method) {
+    case 'GET':
+      const count = await r
```
schuyler1d: please use `r.getCount` here to be better compatible across dbs
(https://github.com/MoveOnOrg/Spoke/blob/main/docs/EXPLANATION-development-guidelines.md#db-queries)

src/server/api/contactsApi.js
```diff
@@ -0,0 +1,165 @@
...
+
+    case 'DELETE':
+      if (campaignaStatusShortCircuit(campaign, res)) {
+        break
+      }
+
+      const deleted = await r
+        .knex('campaign_contact')
+        .delete()
+        .where({ campaign_id: campaignId })
```
schuyler1d: nit: let's move the .delete() to the end of the command, it's easier to see the alteration at the end.

src/server/api/contactsApi.js
```diff
@@ -0,0 +1,165 @@
+import { r } from '../models'
+import _ from 'lodash'
+import { getTimezoneByZip, getOptOutSubQuery } from '../../workers/jobs'
+
+import { getValidatedData } from '../../lib'
+
+function campaignaStatusShortCircuit(campaign, res) {
```
schuyler1d: nit: misspelled

src/server/api/contactsApi.js
```diff
@@ -0,0 +1,165 @@
...
+  const orgId = req.params.orgId
+  const campaignId = req.params.campaignId
+
+  if (!['GET', 'DELETE', 'POST'].includes(req.method)) {
+    res.writeHead(405, { Allow: 'GET, POST, DELETE' })
+    res.end('Not allowed')
+    return
+  }
+
+  const campaignArray = await r
```
schuyler1d: I really like the pattern `const [campaign] = await r.knex(...)... ; if (!campaign) ...` but this is fine.

src/server/api/contactsApi.js
```diff
@@ -0,0 +1,165 @@
...
+        .count('*')
+
+      resp = { contacts: { count: parseInt(count[0].count) } }
+      res.writeHead(200, { 'Content-Type': 'application/json' })
+      res.end(JSON.stringify(resp))
+      break
+
+    case 'DELETE':
+      if (campaignaStatusShortCircuit(campaign, res)) {
+        break
```
schuyler1d: should this be `break` or `return`?  it looks like the the shortCircuit call already does the res.write().

Maybe we should just have something right at the top above the switch statement -- I guess this stops GET from reporting the data for archived campaigns, but I think that's ok.  It'll make the code simpler, and if we need contacts count by api for archived campaigns, we'll probably want a fuller analytics api endpoint anyway.

src/server/api/contactsApi.js
```diff
@@ -0,0 +1,165 @@
...
+    res.writeHead(404)
+    res.end('Not found')
+    return
+  }
+
+  const campaign = campaignArray[0]
+
+  let resp = {}
+
+  switch (req.method) {
```
schuyler1d: Can we change this to if....else if statements (or separate methods)?

switch isn't understood by all devs these days, and as an example of an accidental removal/skip of 'break' from the GET would lead directly into the DELETE code.

src/server/api/contactsApi.js
```diff
@@ -0,0 +1,165 @@
...
+        'cell',
+        'external_id',
+        'zip'
+      ]
+
+      const contactsToSavePromises = validatedData.map(async contact => {
+        const contactToSave = _.pick(contact, standardFields)
+        const customFields = _.omit(contact, standardFields)
+        contactToSave.custom_fields = JSON.stringify(customFields)
+        contactToSave.timezone_offset = await getTimezoneByZip(
```
schuyler1d: we are doing this in the db query, too.  Please test for a timezone_offset column and use it instead of doing the zip lookup if it exists.

src/server/api/contactsApi.js
```diff
@@ -0,0 +1,165 @@
...
+          contactToSave.zip
+        )
+        return contactToSave
+      })
+
+      const contactsToSave = await Promise.all(contactsToSavePromises)
+
+      await r.knex
+        .transaction(async tr => {
+          const existingContacts = await tr('campaign_contact')
```
schuyler1d: please add a query parameter to skip this step.

The goal is for this API to be as pure a tunnel to batchInsert as possible

src/server/api/contactsApi.js
```diff
@@ -0,0 +1,165 @@
...
+          )
+
+          successResponse['dupesInCampaign'] =
+            contactsToSave.length - dedupedContactsToSave.length
+
+          await tr
+            .batchInsert(
+              'campaign_contact',
+              dedupedContactsToSave.map(row => {
+                row.campaign_id = campaignId
```
schuyler1d: wouldn't it be better to do this above when you're originally creating the contact?,  then you don't have to do a map at all, right?

schuyler1d (changes requested):
Looking good!
Besides feedback nits/changes, please move python code out of src/ and maybe into dev-tools/

src/server/api/contactsApi.js
lperson: 👍

src/server/api/contactsApi.js
lperson: 👍

src/server/api/contactsApi.js
lperson: 👍

src/server/api/contactsApi.js
lperson: 👍

src/server/api/contactsApi.js
lperson: 👍  This inspired a bit of refactoring beyond the specific comments here.

src/server/api/contactsApi.js
lperson: 👍

src/server/api/contactsApi.js
lperson: 👍

src/server/api/contactsApi.js
lperson: 👍

src/server/api/contactsApi.js
lperson: 👍

joemcl:
Hey @lperson does this use the OSDI API?

joshco:
@lperson @joemcl @schuyler1d I built upon Larry's pull request to make it OSDI compliant.  I put up a test server for folks to play.  I'll push my code to gh later once I clean it a bit.

# Check it out
http://spoke.dev.joshco.org/hal/browser.html#http://spoke.dev.joshco.org/osdi/org/1/campaigns/1/api/v1/

In the Custom Request Headers box, put
> OSDI-API-Token: osdi_rocks

Browse the contact collection.
I implementation a tweak on the person signup helper to allow the batched updates larry created to fit within.  More in this gist
https://gist.github.com/joshco/2f30c790c7887ac1e1e50ac6c6fd8695

joemcl:
thanks @joshco will definitely take a look

joshco:
Have a look at the gist as well.  For the bulk inserts, I can't decide if the best approach is to have a "people" attribute (which leaves the add_tag and helper functions as global to all uploaded people) or if we should have those helper functions for each individual person uploaded.

I've also been adding functionality to the ngpvan osdi-service, which people are free to use
https://github.com/joshco/osdi-service

joshco:
@lperson Here's my branch, which is based on yours
https://github.com/joshco/Spoke/tree/osdi_contacts

joshco:
@lperson @harpojaeger @joemcl 
I've put together a PR for OSDI for a people_import_helper which takes a batch of people to import.  I've been thinking about the batch import via OSDI and now believe the input representation should be an array of "signups" rather than an array of people.  This allows the helper action functions to add tags or lists to be included on each individual person in the import.  The previous was was to have an array of people with the same helper action functions applied across all people in the import.

I've updated my Spoke branch to comply with this.  One thing I could use feedback on is the response representation for the helper.  Since it's not uploading a single person, it's not appropriate to return a response for a single person object.  Instead, I riffed on @lperson 's counters/errors response, but it would be good to know what information is most useful, and feasible to return.

Feedback can be left on the PR, or on our next OSDI call:
https://github.com/opensupporter/osdi-docs/pull/335

joemcl:
So finally going to take a look at this, cc: @lperson @codygordon @deasterdaywfp

ibrand:
@lperson Are we still planning to merge this in addition to the OSDI work?

schuyler1d:
I'm going to close this old PR as we'd want to do this in the contact-loaders framework now, if we added it.

lperson:
@schuyler1d this is different than the contact-loaders.  This PR exposed an endpoint that would allow people to write scripts to push contacts into Spoke.  Is there still a use-case for that?

joemcl:
@schuyler1d is there a description or plan somewhere of the 'contact-loaders framework'? @lperson yeah there is definitely still a use-case for being able to push contacts into Spoke.

lperson:
@joemcl can you elaborate on that use case?

The contact-loader framework underlies all methods of bringing contacts into Spoke at this point --  including from the VAN/everyaction family. tldr it is empower technology for software developers to build custom importers from a variety of external systems and add them to Spoke, and then have them enabled by a system administrator.

joemcl:
@lperson OK so I found this - https://github.com/MoveOnOrg/Spoke/blob/main/docs/HOWTO-use-contact-loaders.md . So has anyone actually built something using that framework to bring in contacts from the various VAN 'flavors'? If so, any code that can be pointed to? 

@schuyler1d @lperson I'm not a developer and all that but - is the contact-loader framework a step along the way to Spoke having an API (or could it be), which would be cool, and if so why should that not be the OSDI API? Because NGPVAN, Action Network, a number of other platforms already have implemented OSDI, or at least parts.

joemcl:
@lperson a common use case is: an organization is using one of the VAN 'flavors', or Action Network, or a CRM. It wants to send a bunch of its contacts - whether voters, members, etcetera - to Spoke, along with survey questions and answers to those it has created in VAN, Action Network or the CRM. It then texts those folks in a campaign, or campaigns, in Spoke. The org will then want to be able to pull back the results of that texting into its VAN 'flavor', Action Network or CRM. Right now, that is typically done with tedious manual exports from the source to Spoke and imports back into the source . Creating an *automated* way to do that would be a huge time savings.  cc: @deasterdaywfp .

deasterdaywfp:
Agreed and this use case is real and daily as mentioned by @joemcl

lperson:
@joemcl if the use case is VAN, Spoke has fully working VAN import and almost-ready VAN update.  If there are other CRMs it is easy to build an integration.  Does that address the use case?

joemcl:
@lperson thanks cool and yeah I'm seeing a reference to a VAN integration in https://github.com/MoveOnOrg/Spoke/blob/main/docs/RELEASE_NOTES.md . I have obviously not been paying enough attention lately. That's still being tested, it looks like. Does the VAN integration include contacts syncing back and forth,  and 'survey questions' and answers syncing back and forth? Is there any documentation yet? I don't see a HOW TO doc. cc: @deasterdaywfp .

ibrand:
At the moment, we have contact loading from VAN. We should have a v1 of survey questions being able to sync back to VAN in 6.0 and once we've tested it all on prod there will be more documentation. So far there's not a lot of documentation because it's still considered "experimental" since we're rolling out parts incrementally and cause we need to test in a production scenario.

This is all so much amazing work by @lperson 🎉 😄

joemcl:
thanks @ibrand and @lperson . I hope to get that cool new VAN integration stuff loaded up on a staging / test instance soon to take it for a spin. @deasterdaywfp @lperson let's catch up re that. Which branch should I deploy?

ibrand:
@joemcl deploy the main branch. You can use the deploy to heroku button if you want to try it out. There are some env variables documented in the references (VAN API keys etc...) that you'll need to add, and then add it as a contact-loader

Issue #309, opened 2017-10-06
Add test that responds to a reply


sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/277

Issue #308, opened 2017-10-06
Add test that adds a reply message


sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/277

Issue #307, opened 2017-10-06
Add test that sends a message


sreynen:
For https://github.com/MoveOnOrg/Spoke/issues/277

PR #975, opened 2018-10-10
Texter tests

### LINKED ISSUE #951: Create tests for texter 'story arc'
these should be written against branch scaling-messages

after a campaign is created and started,
we should test the backend loading/updating/sending for a texter so that 
when messages are "sent" a message is saved, e.g. and the contact status is updated -- e.g. in future api calls -- that the counts update appropriately as the statuses change.

We should make sure that BOTH the db is correct AND that the graphQL responses (similar to backend.test.js) are returning the right things.

Furthermore, it would be good to setup travis to do a test run when caching is turned on with redis installed and used (REDIS_URL and CACHE_PREFIX=jesttest...)

Note: this is basically a superset issue for #307 #308 #309

jparkrr:
This tests the texter flow by sending graphql queries.

- adds a test that sends initial message to a test contact
- adds a test that receives a response to a test contact and replies to that response
- adds `npm run test-fakeredis` that runs all tests with the redis cache enabled (and also runs this on travis)

Feel free to let me know if there is other data I should check in these tests, or other tests I should run.

closes #951 #307 #308 #309

### DIFF
--- a/__test__/server/texter.test.js
+++ b/__test__/server/texter.test.js
@@ -63,8 +63,8 @@ it('should send an inital message to test contacts', async () => {
     false
   )
 
-  const ret2 = await runGql(getAssignmentContacts, assignVars, testTexterUser)
-  const contact = ret2.data.getAssignmentContacts[0]
+  const contactResponse = await runGql(getAssignmentContacts, assignVars, testTexterUser)
+  const contact = contactResponse.data.getAssignmentContacts[0]
 
   const message = {
     contactNumber: contact.cell,
@@ -112,8 +112,8 @@ it('should send an inital message to test contacts', async () => {
   })
 
   // Refetch the contacts via gql to check the caching
-  const ret3 = await runGql(getAssignmentContacts, assignVars, testTexterUser)
-  expect(ret3.data.getAssignmentContacts[0].messageStatus).toEqual('messaged')
+  const contactResponseAfterSending = await runGql(getAssignmentContacts, assignVars, testTexterUser)
+  expect(contactResponseAfterSending.data.getAssignmentContacts[0].messageStatus).toEqual('messaged')
 })
 
 it('should be able to receive a response and reply (using fakeService)', async () => {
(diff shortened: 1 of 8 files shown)


jparkrr:
I will still look into if I can add tests to cover more things, but this should be fine to merge if you would like.

__test__/server/texter.test.js
```diff
@@ -0,0 +1,207 @@
...
+  })
+
+  const contactsResult = await runGql(getContacts, getContactsVars, testTexterUser)
+
+  const [getAssignmentContacts, assignVars] = mutations.getAssignmentContacts(
+    contactsResult.data.assignment.contacts.map(e => e.id),
+    false
+  )
+
+  const ret2 = await runGql(getAssignmentContacts, assignVars, testTexterUser)
```
shakalee14: can we rename this to be more explicit/clear in purpose

__test__/server/texter.test.js
```diff
@@ -0,0 +1,207 @@
...
+    )
+  })
+
+  await waitForExpect(async () => {
+    const dbCampaignContact = await getCampaignContact(testContact.id)
+    expect(dbCampaignContact.message_status).toBe('needsResponse')
+  })
+
+  // Refetch the contacts via gql to check the caching
+  const ret3 = await runGql(getAssignmentContacts, assignVars, testTexterUser)
```
shakalee14: can we rename this to be more explicit/clear in purpose

__test__/server/texter.test.js
```diff
@@ -0,0 +1,207 @@
...
+      expect.objectContaining({
+        send_status: 'SENT'
+      })
+    )
+    const dbCampaignContact = await getCampaignContact(testContact.id)
+    expect(dbCampaignContact.message_status).toBe('convo')
+  })
+
+  // Refetch the contacts via gql to check the caching
+  const ret4 = await runGql(getAssignmentContacts, assignVars, testTexterUser)
```
shakalee14, line 205: can we rename this to be more explicit/clear in purpose

shakalee14 (approved):

schuyler1d:
This was merged in https://github.com/MoveOnOrg/Spoke/pull/1095

PR #986, opened 2018-10-16
connect opportunistically to DataWarehouse

schuyler1d:
until/if we need it

This should address created/dropped connections.


mathemagica (approved):
👍

PR #989, opened 2018-10-18
a super-light change to track simultaneous texters

schuyler1d:
across a five-minute window -- tracking it will requires other tools looking at the redis instance


schuyler1d:
closing as we found a different way

PR #993, opened 2018-10-22
accept twilio messages with special characters

schuyler1d:
This is a workaround from Twilio sending X-Twilio-Body with the full message and AWS/Lambda/express library rejecting the header values as invalid when there are special characters.

Examples of special characters that were marked invalid:
1. `'I\u0019m sorry. I do not know you....`
2. `'Liked \u001cHi, .... ` (from an iMessage feature where the contact can 'like' or 'dislike' a message)

This has been tested on stage.

### DIFF
--- a/lambda.js
+++ b/lambda.js
@@ -11,6 +11,17 @@ const jobs = require('./build/server/workers/job-processes')
 // See: http://docs.aws.amazon.com/lambda/latest/dg/best-practices.html#function-code
 // "Separate the Lambda handler (entry point) from your core logic"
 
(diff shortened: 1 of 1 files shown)


shakalee14 (commented):
is there a test we can write to accompany this?

schuyler1d:
I added a test (https://github.com/MoveOnOrg/Spoke/pull/1001), but before the election I think we shouldn't merge the test, but we should merge this

shakalee14 (approved):

PR #994, opened 2018-10-23
delete duplicates after the job is complete

schuyler1d:
-- this guards against multiple jobs and other accidents


shakalee14:
We created this thinking we were uploading duplicates - at some point we should deploy this (maybe after elections).

The actual fix was adding an `order by column` to the query in spoke --> this ensures that the limit / offset subqueries consistently grabs the contacts in order vs a random order.

shakalee14 (approved):

shakalee14:
Merged with #1031

PR #996, opened 2018-10-23
People admin pagination

schuyler1d:
This adds a crude pagination drop-down to avoid trying to load too many texters into a single page, capping at 200.

After testing on stage, we could potentially roll this out to beta-only, since only our admin uses the page, and they can access it there.

### DIFF
--- a/src/containers/AdminPersonList.jsx
+++ b/src/containers/AdminPersonList.jsx
@@ -79,7 +79,7 @@ class AdminPersonList extends React.Component {
     const LIMIT = 200
     const { personData: { organization} } = this.props
     if (organization.peopleCount > LIMIT) {
-      const offsetList = Array.apply(null, { length: parseInt(organization.peopleCount / LIMIT, 10) })
+      const offsetList = Array.apply(null, { length: Math.ceil(organization.peopleCount / LIMIT) })
       return (
         <DropDownMenu
           value={Number(this.props.location.query.offset || 0)}
--- a/src/containers/AdminPersonList.jsx
+++ b/src/containers/AdminPersonList.jsx
@@ -19,15 +19,6 @@ import gql from 'graphql-tag'
 import { dataTest } from '../lib/attributes'
 import LoadingIndicator from '../components/LoadingIndicator'
 
-const organizationFragment = `
-  id
-  people(campaignId: $campaignId) {
-    id
-    displayName
-    email
-    roles(organizationId: $organizationId)
-  }
-`
 class AdminPersonList extends React.Component {
 
   constructor(props) {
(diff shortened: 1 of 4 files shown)


src/containers/AdminPersonList.jsx
```diff
@@ -43,13 +34,23 @@ class AdminPersonList extends React.Component {
...
-    const query = value ? `?campaignId=${value}` : ''
+  handleFilterChange = (campaignId, offset) => {
+    let query = '?' + (campaignId ? `campaignId=${campaignId}` : '')
+    query += (offset ? `&offset=${offset}` : '')
     this.props.router.push(
       `/admin/${this.props.params.organizationId}/people${query}`
     )
   }
 
+  handleCampaignChange = (event, index, value) => {
```
shakalee14, line 45: what are we using event and index for here? is this setting up for something else in the future?

shakalee14: same question for `handleCampaignChange` and `handleOffsetChange`

src/server/api/organization.js
```diff
@@ -32,9 +32,21 @@ export const resolvers = {
       return r.table('opt_out')
         .getAll(organization.id, { index: 'organization_id' })
     },
-    people: async (organization, { role, campaignId }, { user }) => {
+    people: async (organization, { role, campaignId, offset }, { user }) => {
       await accessRequired(user, organization.id, 'SUPERVOLUNTEER')
-      return buildUserOrganizationQuery(r.knex.select('user.*'), organization.id, role, campaignId)
+      const query = buildUserOrganizationQuery(
```
shakalee14, line 37: did you and @erikkirkstein huddle about changing this to most recently created?

src/containers/AdminPersonList.jsx
schuyler1d: It seems to just be what arguments are sent on a DropDown change event -- I am just copying the old handleFilterChange method (on the left :-)
Before that was handling the only filterchange -- the champaign.  I changed that to be a more abstracted method that does the redirect, and then we now have two methods that handle different formchange events.

This is all kinda hacky -- the 'right way' would be that we'd add a state and refresh a query value, but apollo/graphql makes that so annoying that whoever added the campaign query did this instead which feels reasonable given how few people are on this page.

src/server/api/organization.js
schuyler1d: Yeah. He preferred reverse by created-at, since most issues are related to new users.  It does make hunting for a specific user annoying but
1. that's only an issue for teams with >200 anyway, which is already a pretty small set of orgs
2. looking forward to Justin's PR that allows you to do a search

PR #998, opened 2018-10-24
Fixes #818

### LINKED ISSUE #818: Strip leading and trailing whitespace from script variables
Script variables include leading and trailing space when inserted. Trailing whitespace especially can be hard to catch in a spreadsheet editor but leads to texts looking like the following (see spacing around the first comma) that can be difficult to debug:

android

yeti-detective:
screenshot of what I did

Issue #818 reported that contacts uploaded from CSV with trailing whitespace were formatted incorrectly when sending texts: _"ConstituentFN , ConstitutentLN "_ would appear like: _Hello, ConstitutentFN , your last name is ConstituentLN ._

I implemented the vanilla JS string method [String.prototype.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) within the function `capitalize` which appears to only be called on FirstName and LastName data from the db before it is inserted into the script.

I chose this method over trimming CSV input fields on insertion to the database, because that approach would have required users to re-upload their contact.csv files, and some of the files may no longer exist.

Also, I linted the file because my editor lints on save. Sorry, or you're welcome, depending on how you feel about that. :smiley:

### DIFF
--- a/src/lib/scripts.js
+++ b/src/lib/scripts.js
@@ -1,52 +1,70 @@
 export const delimiters = {
-  startDelimiter: '{',
-  endDelimiter: '}'
-}
+  startDelimiter: "{",
+  endDelimiter: "}"
+};
 
-export const delimit = (text) => {
-  const { startDelimiter, endDelimiter } = delimiters
-  return `${startDelimiter}${text}${endDelimiter}`
-}
+export const delimit = text => {
+  const { startDelimiter, endDelimiter } = delimiters;
+  return `${startDelimiter}${text}${endDelimiter}`;
+};
 
 // const REQUIRED_UPLOAD_FIELDS = ['firstName', 'lastName', 'cell']
-const TOP_LEVEL_UPLOAD_FIELDS = ['firstName', 'lastName', 'cell', 'zip', 'external_id']
-const TEXTER_SCRIPT_FIELDS = ['texterFirstName', 'texterLastName']
+const TOP_LEVEL_UPLOAD_FIELDS = [
+  "firstName",
+  "lastName",
+  "cell",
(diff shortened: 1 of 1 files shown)


src/lib/scripts.js
```diff
@@ -1,52 +1,70 @@
...
+];
 
 // TODO: This will include zipCode even if you ddin't upload it
-export const allScriptFields = (customFields) => TOP_LEVEL_UPLOAD_FIELDS.concat(TEXTER_SCRIPT_FIELDS).concat(customFields)
+export const allScriptFields = customFields =>
+  TOP_LEVEL_UPLOAD_FIELDS.concat(TEXTER_SCRIPT_FIELDS).concat(customFields);
 
-const capitalize = str => { return str.charAt(0).toUpperCase() + str.slice(1).toLowerCase() }
+const capitalize = str => {
+  const strTrimmed = str.trim();
```
schuyler1d, line 34: This looks like the actual change, here.

schuyler1d (approved):
Thanks!  In the future, please separate out the lint from your file changes into a separate commit -- otherwise, it's very hard to pick out the actual change -- both in the PR and for posterity.

src/lib/scripts.js
yeti-detective: I will do. Thank you for the feedback.

yeti-detective: That is the actual change, yes.
In the future, since my linter is automatic (and for other people with automatic linting set up) I will save the file before making any changes, triggering the linter, then commit the linted changes first.

shakalee14:
Merged with #1031

PR #1008, opened 2018-11-02
change page size to all

shakalee14:


schuyler1d (approved):
I feel like we limited this because it was slowing things down significantly, but I guess we upgrade the server a bunch.....

shakalee14:
Merged with #1031

PR #1009, opened 2018-11-02
skip auth0: create an admin link to share to a texter to skipo login

schuyler1d:
so they can login without going through auth0 -- this only works for existing accounts and requires an environment variable to be present.

This is a disaster-preparedness PR, for a situation where auth0 goes down.


schuyler1d:
closing

PR #1011, opened 2018-11-05
creating an org and then joining

schuyler1d:
-- we only want to save a single role now, since we now assume an hierarchy.


shakalee14 (approved):

shakalee14:
Fixes https://github.com/MoveOnOrg/Spoke/issues/1017

PR #1012, opened 2018-11-05
retry sending upon a twilio error once

schuyler1d:
We seem to get a lot of these errors:
https://github.com/twilio/twilio-node/issues/317
on AWS Lambda platform.
This just does a retry a single time.  If we want to do something more sophisticated then we should consider using node-retry as suggested here:
https://github.com/twilio/twilio-node/issues/338#issuecomment-391902892 (followed from the 317 issue link)


shakalee14:
@schuyler1d are we still considering doing this? i'd imagine we'd go with the suggested way from twilio folks with node-retry

schuyler1d:
I don't think we should use node-retry.

I think the number of these errors turned out to be low, do i think if rather just close this.

PR #1020, opened 2018-11-21
Unique campaign contact cell and campaign_id migration

benmort:
Add migration to create index on campaign contact unique to campaign_id and cell

With the injection of campaign contacts we're needing to restrict rows to unique
combination of cell and campaign id to stop injection of duplicate records.

### DIFF
--- a/src/migrations/index.js
+++ b/src/migrations/index.js
@@ -156,6 +156,20 @@ const migrations = [
       console.log('added texting hours fields to campaign')
     }
   }
+  { auto: true, //13
+    date: '2018-10-26',
+    // eslint-disable-next-line
+    migrate: async function() {
+      // it is ok if this function fails if run again, but
+      // it should be ok to be run twice.  If not, then make auto=false
+      await r.knex.raw(`
+        CREATE UNIQUE INDEX "campaign_contact_cell_campaign_id_idx"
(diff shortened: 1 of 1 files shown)


ibrand:
@schuyler1d given #1154 I want to make sure this is still the right way to do a migration. I have a feeling it's not. Is there guidance we can give @benmort on this?

schuyler1d:
No, it's not after 1154.  Honestly, I would recommend we merge 1154 this round, and then after we've cut a release with that, we can review PRs with migrations and do them the new way.

One flag on this particular PR -- the current way we remove dupes is 'post-facto' of import in some cases (e.g. with datawarehouse import) -- For large imports that might be preferred rather than the database erroring on the entire import with one mistaken duplicate.

ibrand:
@benmort and @schuyler1d What are the next steps here?

schuyler1d:
I'd like to close this.  The challenge with having a hard DB constraint is that we can't fix it or address it at the application level.  The ingest-loader framework that we just merged, runs de-duping on all contact loads:
https://github.com/MoveOnOrg/Spoke/pull/1369/files#diff-ee93a980f871b21055cf3e181d6583beR232
(and can report those issues without the DB breaking the import completely).

Please re-open, if you think it still makes sense to have this constraint and argue the case.

PR #654, opened 2018-06-15
Use relative links in all docs MD

harpojaeger:
Closes #653

PR #1026, opened 2018-12-05
Adding cross-reference to Exports doc

joemcl:
Added link to https://github.com/MoveOnOrg/Spoke/blob/main/docs/DATA_EXPORTING.md .


bdatkins (dismissed):

harpojaeger (changes requested):
2 comments: 1) the text of the link is ambiguous and doesn't make it clear what the doc we're linking to is about. Since these are setup docs targeted at beginners, we should be as clear and specific as possible. 2) We should use relative links whenever possible so the directoey hierarchy is preserved across forks (e.g. #654: https://github.com/MoveOnOrg/Spoke/commit/56ad367070374185003142b562d13a7bd809c35b#diff-68f7560bfa04421fcb343b9486650667)

joemcl:
@harpojaeger , thanks, updated now.

joemcl:
and @bdatkins this is an automatic thing, apparently - ```@joemcl joemcl dismissed bdatkins’s stale review via 7d53ba1 3 minutes ago```

bdatkins (dismissed):

shakalee14 (approved):

joemcl:
thanks @shakalee14 !

PR #1001, opened 2018-10-25
Lambda accept special chartexts tests

schuyler1d:
this adds a test for PR https://github.com/MoveOnOrg/Spoke/pull/993
however the changes for the test are much riskier than the change itself!

So I'm separating this PR for us to merge post-election

PR #1031, opened 2018-12-24
Stage main 2

shakalee14:
#994: Deleting duplicate jobs apart of SQL warehouse work
#1008: Changing pagination default to all for campaigns
#1011: Fixing creating org and joining bug 
#998: Fixing trailing spaces in fields in scripts
#1001: lambda special chars tests 
#1033: stop gap for 'no org' message for organization owners after login
#1025: improvements to message review board, migration to message table (send_before), etc

Migrations:
send_before (timestamp) on message table


schuyler1d (approved):
:+1:

PR #1033, opened 2018-12-27
Fixes #1032

### LINKED ISSUE #1032: [Bug] Admin Sees "You Are Not Apart of An Organization" Message After Login
This bug is important because it degrades the owner user experience

shakalee14:
Actually an issue if user is an owner in the db.


schuyler1d (approved):
This is ok as a stop-gap, so :+1: 

The downside of this approach is that we add an extra db request on-load.  The end goal (true in the caching branch, but could be a separate step here) is just to make `role` arg filter based on acceptable roles rather than the exact role -- and then we can just call it with `role:"TEXTER"` and get all of them.

PR #1038, opened 2019-01-28
Fix send_before column type

bchrobot:
`send_before` should be an optional column, not a required one. `migrations.js` declares it as optional and there are a few places in the codebase where it is checked for truthyness indicating it could be `null`. This causes crashes if texting hours are not enforced.

Issue stems from this block: https://github.com/moveonorg/Spoke/blob/main/src/server/api/schema.js#L942-L972

On line 942, `getSendBeforeTimeUtc()` returns `null` if `campaign.texting_hours_enforced === false`. This results in `sendBeforeDate = null` being passed to `Message:send_before`.

### DIFF
--- a/src/server/models/custom-types.js
+++ b/src/server/models/custom-types.js
@@ -31,3 +31,8 @@ export function timestamp() {
     .allowNull(false)
     .default(r.now())
 }
+
+export function optionalTimestamp() {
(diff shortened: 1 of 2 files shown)


schuyler1d (approved):
:+1:

PR #1039, opened 2019-01-29
Fix Auth0 browser compatibility

### LINKED ISSUE #776: New texters reporting Spoke login issues when using Google Chrome
A handful of new MoveOn texters are reporting an issue setting up their Spoke accounts using Google Chrome as their web browser.

Each texter says that when the attempt to login after creating their account, Spoke just sends them back to the login screen again, and this repeats no matter how many times they try to login. They do not see a message saying their is anything wrong with their username or password.

### LINKED ISSUE #995: Longtime texters can not longer log into Spoke via iPhone/Firefox/Chrome
A growing challenge we are seeing is team members with existing Spoke accounts no longer being able to log in to Spoke via their regular devices or browsers.

In nearly all cases, they attempt to log in and just get sent back to the login screen. This process just repeats, and cache and cookie clearing does not solve the issue.

In some cases, even changing browsers does not help, and in some cases it does.

This seems most acute with iPhone users and FireFox desktop users.

bchrobot:
Various browsers block third-party cookies when doing Cross-Origin authentication. This breaks Auth0's embedded Lock unless you pay for a custom domain with Auth0 to keep requests on the same domain. This PR migrates from Auth0 Lock to Auth0 Universal Login page.

From [Auth0 Employee](https://community.auth0.com/t/unable-to-configure-verification-page-error-on-hosted-login/9214/20):

> [If] using embedded login, custom domain becomes almost a necessity.
> ...
> [We] recommend either using custom domains or moving to universal login (hosted login page).

[Rendered readme changes](https://github.com/bchrobot/Spoke/blob/fix-auth0-compatibility/README.md#getting-started).

This fixes #995 and fixes #776.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,38 @@ context.idToken["https://spoke/user_metadata"] = user.user_metadata;
 callback(null, user, context);
 }
 ```
-11. If the application is still running from step 8, kill the process and re-run `npm run dev` to restart the app. Wait until you see both "Node app is running ..." and "webpack: Compiled successfully." before attempting to connect. (make sure environment variable `JOBS_SAME_PROCESS=1`)
-12. Go to `http://localhost:3000` to load the app.
-13. As long as you leave `SUPPRESS_SELF_INVITE=` blank and unset in your `.env` you should be able to invite yourself from the homepage.
+11. Update the Auth0 [Universal Landing page](https://manage.auth0.com/#/login_page) to include custom fields. You may also include other Auth0 Lock [configuration options](https://auth0.com/docs/libraries/lock/v11/configuration).
+    ```javascript
+    ...
+
+    var lock = new Auth0Lock(config.clientID, config.auth0Domain, {
+      ...
+
+      allowedConnections: ['Username-Password-Authentication'],
+      additionalSignUpFields: [{
+        name: 'given_name',
+        icon: 'https://upload.wikimedia.org/wikipedia/commons/c/ca/1x1.png',
+        placeholder: 'First Name'
+      }, {
+        name: 'family_name',
+        placeholder: 'Last Name',
+        icon: 'https://upload.wikimedia.org/wikipedia/commons/c/ca/1x1.png'
+      }, {
+        name: 'cell',
+        placeholder: 'Cell Phone',
+        icon: 'https://upload.wikimedia.org/wikipedia/commons/c/ca/1x1.png',
+        validator: (cell) => ({
+          valid: cell.length >= 10,
+          hint: 'Must be a valid phone number'
+        })
+      }],
+
+      ...
+    });
+    ```
+12. If the application is still running from step 8, kill the process and re-run `npm run dev` to restart the app. Wait until you see both "Node app is running ..." and "webpack: Compiled successfully." before attempting to connect. (make sure environment variable `JOBS_SAME_PROCESS=1`)
+13. Go to `http://localhost:3000` to load the app.
+14. As long as you leave `SUPPRESS_SELF_INVITE=` blank and unset in your `.env` you should be able to invite yourself from the homepage.
     - If you DO set that variable, then spoke will be invite-only and you'll need to generate an invite. Run:
 ```
 echo "INSERT INTO invite (hash,is_valid) VALUES ('abc', 1);" |sqlite3 mydb.sqlite
@@ -54,7 +83,7 @@ echo "INSERT INTO invite (hash,is_valid) VALUES ('abc', 1);" |sqlite3 mydb.sqlit
 ```
   - Then use the generated key to visit an invite link, e.g.: http://localhost:3000/invite/abc. This should redirect you to the login screen. Use the "Sign Up" option to create your account.
 
-14. You should then be prompted to create an organization. Create it.
+15. You should then be prompted to create an organization. Create it.
 
 If you want to create an invite via the home page "Login and get started" link, make sure your `SUPPRESS_SELF_INVITE` variable is not set.
 
--- a/README.md
+++ b/README.md
@@ -44,34 +44,99 @@ context.idToken["https://spoke/user_metadata"] = user.user_metadata;
 callback(null, user, context);
 }
 ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -44,17 +44,46 @@ context.idToken["https://spoke/user_metadata"] = user.user_metadata;
 callback(null, user, context);
 }
 ```
-11. If the application is still running from step 8, kill the process and re-run `npm run dev` to restart the app. Wait until you see both "Node app is running ..." and "webpack: Compiled successfully." before attempting to connect. (make sure environment variable `JOBS_SAME_PROCESS=1`)
-12. Go to `http://localhost:3000` to load the app.
-13. As long as you leave `SUPPRESS_SELF_INVITE=` blank and unset in your `.env` you should be able to invite yourself from the homepage.
+11. Update the Auth0 [Universal Landing page](https://manage.auth0.com/#/login_page) to include custom fields. You may also include other Auth0 Lock [configuration options](https://auth0.com/docs/libraries/lock/v11/configuration).
```
shakalee14: For the instructions for this part, is the work flow?

1) Visit [Universal Landing Page](https://manage.auth0.com/#/login_page), click on the `customize login page` toggle, and then update the javascript within the script tag? If this is true, can we add the exact copy and paste (so for the HTML and script tag). Then the instructions can be - copy and paste following code in `Default Templates` space.

bchrobot: That is correct, yes. I debated including the fully contents, but chose not to for the following reasons:

1. The only changes Spoke actually _requires_ are the three additional fields.
2. Auth0 may change the contents of the universal page at some point so keeping the Spoke-specific changes to page as small as possible seems like a good idea.
3. Customization like the terms of service link (and maybe non-Spoke theming) require additional user setup.
4. It's a lot of text to add to the readme.

I think you are right that it should be added though. Regarding 1 and 2: Auth0 probably won't update their universal login page in a breaking way, and if they do I'm sure they would email out beforehand to users that have enabled customization. Regarding 3: as long as here is a link to configuration options, then users can go the extra mile to customize. Regarding 4: this is a poor reason to begin with haha

shakalee14: K - just checking - The other idea is that we can add a link to a file somewhere else in the project to use for the copy.

bchrobot: Not sure if you saw, but I updated the readme doc to be more explicit!

shakalee14: yes, see it! thanks!

shakalee14 (approved):

README.md
jlegrone: > It's a lot of text to add to the readme.

<details><summary>You could hide the html behind a dropdown 😄</summary>

```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
  <title>Sign In with Auth0</title>
  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
</head>
<body>

  <!--[if IE 8]>
  <script src="//cdnjs.cloudflare.com/ajax/libs/ie8/0.2.5/ie8.js"></script>
  <![endif]-->

  <!--[if lte IE 9]>
  <script src="https://cdn.auth0.com/js/base64.js"></script>
  <script src="https://cdn.auth0.com/js/es5-shim.min.js"></script>
  <![endif]-->

  <script src="https://cdn.auth0.com/js/lock/11.11/lock.min.js"></script>
  <script>
    // Decode utf8 characters properly
    var config = JSON.parse(decodeURIComponent(escape(window.atob('@@config@@'))));
    config.extraParams = config.extraParams || {};
    var connection = config.connection;
    var prompt = config.prompt;
    var languageDictionary;
    var language;

    if (config.dict && config.dict.signin && config.dict.signin.title) {
      languageDictionary = { title: config.dict.signin.title };
    } else if (typeof config.dict === 'string') {
      language = config.dict;
    }
    var loginHint = config.extraParams.login_hint;

    // Available Lock configuration options: https://auth0.com/docs/libraries/lock/v11/configuration
    var lock = new Auth0Lock(config.clientID, config.auth0Domain, {
      auth: {
        redirectUrl: config.callbackURL,
        responseType: (config.internalOptions || {}).response_type ||
          (config.callbackOnLocationHash ? 'token' : 'code'),
        params: config.internalOptions
      },
      /* additional configuration needed for custom domains
      configurationBaseUrl: config.clientConfigurationBaseUrl,
      overrides: {
        __tenant: config.auth0Tenant,
        __token_issuer: 'YOUR_CUSTOM_DOMAIN'
      }, */
      assetsUrl:  config.assetsUrl,
      allowedConnections: ['Username-Password-Authentication'],
      rememberLastLogin: !prompt,
      language: language,
      languageDictionary: {
        title: 'Spoke',
        signUpTerms: 'I agree to the <a href="YOUR_LINK HERE" target="_new">terms of service and privacy policy</a>.'
      },
      mustAcceptTerms: true,
      theme: {
        logo:            '',
        primaryColor:    'rgb(83, 180, 119)'
      },
      additionalSignUpFields: [{
        name: 'given_name',
        icon: 'https://upload.wikimedia.org/wikipedia/commons/c/ca/1x1.png',
        placeholder: 'First Name'
      }, {
        name: 'family_name',
        placeholder: 'Last Name',
        icon: 'https://upload.wikimedia.org/wikipedia/commons/c/ca/1x1.png'
      }, {
        name: 'cell',
        placeholder: 'Cell Phone',
        icon: 'https://upload.wikimedia.org/wikipedia/commons/c/ca/1x1.png',
        validator: (cell) => ({
          valid: cell.length >= 10,
          hint: 'Must be a valid phone number'
        })
      }],
      prefill: loginHint ? { email: loginHint, username: loginHint } : null,
      closable: false,
      defaultADUsernameFromEmailPrefix: false,
      // uncomment if you want small buttons for social providers
      // socialButtonStyle: 'small'
    });

    lock.show();
  </script>
</body>
</html>
```

</details>

---

https://gist.github.com/citrusui/07978f14b11adada364ff901e27c7f61

jlegrone (approved):
This works well for me locally!

Note that if the user has cookies blocked, authentication will still fail. As far as I can tell there is no workaround for this, but I'll make a note on the linked issues with troubleshooting steps.

PR #1045, opened 2019-01-31
Hide 'Teams' nav menu from normal texter accounts.

bchrobot:
This removes the Teams list and header from the dropdown navigation menu in the upper left for users with only the `TEXTER` role.

A user's roles must be looked up separately, hence why each team `MenuItem` is wrapped in a GraphQL query. These queries are cached and do not fire off each time the menu is opened.

The `TeamMenu` intermediate component is introduced pull out state management for hiding/showing teams and the teams heading.

### DIFF
--- a/src/containers/UserMenu.jsx
+++ b/src/containers/UserMenu.jsx
@@ -11,8 +11,106 @@ import { connect } from 'react-apollo'
 import { withRouter } from 'react-router'
 import gql from 'graphql-tag'
 import { dataTest } from '../lib/attributes'
+import { hasRole } from '../lib/permissions'
 
 const avatarSize = 28
+
+class OrganizationItemInner extends Component {
+  componentDidMount() {
+    this.updateHasTeam()
+  }
+
+  componentDidUpdate() {
+    this.updateHasTeam()
+  }
+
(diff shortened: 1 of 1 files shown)


joemcl:
@bchrobot - 
- You mean the dropdown nav menu in the upper right, yes?
- Why do we use the word Teams, anyway? Should it not be Organizations?

shakalee14 (approved):

schuyler1d (changes requested):
I think I object to the purpose of this PR.  Why should the 'teams' be hidden from the texter if they are part of multiple organizations?
Workflow might be "login to spoke.example.com" and if they are part of multiple teams, having the menu would be the only way they could get to the right view.

If someone really shouldn't see the mulitple orgs they are in then have an environment variable for that choice.

Also on the implementation side, even if the client caches this, it's still too many calls -- especially a 'thundering herd of volunteers'.  If this were implemented it should call `user.organizations(role='SUPERVOLUNTEER')` from the parent component and then just hide/show the whole menu depending on if there are any answers.

schuyler1d:
I'd be :+1: to adding (or maybe even moving) those routes.

PR #1050, opened 2019-02-03
Fix for 904 campaign starts without scripts

### LINKED ISSUE #904: Campaigns are allowed to start without scripts
Texters can type in messages without scripts, however they error out when trying to send the message:
image

Fixes: 
- Don't allow campaign to start without scripts
- Prevent texter from entering text when the campaign doesn't have a script

azuzunaga:
Resolves #904.

The `AdminCampaignEdit.state.campaignFormValues.interactionSteps` array can still have an item in it, even if the script is not filled in. So rather than check for length, check that the `script` key in any of the `interactionSteps` array items is not an empty string.


harpojaeger:
Nice!

shakalee14 (approved):

PR #1053, opened 2019-02-05
Support Heroku Review Apps

### LINKED ISSUE #1052: Enable Heroku Review Apps
This will make it easier to do incremental QA and give us some peace of mind about potential configuration drift issues.

Once review apps are enabled, it would also be great to set up a seed script with some standard dummy data.

jlegrone:
Closes #1052 

This builds on https://github.com/MoveOnOrg/Spoke/pull/1039. Since that change makes it possible to use auth0 without cross origin requests, we are able to configure an auth0 instance for all review apps with a simple catch-all subdomain `https://*.herokuapp.com`. *This configuration should NOT be used in a production app!*

screen shot 2019-02-04 at 6 03 51 pm

## Major changes in this PR
1. Update dockerfile to build directly from source. In the future we can [automate pushes to a public registry](https://docs.docker.com/docker-hub/builds/) and use git tags to make supported versions available without the need to clone and build from source. Also, image size was decreased from 1.17GB to 284MB! 😄 

2. Use Heroku container runtime directly rather than nodejs (this ensures we always have a buildable dockerfile that can also be deployed with docker-compose, kubernetes, etc)

3. `window.location` is used client side if `BASE_URL` is not set (this was tricky to do consistently in review apps and one-click deploys... if there is a way to enable the [heroku metadata](https://devcenter.heroku.com/articles/dyno-metadata) labs feature on regular apps at provisioning time without the cli please let me know!)

4. Docker builds use yarn, and a yarn lockfile has been committed. This change isn't super necessary and can be removed, but there are lots of benefits to using lockfiles. It would be nice to take advantage of other yarn features, like [workspaces](https://yarnpkg.com/lang/en/docs/workspaces/), in the future.

5. At least one of our dependencies had dropped support for node 6, so I also updated the engines config to allow higher versions and used node 8.10 (the latest supported by AWS Lambda) in our dockerfile.


harpojaeger:
This is great. For ease of testing, I would suggest that steps 1, 4 and 5 be factored out as separate PRs: I think all three of those are changes we should consider regardless of what decision is made about Heroku review apps, and I also think each of them is significant enough on its own that they should be thoroughly tested in isolation. FTR, I support commiting all of them, but it'd be good to have discussion & code review on each of them individually.

jlegrone:
So in response to factoring things out in separate PRs, I was hoping to get things merged together since all of these changes are required for review apps (with the exception of yarn), and this is also a base for further work on containerization and scaling I have planned.

To get relatively deterministic docker builds from source, we need to be using yarn or a newer version of npm, both of which require node 8. Node 6 runtimes should still work fine as well though outside of docker. No changes in this PR (other than Ben's) should introduce functional changes or impact bundle size, so I feel like it's reasonable to validate them together.

shakalee14:
Do we need this `yarn.lock` file @jlegrone ?

jlegrone:
@shakalee14 yes if we are okay with yarn (I have some future work that I'd like to do with workspaces so this would be nice). Otherwise we could use npm lockfiles. Either yarn or npm 5 are required if we want deterministic docker builds.

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Spoke",
   "main": "src/server",
   "engines": {
-    "node": "6.10.3",
+    "node": ">=6.10.3",
```
schuyler1d: This should be `8.10` and not just anything -- otherwise e.g. node 10 will break lambda

jlegrone: I'm not super familiar with the Lambda node runtime -- would anything that's not an exact match cause Lambda deployments to fail?

My thought is that we shouldn't _disallow_ folks from running Spoke with the node version of their choosing, which provides flexibility for those who want to take advantage of improvements in newer runtimes.

schuyler1d: Yes, unfortunately that's how it works.
https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html
(yes, custom runtimes could in theory open other possibilities, but I think easy setup is an important goal/target for this project)

jlegrone: So lambda actually picks up on the engine value from `package.json`? Or is that maybe a claudia thing? I just took a quick look but I don't see anything about package.json in the Lambda runtime docs.

harpojaeger:
I'm generally in favor of specifying dependency & engine versions as sppecifically as possible, to reduce vulnerabilities down the line (i.e. forcing the software to use known-good packages, and not allowing it to install whatever the newest version of a thing is on NPM, which increases susceptibility to attacks like this one: https://github.com/dominictarr/event-stream/issues/116). Not quite as concerned about this with Node, but I think the principle still applies, and specifying an exact version is preferable.

Also, what's the value added of users being able to run under the Node version of their choice?

package.json
schuyler1d: Probably a claudia thing -- either way, I'm with @harpojaeger -- I think being specific is better because then it's a real target for devs to code within and for deployers/maintainers to depend on.  Otherwise, when do we say "no you can't use X because it's not supported" -- never.  But then deployers have no idea what to target or have to use the bleeding edge -- often not supported as widely in hosting contexts.

jlegrone: I haven't been able to reproduce the issue with claudia failing to build when the node version is not a pinned version. I don't have an AWS account handy for testing lambda unfortunately, but I can't find anything to indicate unpinning could cause problems in lambda either.

I think avoiding pinning actually enhances security, since users and managed services won't have to fight tooling to consume critical security patches to their runtimes without first modifying the engine configuration in source/upstream.

See more discussion here on security and support: https://github.com/spokecommunity/Spoke/pull/5#discussion_r267599167

I did go ahead and add an upper version limit so that only LTS node releases will match. If we don't already, then I think clearly documenting version support policy in a more human-centric format would be useful.

schuyler1d: I'm not saying that it will fail to build if the version is not pinned.  However, if we do not pin the version, then people will inevitably write code that requires version 9, and then people running an environment with node version 8 will be unable to run.

++ to documenting it in the readme and/or contributing that we are currently targeting version 8.10.  I still think we should have something pinned in the package.json, but maybe I'm not understanding what happens if we have a pinned version -- does that mean node version 9 will refuse to run it if it says '`"node": "8.10"` (or 6.10)?

jlegrone: > what happens if we have a pinned version -- does that mean node version 9 will refuse to run it if it says '"node": "8.10" (or 6.10)?

No, node and npm (at least up to node 10 and npm 5 so far) do not pay attention to engines at runtime. The engines section of package.json was introduced only in order to make sure that your stuff could work with other stuff when consumed as an npm dependency, so most tools don't care about the top level package. Yarn is an exception, and will exit with an error if you run package scripts or try to install while your node runtime isn't a match for the engines config.

> However, if we do not pin the version, then people will inevitably write code that requires version 9, and then people running an environment with node version 8 will be unable to run.

This is possible today due to the way engines config works, and if our test automation strategies don't already account for this then we should work on that. 😄The docker updates in this PR should make testing multiple node versions pretty simple. [Travis build matrices](https://docs.travis-ci.com/user/build-matrix/) are awesome for this!

schuyler1d:
Another concern here with the yarn.lock file is diff/version churn.  For a while we had an npm lock file in the repo, and PRs got pretty insane -- the lock file would sometimes NOT be updated or too much would get updated (e.g. other deps would get changed too) -- and merges across PRs that both added dependencies were nightmares to resolve.

Perhaps it's worth trying again, but unfortunately our dependencies are far from stable, so will this lock file cause friction just at the wrong time (since ultimately it will make updating harder)?

IIRC, it also caused cross-platform issues -- because lock-file dependencies are different for e.g. Mac and Linux.  I suppose the argument is that if Mac folks or everyone always uses Docker, then that will be less of an issue?  But that means Docker is now a required development tool rather than just a (good/better) way to deploy Heroku.

At the very least, we would need to significantly update the README and CONTRIBUTING, etc on installing/running/debugging in Docker and also how/when to update the yarn.lock file.

Just to play with alternatives, would there be a way to have a separate repo for the Docker/Heroku deploy?  We haven't done that in the past with Heroku, but I'm wondering if it could/would make sense -- then it could independently sync with production on yarn.lock file updates -- we'd also be able to have several app.json files, perhaps....

jlegrone:
> will this lock file cause friction just at the wrong time (since ultimately it will make updating harder)?

Since [yarn 1.0](https://code.fb.com/developer-tools/announcing-yarn-1-0/) yarn automatically handles merge conflicts, which is a huge lifesaver. 😄 

> IIRC, it also caused cross-platform issues -- because lock-file dependencies are different for e.g. Mac and Linux.

Yarn understands [optional dependencies](https://yarnpkg.com/lang/en/docs/dependency-types/#toc-optionaldependencies) like fsevents and [will not error if they fail to install](https://travis-ci.org/MoveOnOrg/Spoke/builds/510384990#L506). The lockfile in this PR works equally well in linux and mac environments.

> At the very least, we would need to significantly update the README and CONTRIBUTING, etc on installing/running/debugging in Docker and also how/when to update the yarn.lock file.

I've made some updates to README to simplify the docker-compose instructions and use yarn commands. I don't think we have any documentation on how to add/modify project dependencies at the moment so I left that out.

`yarn.lock` will be automatically updated when dependencies change as a result of `yarn add` etc. I've also added a check to ensure that `yarn.lock` remains up to date by adding the `--frozen-lockfile` flag in travis.

> would there be a way to have a separate repo for the Docker/Heroku deploy?

I think it makes sense to keep deploy instructions and configuration as close to the source code as possible. Especially if things like #1056 are a goal.

schuyler1d:
Cool -- so the merge conflict resolution feature of Yarn makes it sound like we should at least give it a try.  I do think we want to document how Yarn can/should be used in the project.

Can you add a [small section here](https://github.com/MoveOnOrg/Spoke/blob/main/docs/EXPLANATION-development-guidelines.md) with what you would tell a first-time user of Yarn -- i.e. 
* If they add a dependency to package.json, what command do they run after that?
* If they, e.g. run `git merge origin/main` and a conflict comes up what they would run (a link to e.g. [a stackoverflow item](https://stackoverflow.com/questions/42939113/how-do-you-resolve-git-conflicts-in-yarn-lock) is fine)

schuyler1d (dismissed):
approving this PR.  Excited to see how yarn works with the Spoke project and make Docker/Heroku more standardized.

Also very excited about trying out Heroku review apps.

On merge to `main` @shakalee14 -- note that we should include the notice on how to migrate an existing Heroku app -- and probably enter specific dates in that documentation.

src/server/index.js
```diff
@@ -63,8 +63,10 @@ const port = process.env.DEV_APP_PORT || process.env.PORT
 
 // Don't rate limit heroku
 app.enable('trust proxy')
-if (!DEBUG && process.env.PUBLIC_DIR) {
-  app.use(express.static(process.env.PUBLIC_DIR, {
+
+// Serve static assets
+if (existsSync(process.env.ASSETS_DIR)) {
```
schuyler1d, line 68: is this an intentional change from PUBLIC_DIR => ASSETS_DIR?
Can you explain why it's ASSETS_DIR and not PUBLIC_DIR?

Also, why are we removing the DEBUG condition? I suppose some containerization deployments have simplicity benefits in serving static assets through node rather than a static file server? If so, I think we should at least have an env var that can disable this (no matter the presence of ASSETS_DIR).

(cc: https://github.com/MoveOnOrg/Spoke/pull/1085#pullrequestreview-222779938)

jlegrone: > is this an intentional change from PUBLIC_DIR => ASSETS_DIR?
Can you explain why it's ASSETS_DIR and not PUBLIC_DIR?

Yes, I noticed that `PUBLIC_DIR` only works if it contains `assets`, which is potentially very confusing (if you set `ASSETS_DIR` to something else during build then you'll have to do some digging to figure out why the app doesn't load).

I also could not figure out why specifying `PUBLIC_DIR` as the level above `assets` might be necessary (if you have insight here please let me know), and noticed that `ASSETS_DIR` is also required at runtime:

https://github.com/MoveOnOrg/Spoke/blob/9078ef59606e2608817bdb8a56a8fce5e4992256/src/server/middleware/app-renderer.jsx#L30

So by removing `PUBLIC_DIR`, we only have to set one environment variable and its usage is more declarative.

> Also, why are we removing the DEBUG condition?

I had no idea why we might want to turn off static asset serving when DEBUG is enabled. Is there a use-case for this? Unsetting the `ASSETS_DIR` or removing the `ASSETS_DIR` from path would also disable static file serving.

PR #1054, opened 2019-02-05
Fix deleting a texter

bchrobot:
Two fixes:

1. Calculate `unchangedMaxContacts` and `unchangedNeedsMessageCount` after confirming `texter` exists as these two values depend on `texter`.
2. If a database assignment's texter does not exist in `payload.texters`, then that texter must have been deleted. Handle this as a demotion (takes care of complete removal as well as partial removal due to having sent some texts).


schuyler1d (approved):
This looks good.

PR #1059, opened 2019-02-13
Local login AZ

azuzunaga:
Hi @schuyler1d! As discussed, I've implemented the following:

- Log in / sign up pages (looks a bit ugly still, but the functionality is there).
- Repurpose the `UserEdit` component so that it can be used for user log in / sign up.
- When signing up, user info is included in the request so that the passport `LocalStrategy` function can access the data and create the user.
- Added a `localauth` key to the password to be saved.
  - FYI, any local users in your dev env created before won't work anymore, since the index position has changed for the salt and password.
- A whole bunch of changes to make everything work together.

Next steps:

- [x] Incorporate feedback.
- [x] Logging in and Signing up should have their own logic - right now logins without an email address match automatically get added as new users
- [x] Allow users to change their password.
- [x] Allow admins to reset user's passwords.
- [x] Error message for existing users trying to sign up again.
- [x] Style log in / sign up page.
- ~~[ ] Refactor out auth functionality from `UserEdit`?~~
- [x] Pass through and validating org uuid and/or invite code

Let me know what you think and thanks for all your help with this one!


src/components/Login.jsx
```diff
@@ -1,31 +1,71 @@
...
+            Sign up
+          </button>
+        </section >
+
+        {!auth0Login &&
+          <UserEdit
+            authType={this.state.active}
+            saveLabel={this.state.active === 'login' ? 'Log in' : 'Sign up'}
+            router={router}
+            location={location}
```
schuyler1d: for `signup` state we need to test for and pass an invite code (either an invite table key or an org uuid).  If neither is present, then it should just show login without an option to signup.

(I'll comment further down, but we also need to validate this on the server and send through the UserEdit form)

src/components/Login.jsx
```diff
@@ -1,31 +1,71 @@
...
-      </div>
-    </form>
-    <UserEdit  />
-  </div>
-)
+        {/* Use auth0 */}
+        {auth0Login && window.AuthService.login(location.query.nextUrl)}
+
+        {/* Show UserEdit component configured for login / signup */}
+        <section>
```
schuyler1d: I think this part will also have to be hidden if auth0Login is enabled.  My PR was naively replacing it.

src/containers/UserEdit.jsx
```diff
@@ -23,40 +23,93 @@ class UserEdit extends React.Component {
   };
 
   async componentWillMount() {
-    await this.props.mutations.editUser(null)
+    if (!this.props.authType === 'login' && !this.props.authType === 'signup') {
```
schuyler1d: Maybe just `if (!this.props.authType)` is sufficient (will it be null in contexts where they are already logged in)?

src/containers/UserEdit.jsx
```diff
@@ -23,40 +23,93 @@ class UserEdit extends React.Component {
...
+      })
+      const { redirected, status, url } = res
+      if (redirected && status === 200) {
+        this.props.router.push(url)
+      } else if (!redirected && status === 401) {
+        throw new Error('Invalid username or password')
+      } else {
+        throw new Error()
+      }
+    } else if (!this.props.authType === 'login' && !this.props.authType === 'signup') {
```
schuyler1d: Some style suggestions on if....else block here.

Generally I prefer the shorter IF block first, so the if...else here should probably be flipped.  Second, is the `else if` necessary? -- i.e. can we just have 'else'?  So combined, I think I would restructure this as:
```
if (!this.props.authType) {
  ... // await this.props.mutations.editUser(formData)....
} else {
   const allData = {
   ....
}
```

src/server/auth-passport.js
```diff
@@ -74,37 +74,38 @@ export function setupAuth0Passport() {
...
-                            : done('Validation failed', null))
+          ? done(null, existingUser[0])
+          : done(null, false))
       )
     } else {
       // create the user
-      AuthHasher.hash(password, async function(err, hashed) {
-        const passwordToSave = `${hashed.salt}|${hashed.hash}`
-          //.salt and .hash
+      AuthHasher.hash(password, async function (err, hashed) {
```
schuyler1d: here we need to validate an invite code or org uuid that is passed in to make sure they have been invited to create an account first.

schuyler1d: I think we also want to validate the passwords are the same here.  Just because we validate on the client, we also have to validate on the server.

src/server/auth-passport.js
```diff
@@ -119,9 +120,8 @@ export function setupLocalAuthPassport() {
 
   return {
     loginCallback: [
-      passport.authenticate('local', { failureRedirect: '/login' }),
+      passport.authenticate('local'),
```
schuyler1d, line 116: Can you explain why we remove the failureRedirect path? Not in code comments, necessarily, but to me?

If it's because we 'cannot fail' above because we either create a user or login, then that won't be the case after testing for the conditions necessary for account creation, and then we'll need it again.

schuyler1d (commented):
@azuzunaga This looks great so far!  Added some comments -- some style nits, but mostly just adding a TODO item of passing through and validating org uuid and/or invite code

schuyler1d:
Also a comment on your todo list -- I would remove 'allow user to reset password' -- I think it's better for now just to allow admins to reset passwords if needed -- assuming a separate channel for validating the texter.

src/server/auth-passport.js
azuzunaga: I removed it because the server response was coming back with a `302 Found` or `500 Internal Server Error` status code (depending on how verification is handled here: https://github.com/MoveOnOrg/Spoke/pull/1059/files#diff-5f814da17f81ded4f5bf5d984739c0b0R92). Without redirecting, Passport by default will respond with a `401 Unauthorized` status if authentication fails, which to me made more sense.

src/server/auth-passport.js
azuzunaga: > here we need to validate an invite code or org uuid that is passed in to make sure they have been invited to create an account first.

Do we need to also do this when using Auth0? Right now, when a texter signs up using either local login or Auth0 and there isn't an invite code, the following message is displayed:
image

src/components/Login.jsx
azuzunaga: Should I also do the opposite, only show signup if there is an invite code?

src/components/Login.jsx
azuzunaga: Yup! The conditional check on line 53 should cover this whole section. Good catch!

src/components/Login.jsx
schuyler1d: No, an invite is also to a new campaign eg for dyn assignment so often invites/uuid links will be clicked by existing users too

src/server/auth-passport.js
schuyler1d: If we can then yes, that would actually be good for security. But if it's tricky then you don't need to do it for this PR

src/server/auth-passport.js
schuyler1d: Hmm. I think we do want it to redirect to login if they are unauthorized. If that's not the way to do it then I think we need to figure out how. Maybe I setup the options incorrectly?

src/containers/UserEdit.jsx
```diff
@@ -78,7 +77,7 @@ class UserEdit extends React.Component {
     }
 
     let userFields = {}
-    if (!this.props.authType === 'login' || this.props.authType === 'signup') {
+    if (!this.props.authType || this.props.authType === 'signup') {
```
schuyler1d: This seems more easily expressed as `this.props.authType !== 'login'` (and below

azuzunaga: Duh!

src/server/auth-passport.js
azuzunaga: Hey @schuyler1d, from what I understand, since we are initiating the `POST` request via JS, the browser won't actually redirect, even if the response from the server comes with a redirect request. So right now, the user won't actually leave the login page until their authentication request is successful.

Is that your concern, or are you thinking of something else? If you'd like, we can chat about it so I can better understand where you are coming from.

Thanks!

azuzunaga:
Update with pending tasks:
- [x] Allow users to change their password.
- [x] Allow admins to reset user's passwords.

src/server/local-auth-helpers.js
```diff
@@ -0,0 +1,185 @@
...
+      const updatedUser = await User
+        .get(existingUser[0].id)
+        .update({ auth0_id: passwordToSave })
+        .run()
+      resolve([null, updatedUser])
+    })
+  })
+}
+
+const change = ({
```
schuyler1d: so this is the one endpoint that I think we should do through graphQL (or maybe I don't see it being done that way, but it is?  changing a password should require an authenticated person and we should also validate that the logged in user is the same as the one changing the password).

Otherwise, this endpoint is overly vulnerable

schuyler1d (changes requested):
This is fantastic work @azuzunaga !

All of this is really elegant.  I have a few security flags that I believe should be pretty minor to fix, but lmk if something is overly burdensome and we'll figure out a way to merge first and fix later.

This is going to make local development infinitely easier.  wow!

schuyler1d:
Oh!  Also merge `main` and make this PR against `main` instead of local-login.

src/server/local-auth-helpers.js
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  // Get user resetHash and date of hash creation
+  const pwFieldSplit = existingUser[0].auth0_id.split('|')
+  const [resetHash, datetime] = [pwFieldSplit[1], pwFieldSplit[2]]
+
+  // Verify hash was created in the last 15 mins
+  const isExpired = (Date.now() - datetime) / 1000 / 60 > 15
+  if (isExpired) {
+    return [null, false, errorMessages.resetHashExpired]
```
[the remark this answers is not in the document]
azuzunaga: Sounds good!

src/server/local-auth-helpers.js
azuzunaga: OK, makes sense. I'll change this up.

azuzunaga:
Thanks @schuyler1d!! I'll work on your comments over the next week.

src/containers/UserEdit.jsx
```diff
@@ -5,91 +5,285 @@ import wrapMutations from './hoc/wrap-mutations'
...
   async handleSave(formData) {
-    const result = await this.props.mutations.editUser(formData)
-    if (this.props.onRequestClose) {
-      this.props.onRequestClose()
+    if (!this.props.authType) {
+      await this.props.mutations.editUser(formData)
+      if (this.props.onRequestClose) {
+        this.props.onRequestClose()
+      }
+    } if (this.props.authType === 'change') {
```
schuyler1d: is this an `else if`?  otherwise it should be a newline here.  It wouldn't matter for this condition-block, but the `else` block below will e.g. trigger for all non-authType=change cases.

src/server/local-auth-helpers.js
```diff
@@ -0,0 +1,187 @@
...
+  existingUser,
+  nextUrl,
+  uuidMatch
+}) => {
+  if (existingUser.length === 0) {
+    return [null, false, errorMessages.invalidCredentials]
+  }
+
+  // Verify UUID validity
+  if (nextUrl) validUuid(nextUrl, uuidMatch)
```
schuyler1d: When do we validate UUID validity for login?  Do we need it here?
If so, we need an `await` here -- and also need to test whether it succeeded -- and also let's document when we need to validate it here.

src/server/local-auth-helpers.js
```diff
@@ -0,0 +1,187 @@
...
+const signup = async ({
+  lowerCaseEmail,
+  password,
+  existingUser,
+  nextUrl,
+  uuidMatch,
+  reqBody
+}) => {
+  // Verify UUID validity
+  if (nextUrl) validUuid(nextUrl, uuidMatch)
```
schuyler1d: Remember nextUrl is user input controlled -- I think for signup we *always* want to validate a Uuid, right?  so if there's no nextUrl, we can/should fail-fast -- but that might be better done in validUuid.

Also, we need an `await` here -- and also need to test whether it succeeded and maybe send back the error?  (based on the template below)

src/server/local-auth-helpers.js
```diff
@@ -0,0 +1,187 @@
...
+  invalidCredentials: 'Invalid username or password',
+  emailTaken: 'That email is already taken.',
+  passwordsDontMatch: 'Passwords don\'t match.',
+  invalidResetHash: 'Invalid username or password reset link. Contact your administrator.',
+  noSamePassword: 'Old and new password can\'t be the same'
+}
+
+const validUuid = async (nextUrl, uuidMatch) => {
+  let foundUUID
+  if (nextUrl.includes('join')) {
```
schuyler1d, line 18: * this needs to fail fast if `!uuidMatch`  -- if someone submits a request without a validUuid at all, we don't even need to do a search.  
* Also let's return something in the non-error case so we can test affirmatively for that below.

schuyler1d (changes requested):
Really really really close!

src/containers/UserEdit.jsx
azuzunaga: Good catch! It should be an `else if`.

src/server/local-auth-helpers.js
azuzunaga: Good catch. UUID's are validated on login in case it is an invite to an organization or a new campaign.

src/server/local-auth-helpers.js
azuzunaga: Instead of returning an array that can be used by the `done` callback, I am now throwing an error that is handled in the `LocalStrategy` constructor. I explain why in the commit body of my last commit. Curious to hear your opinion, and happy to change it back.

src/server/local-auth-helpers.js
```diff
@@ -30,11 +30,13 @@ const login = async ({
...
-    return [null, false, errorMessages.invalidCredentials]
+    throw new Error(errorMessages.invalidCredentials)
   }
 
-  // Verify UUID validity
-  if (nextUrl) validUuid(nextUrl, uuidMatch)
+  // Verify UUID validity when an existing user is invited to a new
+  // campaign, e.g. dynamic assignament.
+  // If there is an error, it will be caught on local strategy invocation
+  if (nextUrl) await validUuid(nextUrl, uuidMatch)
```
schuyler1d: I like the new structure with throwing errors -- it's much easier to read and understand.

I'm still not clear if we actually need to validate this on login -- I know there will *be* a nextUrl with a uuid that we could test, but after login, isn't there another GraphQL api call to 'join' the org (in src/containers/JoinTeam.jsx)?  I suppose it doesn't hurt here, but I think it might make the logic a bit more confusing.

schuyler1d (approved):
:100: 
:fireworks:

Issue #1046, opened 2019-02-01
add user_id for campaign creator to campaigns table


joemcl:
add user_id for campaign creator to campaigns table, require that, and display user_id, user name and created at date and time in campaign info in admin Campaigns in dashboard - related to/similar to #809 #906

benhiller:
I'll take a look into implementing this.

PR #1063, opened 2019-02-25
Add creator_id to campaign

benhiller:
Summary: This adds a column to the campaign table to track which user
created the campaign, and displays that information on the admin
campaign list, as requested in #1046.

I modeled the migration off of migration #11, which added a similar
column to the message table.

I updated the two GraphQL mutations that can create campaigns,
`createCampaign` and `copyCampaign` to store the ID of the logged in
user as creator_id.

Finally, I added a `creator` field to `Campaign` in GraphQL and used
that in `CampaignList` to show which user created the campaign. Let me
know if you have feedback on how I added it to the UI, I added it to the
same line as the description since I didn't want to increase the height
of each campaign row, but I can make it it's own line if you'd prefer
that instead. I see the github issue also mentioned showing the raw ID,
and the time the campaign was created. I think I'll check with joemcl
first to make sure what I added is about what he was looking for, and
then look into adding those pieces of information.

I did notice that when adding the `user_id` field to the `message` schema,
the same commit also updated the `dependencies` we pass as options to
include `User` (this line: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/models/message.js#L36).
I looked at `createModel` in `rethink-knex-adapter`, and it
looks like `dependencies` isn't used at all, so I didn't update that in
this commit, but let me know if I missed something and I should update
it!

Test Plan: Confirmed that a campaign with a null creator_id still
rendered properly, and that a campaign with a non-null creator_id showed
the correct name on the campaign list. Tested copying a campaign with no
creator_id to confirm the copied campaign had the logged-in user's id as
creator_id.

### DIFF
--- a/src/server/api/campaign.js
+++ b/src/server/api/campaign.js
@@ -235,6 +235,11 @@ export const resolvers = {
         return currentEditors(r.redis, campaign, user)
       }
       return ''
-    }
+    },
+    creator: async (campaign, _, { loaders }) => (
+      campaign.creator_id
+      ? loaders.user.load(campaign.creator_id)
+      : null
+    )
   }
 }
--- a/__test__/containers/CampaignList.test.js
+++ b/__test__/containers/CampaignList.test.js
@@ -1,0 +1,63 @@
+/**
+ * @jest-environment jsdom
+ */
+import React from 'react'
+import { mount } from 'enzyme'
+import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider'
+import { CampaignList } from '../../src/containers/CampaignList'
+
+describe('Campaign list for campaign with null creator', () => {
+  // given
+  const campaignWithoutCreator = {
+    id: 1,
+    title: 'Yes on A',
+    creator: null,
+  }
+
+  const data = {
+    organization: {
+      campaigns: {
+        campaigns: [ campaignWithoutCreator ],
+      },
+    },
+  }
+
+  // when
+  test('Renders for campaign with null creator, doesn\'t include created by', () => {
+    const wrapper = mount(
+      <MuiThemeProvider>
+        <CampaignList data={data} />
+      </MuiThemeProvider>
+    )
+    expect(wrapper.text().includes('Created by')).toBeFalsy()
+  })
+})
+
+describe('Campaign list for campaign with creator', () => {
+  // given
+  const campaignWithCreator = {
+    id: 1,
+    creator: {
+      displayName: 'Lorem Ipsum'
+    },
+  }
+
+  const data = {
+    organization: {
+      campaigns: {
+        campaigns: [ campaignWithCreator ],
+      },
+    },
+  }
+
+  // when
+  test('Renders for campaign with creator, includes created by', () => {
+    const wrapper = mount(
+      <MuiThemeProvider>
+        <CampaignList data={data} />
+      </MuiThemeProvider>
+    )
+    expect(wrapper.containsMatchingElement(<span> &mdash; Created by Lorem Ipsum</span>)).toBeTruthy()
+  })
+})
+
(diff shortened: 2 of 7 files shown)


joemcl:
Thanks @benhiller though since a campaign must be created by a logged in user or via API call, think we want campaign creator user id to be mandatory and not allow a null value

joemcl:
Same if a user clones an existing campaign that does not have a creator user id - add user id of cloned campaign creator on cloning

harpojaeger:
Nice work @benhiller! Can you add Enzyme tests for the cases you describe in the test plan? It should be fairly easy to render the `CampaignList` with fake data and assert that the expected UI elements are present when they should be and not when they're not.

benhiller:
Thanks for taking a look!

@joemcl - I made `creator_id` nullable since it seemed possible that an existing Spoke instance with campaigns already created could upgrade their version of Spoke to this version, and then existing campaigns would have a null `creator_id`. Let me know if this was a faulty assumption. I do handle the cloning case already (assuming that is referring to the same operation as copying a campaign).

@harpojaeger - Sure, will take a look into that!

Also, this is a screenshot of the change, which I forgot to include before:
screen shot 2019-02-25 at 9 32 57 am

joemcl:
@benhiller thanks so much again for tackling this. Null value for creator user id is cool for past campaigns of course - as long as new or cloned campaigns on creation will automatically add creator user id and the creator can't not include that. Re the screenshot "Created by" display - besides creator user name, can you add user ID, day, date and time (in instance local TZ) created so it reads "Created by Ben Hiller (Spoke user ID XXX) on Monday Feb 25 at 9:55am US Eastern Time" for example. Bonus points if you can add a way to *filter and /or sort* by all of those fields in the Campaigns admin UI too :) but that's not in my filed Issue.

joemcl:
cc: @deasterdaywfp @suegibson ^^^^  - thoughts?

harpojaeger:
Why do we want to display the user ID?

joemcl:
@harpojaeger so we can find the correct user quickly if necessary and the user name is incomplete or duplicative. It's only in the Campaigns admin UI anyway. We have a lot of admin users.

shakalee14 (approved):

src/server/api/campaign.js
```diff
@@ -235,6 +235,11 @@ export const resolvers = {
         return currentEditors(r.redis, campaign, user)
       }
       return ''
-    }
+    },
+    creator: async (campaign, _, { loaders }) => (
+      campaign.creator_id
```
schuyler1d, line 240: this is fine for now, but can you just add a one line comment that says:
`TODO: optimization to include these fields with organizations.campaigns query`
If you want to stub a bit of that out now it would
1. be `campaign.creator || (campaign.creator_id && loaders.user.load(campaign.creator_id)) || null` here
2. if server/api/organization.js in campaigns you'd also join on user and get creator name.

but the TODO line is fine for now.

schuyler1d (approved):
:+1: yei, tracking!

joemcl:
just seeing this and yay! thanks @benhiller

joemcl:
@benhiller this is working as expected, thanks so so much, on our staging site running current stage-main. For the next iteration I'd like it to show this for example -  "Created by Ben Hiller (Spoke user ID XXX) on Monday Feb 25 at 9:55am US Eastern Time" but that's for another time.

benhiller:
@joemcl - awesome, great to hear it is working on the staging site! I did start to look into adding the date & time as well, but ran into some odd issues with timezones. I can definitely send out another PR to add user ID to this line, which should be much less trouble.

Thanks for the review @shakalee14 and @schuyler1d! I'll also look into implementing the optimization you mentioned when I get a chance.

deasterdaywfp:
This is exciting, thank you Ben!

PR #1067, opened 2019-02-27
WIP: Fix for Dynamic Assignment Bug

### LINKED ISSUE #1066: Dynamic assignment is broken
We updated to latest upstream branch main last week. Previously, with Dynamic on, we could post a link and folks could just jump in and start sending. Now, each person has to be assigned a batch and/or max count has to be updated manually from 0 to a number. cc: @deasterdaywfp

shakalee14:
WIP 
fixes #1066

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -454,7 +454,7 @@ const rootMutations = {
         await Assignment.save({
           user_id: user.id,
           campaign_id: campaign.id,
-          max_contacts: parseInt(process.env.MAX_CONTACTS_PER_TEXTER || 0, 10)
+          max_contacts: (process.env.MAX_CONTACTS_PER_TEXTER ? parseInt(process.env.MAX_CONTACTS_PER_TEXTER, 10) : null )
         })
       }
       return campaign


deasterdaywfp:
@joemcl @codygordon @lperson What can we do to address these errors upon check? Can we still address and merge the fix to run dynamic in the latest build again?

Also, thank you 1000% @shakalee14 !

bdatkins (dismissed):

deasterdaywfp:
Thank you @bdatkins !

deasterdaywfp (commented):
@bdatkins is there anything that I need to do to approve changes? Shall we attempt the merge again and test?

src/components/AssignmentSummary.jsx
```diff
@@ -117,7 +117,7 @@ export class AssignmentSummary extends Component {
               title: 'Send first texts',
               count: unmessagedCount,
               primary: true,
-              disabled: (useDynamicAssignment && !hasUnassignedContacts && unmessagedCount == 0) || (useDynamicAssignment && maxContacts === 0),
+              disabled: (useDynamicAssignment && !hasUnassignedContacts && unmessagedCount == 0 && maxContacts !== 0) ? true : false,
```
harpojaeger: It might make sense to move this boolean logic to L14, before `this.renderBadgedButton` is called, so it can be described more clearly. Also, is the `? true : false` necessary here?

joemcl:
Hey @deasterdaywfp - we can test this branch on our staging site to see if it works. But merging into upstream branch main requires @schuyler1d or Ann ( @mathemagica ) with MoveOn to review the code, it seems.

deasterdaywfp:
Right thank you Joe, makes perfect sense. Thank you for that clarification.

schuyler1d (dismissed):
:+1:

src/components/AssignmentSummary.jsx
schuyler1d: agree.  I also think the `&& maxContacts != 0` is maybe incorrect.  If maxContacts is not zero, it's still possible that it's disabled, right?  The logic might be a bit weirder....

shakalee14: thanks for the feedback. i think this part doesnt need to be changed after doing further qa

schuyler1d (approved):
:+1:

PR #1072, opened 2019-03-14
Make empty screen messages more accessible

azuzunaga:
Resolves MoveOnOrg#991.

Will also affect all the following messages, which use the `<Empty />` component on a white background:
- https://github.com/MoveOnOrg/Spoke/blob/main/src/components/AssignmentTexter.jsx#L302
- https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/AdminOptOutList.jsx#L16
- https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/AdminPersonList.jsx#L105
- https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/AssignmentTexterContact.jsx#L577
- https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/CampaignList.jsx#L140

### DIFF
--- a/src/components/Empty.jsx
+++ b/src/components/Empty.jsx
@@ -32,7 +32,6 @@ const styles = StyleSheet.create({
   },
   title: {
     ...theme.text.header,
-    opacity: 0.2,
     textAlign: 'center'
   },
   content: {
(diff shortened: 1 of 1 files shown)


shakalee14 (approved):

schuyler1d (approved):
yei!

Issue #974, opened 2018-10-10
Texters order


suegibson:
Can texters please be listed in alpha order regardless of whether they have capitalized the first letter in their name? Thank you!
@joemcl 
@lperson 
![a156c1a9-4bf0-47ab-83a7-ad8e886e4696](https://user-images.githubusercontent.com/39012138/46732266-a86e8500-cc52-11e8-8bf2-e8ecebd111c6.jpeg)

joemcl:
Hey @lperson noticing that this is also an issue in People - so how about just making sorting by name in People and Texters case-insensitive ? cc: @azuzunaga

lperson:
@joemcl when you say it's an issue in People do you mean the people page?

lperson:
Because the people page in WFP staging is sort case insensitively.

joemcl:
> Because the people page in WFP staging is sort case insensitively.

@lperson yes in People and I just tried it - I changed my first name from Joe to joe and joe appears first in sort , changed back to Joe and it sorts correctly.

lperson:
@joemcl the people page in your staging right now is doing a correct case insensitive sort.  Look at Joyce, Judith and Julie.

<img width="42" alt="Screen Shot 2019-06-28 at 12 55 47 AM" src="https://user-images.githubusercontent.com/1637288/60318055-8b442880-993f-11e9-9d47-e25de984777d.png">

joemcl:
@lperson just changed my first name on our staging from Joe to joe. Sorted by first name. Now my name with lowercase joe and upper case McL is sorted before Joe D. 

![38bd71e0-88af-4d11-b674-46e4af4514c7](https://user-images.githubusercontent.com/3912434/60318325-6f348d00-9927-11e9-93e4-2e793f68618a.jpg)

joemcl:
I changed it back to Joe and now it sorts correctly - Joe McL comes after Joe D.

lperson:
@joemcl  Oy gevault.  I'll fix the people page but let's keep the issue open for texter order in campaign edit.

lperson:
@joemcl https://github.com/WorkingFamilies/Spoke/pull/110 takes care of the people page problems.  To reiterate I did not address the sort of testers in the assignment tool at this time.

lperson:
https://github.com/WorkingFamilies/Spoke/pull/114

ibrand:
[...]

PR #1074, opened 2019-03-14
948 always capitalize first letter

### LINKED ISSUE #948: MoveOn Spoke: Create fix to ensure first letters of texter first & last name are always capitalized
Currently, if a texter creates a Spoke account like this "erik kirkstein" instead of like this "Erik Kirkstein", the texter's name does not appear in alphabetical order in the texter menu of a campaign, but instead at the bottom of the list.

it would be ideal if regardless of how a texter enters their name during the Spoke account setup process, that their name always appears in Spoke with the first letter of the first and last name capitalized.

azuzunaga:
Resolves MoveOnOrg/Spoke#948 and MoveOnOrg/Spoke#974.

This fix does not apply retroactively, only to campaigns created after the fix is deployed. I've also sorted the users in the "People" menu by first name. Let me know if both are OK.

An alternative would be to capitalize user's first and last name on account creation, and then create a migration to change the existing data in the DB. Let me know if this would be a better approach.

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -7,6 +7,7 @@ import { organizationCache } from '../models/cacheable_queries/organization'
 
 import { gzip, log, makeTree } from '../../lib'
 import { applyScript } from '../../lib/scripts'
+import { capitalizeWord } from './lib/utils'
 import { assignTexters, exportCampaign, importScript, loadContactsFromDataWarehouse, uploadContacts } from '../../workers/jobs'
 import {
   Assignment,
@@ -380,23 +381,23 @@ const rootMutations = {
         return null
       } else {
         const member = userRes[0]
+
+        const newUserData = {
+          first_name: capitalizeWord(userData.firstName),
+          last_name: capitalizeWord(userData.lastName),
+          email: userData.email,
+          cell: userData.cell
+        }
+
         if (userData) {
           const userRes = await r
             .knex('user')
             .where('id', userId)
-            .update({
-              first_name: userData.firstName,
-              last_name: userData.lastName,
-              email: userData.email,
-              cell: userData.cell
-            })
+            .update(newUserData)
           await cacheableData.user.clearUser(member.id, member.auth0_id)
           userData = {
             id: userId,
-            first_name: userData.firstName,
-            last_name: userData.lastName,
-            email: userData.email,
-            cell: userData.cell
+            ...newUserData
           }
         } else {
           userData = member
@@ -1034,8 +1035,8 @@ const rootMutations = {
 
       log.info(
         `Sending (${service}): ${messageInstance.user_number} -> ${
-          messageInstance.contact_number
-          }\nMessage: ${messageInstance.text}`
+        messageInstance.contact_number
+        }\nMessage: ${messageInstance.text}`
       )
 
       service.sendMessage(messageInstance, contact)


src/server/api/lib/utils.js
```diff
@@ -15,3 +15,7 @@ export function mapFieldsToModel(fields, model) {
   })
   return resolvers
 }
+
+export function capitalize(word) {
+  return word[0].toUpperCase() + word.slice(1)
```
schuyler1d: this will fail on empty fields, so you'll need to guard against that.

src/server/api/organization.js
```diff
@@ -28,7 +28,19 @@ export const resolvers = {
     },
     people: async (organization, { role, campaignId }, { user }) => {
       await accessRequired(user, organization.id, 'SUPERVOLUNTEER')
-      return buildUserOrganizationQuery(r.knex.select('user.*'), organization.id, role, campaignId)
+      let users = await buildUserOrganizationQuery(r.knex.select('user.*'), organization.id, role, campaignId)
```
schuyler1d: I think we should separate this from the captilize PR.  If we were going to do this, I think it would be better to change sorting in `buildUserOrganizationQuery` and make it `orderBy('first_name DESC')`  otherwise it could mess up pagination (if/when we have it)

schuyler1d (changes requested):
I think I would rather just uppercase names on initial user save and then we don't have to muck in all these other places.  If we do, then it should just be `displayName` and then we can rely on that to show whatever we want correctly.

src/server/api/lib/utils.js
azuzunaga: ++

src/server/api/organization.js
azuzunaga: ++

azuzunaga:
> I think I would rather just uppercase names on initial user save and then we don't have to muck in all these other places. If we do, then it should just be `displayName` and then we can rely on that to show whatever we want correctly.

Sounds good. I'll update the user creation process and write a migration to change existing records.

joemcl:
hey @azuzunaga how can we move this along? thanks.

azuzunaga:
@joemcl, users' first and last names are already being capitalized on creation. This PR adds a migration to capitalize existing users' first and last names, and works with sqlite or postgres.

To move this along we need a PR review.

Note: Newest changes were force pushed because branch was very outdated and contained duplicated/reverted commits.

azuzunaga:
@joemcl, I saw your comments on #974 and added a change so that users' first and last name are also capitalized when a user changes their name. Since we already do this on account creation I don't think it's too big of an issue.

joemcl:
@azuzunaga thanks did not realize new users first and last now capitalized on creation....

ibrand:
@azuzunaga and @schuyler1d what's the status of this PR? Were the requested changes made?

azuzunaga:
hey @ibrand! I just fixed the latest merge conflict. This PR is ready for review.

ibrand:
@schuyler1d Could you give this a final review? Looks like you had requested changes initially and I want to make sure we're all good to go. Thanks everyone!

src/migrations/index.js
```diff
@@ -195,6 +195,17 @@ const migrations = [
         console.log(`set sequence canned_response_id_seq to ${maxId + 1}`)
       }
     }
+  },
```
schuyler1d: we'll need to remove this migration with #1154 (maybe just make it a separate PR, or don't actually change old campaigns, and skip it)

schuyler1d (dismissed):
modulo removing the migration :+1:

src/migrations/index.js
azuzunaga: Thanks @schuyler1d, I'll remove the migration later today and push up the changes.

@ibrand I'll leave it up to you to decide whether we add the migration as a separate PR or if we just skip it. Thanks!

azuzunaga:
Migration has been removed.

ibrand:
@azuzunaga Awesome. I'll follow up once #1154 is merged since we're switching to new types of migrations so you can add a follow-up migration backfill PR. Thank you for your work on this!

azuzunaga:
Thank you, @ibrand. And sounds good!

shakalee14:
Merged with https://github.com/MoveOnOrg/Spoke/commit/2032a9b1004b41a4622896df9b0bb98ee7b136ed

PR #1051, opened 2019-02-03
Message Review: clear selections after reassigning texters

lperson:
Closes #1023

PR #1078, opened 2019-03-27
Stage main

shakalee14:
@jlegrone:
#1053 

@schuyler1d:
#1077 

@benhiller:
#1063 

@lperson:
#1051 

Migrations:
- creator_id to campaign table apart of #1063


schuyler1d (dismissed):
:+1:

PR #1085, opened 2019-04-04
Update lambda and local development docs to node 8 runtime

jlegrone:
Node 6 [reaches end-of-life status](https://nodejs.org/en/about/releases/) as of April 30th 2019.

This PR is based on #1053, which switches the Heroku/docker runtime to node 8.


schuyler1d:
<strike>Should this trigger a 2.0 release?  E.g. with Lambda, merging this would be incompatible with updating an existing deploy.
</strike>

Belay that -- after seeing that the upgrade path on most platforms is seemless, this is not as big a deal as I was worried about.

package.json
```diff
@@ -159,13 +160,12 @@
     "eslint-plugin-react": "^5.2.2",
     "fakeredis": "^2.0.0",
     "foreman": "^1.4.1",
-    "jest": "^22.1.4",
```
schuyler1d, line 162: don't we still need jest for the e2e tests?  Lines 11-15 still call jest.  Happy for them to move too -- but also make sure the docs don't mention jest

shakalee14: our test scripts use `jest`

jlegrone:
> Should this trigger a 2.0 release? E.g. with Lambda, merging this would be incompatible with updating an existing deploy.

According to this [blog post](https://claudiajs.com/news/2018/04/09/claudia-4.html), updating the runtime for existing functions is possible with `claudia update --runtime nodejs8.10`.

Full update command docs available here: https://github.com/claudiajs/claudia/blob/master/docs/update.md

src/server/index.js
```diff
@@ -63,8 +63,10 @@ const port = process.env.DEV_APP_PORT || process.env.PORT
 
 // Don't rate limit heroku
 app.enable('trust proxy')
-if (!DEBUG && process.env.PUBLIC_DIR) {
-  app.use(express.static(process.env.PUBLIC_DIR, {
+
+// Serve static assets
+if (existsSync(process.env.ASSETS_DIR)) {
```
schuyler1d, line 68: is this an intentional change from PUBLIC_DIR => ASSETS_DIR?
Can you explain why it's ASSETS_DIR and not PUBLIC_DIR?

Also, why are we removing the DEBUG condition?  I suppose some containerization deployments have simplicity benefits in serving static assets through node rather than a static file server?  If so, I think we should at least have an env var that can disable this (no matter the presence of ASSETS_DIR).

jlegrone: Also part of the heroku PR. 😄Can we move this discussion over to #1053?

package.json
schuyler1d: ok, sounds good -- can we make a comment for that, at least.  I'm ++ for a downstream dependency, but just to avoid the perception of magic, maybe add a comment above the jest script lines?

Is it still impossible to add comments to package.json?

jlegrone: Still impossible... but for posterity here is some more information.

This is the documentation for create-react-app testing, which is what we're using:

https://facebook.github.io/create-react-app/docs/running-tests

The front page of [create-react-app](https://facebook.github.io/create-react-app/) says:

> Your app only needs one build dependency. We test Create React App to make sure that all of its underlying pieces work together seamlessly – no complicated version mismatches.

So to get the benefits of cra, we want to stay out of the business of manually managing dependencies that it already declares, and only [bump the react-scripts version](https://facebook.github.io/create-react-app/docs/updating-to-new-releases) when it's time to upgrade.

PR #1095, opened 2019-04-25
add texter tests and autorespond fakeservice feature

schuyler1d:
@jparkrr  added texter tests last September.  I've ported that work to current main


joemcl:
what does "texter tests" do?

PR #1096, opened 2019-04-25
add permissions for campaign queries 

schuyler1d:
and add a shortcut for texters with campaign data.

closes up a security hole (this is ported from scaling branches)

### DIFF
--- a/src/server/api/campaign.js
+++ b/src/server/api/campaign.js
@@ -1,3 +1,4 @@
+import { accessRequired } from './errors'
 import { mapFieldsToModel } from './lib/utils'
 import { Campaign, JobRequest, r, cacheableData } from '../models'
 import { currentEditors } from '../models/cacheable_queries'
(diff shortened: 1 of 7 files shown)


src/server/api/campaign.js
```diff
@@ -230,9 +250,11 @@ export const resolvers = {
       || cacheableData.campaign.dbCustomFields(campaign.id)
     ),
     stats: async (campaign) => campaign,
+    cacheable: (campaign, _, { user }) => Boolean(r.redis),
     editors: async (campaign, _, { user }) => {
+      await accessRequired(user, campaign.organization_id, 'SUPERVOLUNTEER', true)
       if (r.redis) {
-        return currentEditors(r.redis, campaign, user)
+        return cacheableData.campaign.currentEditors(campaign, user)
```
shakalee14, line 257: currentEditors currently isnt a query attached to the campaign cache and is apart of the user cache

shakalee14 (changes requested):
I'm getting the following error here: http://localhost:3000/app/1/todos

`Object { status: 500, message: "Cannot query field \"cacheable\" on type \"User\"." }`
because `this.props.data.currentUser` is currently undefined. I'm also getting an error associated with `currentEditors` which comes up on the admin campaign edit page.

schuyler1d:
fixed. (I queried user cacheable, but added campaign's cacheable property -- both will eventually be added, but let's keep this one simple)

PR #1104, opened 2019-05-01
MESSAGE REVIEW doc draft

joemcl:
A bare bones outline/draft for a Message Review doc, with one tip/trick added to start.


shakalee14:
great idea - a how to guide would be super helpful for us!

joemcl:
Yeah def, and @lperson I'm just getting started on this, FYI.

Issue #1106, opened 2019-05-12
Canned Responses - all disappearing after one or multiple saved


joemcl:
So for awhile (months, at least) we have had a problem with all Canned Responses disappearing from some (not all) Campaigns after one or more Canned are created in a new Campaign and saved. It happened again earlier today, and I just recreated it. 
I cloned a campaign from earlier today that had Interactions save successfully and was having Canned added, which then disappeared.
I started adding all the same Canned. I was able to add all from the script, saving periodically, and when I finished adding all and Saved one last time, I refreshed Canned and all the Canned Responses had disappeared. 
I'm attaching an excerpt of the log from the time of the error. It seems a SQL error is being created - ```sql_error_code = 23505 ERROR:  duplicate key value violates unique constraint "canned_response_pkey" ``` . 

cc: @lperson @deasterdaywfp @codygordon  . 

[20190511-639PMPacificSpokeCannedResponseVanishingErrorLog.txt](https://github.com/MoveOnOrg/Spoke/files/3169638/20190511-639PMPacificSpokeCannedResponseVanishingErrorLog.txt)

deasterdaywfp:
Thank you @joemcl this is a really big argument for the Google doc integration to load canned responses IMHO.

joemcl:
@shakalee14 I'm wondering if it is something related to the random Canned IDs created via this - 
https://github.com/MoveOnOrg/Spoke/pull/122/commits/7e51f58338efe44a018e21eb4f1cf123846822ed  
 maybe they are not that random?

joemcl:
> Thank you @joemcl this is a really big argument for the Google doc integration to load canned responses IMHO.

@deasterdaywfp yeah you mean #1102 - I think the 'insertion' method - from a Google Doc somehow, or from our current copying and pasting text from a Google Doc into the Canned Response UI - would not make a difference, something else is going on.

joemcl:
So @shakalee14 see screenshot below - the Canned response ID created randomly in the new Campaign already existed as an ID for a Canned in an earlier Campaign. So rather than Canned response IDs being generated randomly - with apparently random dupes appearing - why can't Canned response IDs just be numbered sequentially on creation, i.e., autoincrement, to insure a unique value for each? 

![3884b1c7-7987-4303-a108-9b0624e2d795 (1)](https://user-images.githubusercontent.com/3912434/57577420-0e123200-742b-11e9-996c-4097ade2fc25.jpg)

PR #1107, opened 2019-05-13
Use the sequence for canned_response ids

lperson:
`canned_response` id is being assigned a random number.  This is suspected as the cause of a problem WFP is experiencing which presents as "all the canned responses I just created disappear."  It's plausible that if we attempt to save a `canned_response` with a duplicate ID, the update fails, the new canned_responses are not saved, and it appears to the user as if they disappeared.

This PR sets `canned_response_id_seq` to `max(id)+1` in a migration, then sets ids of new canned responses using the sequence.

### DIFF
--- a/src/migrations/index.js
+++ b/src/migrations/index.js
@@ -188,9 +188,9 @@ const migrations = [
     date: '2019-05-13',
     migrate: async () => {
       console.log('setting sequence value for canned_response')
-      const maxId = await r.knex('canned_response').max('id').first()
-      await r.knex.raw(`ALTER SEQUENCE canned_response_id_seq RESTART WITH ${maxId.max + 1}`)
-      console.log(`set sequence canned_response_id_seq to ${maxId.max + 1}`)
+      const maxId = (await r.knex('canned_response').max('id').first()).max || 0
+      await r.knex.raw(`ALTER SEQUENCE canned_response_id_seq RESTART WITH ${maxId + 1}`)
+      console.log(`set sequence canned_response_id_seq to ${maxId + 1}`)
     }
   }
 
--- a/src/migrations/index.js
+++ b/src/migrations/index.js
@@ -1,4 +1,4 @@
-import { r, Migrations } from '../server/models'
+import { r, Migrations, CannedResponse } from '../server/models'
 import { log } from '../lib'
 
 // To add a migrations, add a new migration object to the
@@ -183,6 +183,16 @@ const migrations = [
       console.log('added creator_id field to campaign')
     }
   },
+  {
+    auto: true, // 14
+    date: '2019-05-13',
+    migrate: async () => {
+      console.log('setting sequence value for canned_response')
+      const maxId = await r.knex('canned_response').max('id').first()
+      await r.knex.raw(`ALTER SEQUENCE canned_response_id_seq RESTART WITH ${maxId.max + 1}`)
+      console.log(`set sequence canned_response_id_seq to ${maxId.max + 1}`)
+    }
+  }
 
   /* migration template
      {auto: true, //if auto is false, then it will block the migration running automatically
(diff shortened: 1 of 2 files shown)


joemcl:
Fixes Issue in #1106 (hopefully)

lperson:
> Fixes Issue in #1106 (hopefully)

inshAllah

src/server/api/schema.js
```diff
@@ -177,12 +178,10 @@ async function editCampaign(id, campaign, loaders, user, origCampaignRecord) {
     const convertedResponses = []
     for (let index = 0; index < cannedResponses.length; index++) {
       const response = cannedResponses[index]
-      const newId = await Math.floor(Math.random() * 10000000)
-      convertedResponses.push({
+      convertedResponses.push(_.omit({
```
schuyler1d: Two things here:
1. We don't need `lodash` or _.omit here, To do this in ES6, you could do this:
```
{ ...response, 
 campaign_id: id,
 id: undefined }
```
2. I'm a little nervous about us deleting and recreating canned_responses every time, but I guess if we've done that in the past, then there's no harm.

src/migrations/index.js
```diff
@@ -183,6 +183,16 @@ const migrations = [
       console.log('added creator_id field to campaign')
     }
   },
+  {
+    auto: true, // 14
+    date: '2019-05-13',
+    migrate: async () => {
+      console.log('setting sequence value for canned_response')
+      const maxId = await r.knex('canned_response').max('id').first()
```
schuyler1d: I think we need to accomodate zero-canned responses case, so maybe `const maxId = (await ....) || 0`

schuyler1d (changes requested):
small tweaks, but close.

joemcl:
Re "It's plausible that if we attempt to save a canned_response with a duplicate ID, the update fails, the new canned_responses are not saved, and it appears to the user as if they disappeared" - yup that definitely happened to me, see the log in #1106 . None of the Canned in that batch were saved for sure with the fail of the entire batch insert because of a dupe Canned ID. I checked the back end table.

src/migrations/index.js
lperson: I tested that case @schuyler1d.  If there are no rows `await r.knex('canned_response').max('id').first()` returns `NULL`, but adding 1 to `NULL` results in 1.

src/server/api/schema.js
lperson: 👍

src/migrations/index.js
lperson: 👍

src/migrations/index.js
```diff
@@ -1,4 +1,4 @@
-import { r, Migrations } from '../server/models'
+import { r, Migrations, CannedResponse } from '../server/models'
```
schuyler1d, line 1: Let's remove CannedResponse here -- I don't think it's required and it's possible it will subtly affect import order.

schuyler1d (approved):
:+1:

joemcl:
thanks @lperson and @schuyler1d - how fast can we get this into staging-main and I'll deploy that on our staging for testing? Or @lperson guess I can just deploy your branch to our staging...

schuyler1d:
> thanks @lperson and @schuyler1d - how fast can we get this into staging-main and I'll deploy that on our staging for testing? Or @lperson guess I can just deploy your branch to our staging...

Added to stage-main in #1108 (we probably won't get to test it this week though)

joemcl:
thanks @schuyler1d will deploy and test stage-main then

lperson:
> thanks @schuyler1d will deploy and test stage-main then

@joemcl which of the WFP branches will we be merging this into?  I'll make a PR in the WFP repo.

joemcl:
> > thanks @schuyler1d will deploy and test stage-main then
> 
> @joemcl which of the WFP branches will we be merging this into?

@lperson I'll create a new branch on our repo

PR #1110, opened 2019-05-15
WIP: Remove texter frontend race conditions

schuyler1d:
This ports the work from #965 on the frontend which works through some race conditions depending on the following things 'racing'
* loading more assigned contacts
* loading contact information
* completing text messaging

This crops up when dynamic assignment texting is extremely fast (like when it is cached).

It also stubs out some tests for texter components

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,6 @@ cache:
   yarn: true
 services:
 - postgresql
-addons:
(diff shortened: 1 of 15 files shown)


joemcl:
Hey @schuyler1d - isn't all Dynamic assignment texting cached?

PR #1111, opened 2019-05-18
Document local login as the dev default and add docs list to README

schuyler1d:
@jlegrone can you review this to:
1. make sure this won't break any of the Heroku/Docker things
2. Allow us to use PASSPORT_STRATEGY=local in Heroku pipelines

@azuzunaga -- Local login to be the default!


app.json
```diff
@@ -11,18 +11,24 @@
...
+
     "AUTH0_DOMAIN": {
-      "description": "auth0 domain attached to auth0 account",
+      "description": "auth0 domain attached to auth0 account (if not using auth0, enter anything)",
       "required": true
     },
 
     "AUTH0_CLIENT_ID": {
-      "description": "auth0 client id attached to auth0 account",
+      "description": "auth0 client id attached to auth0 account (if not using auth0, enter anything)",
```
jlegrone, line 26: Is this note necessary? Ideally I think leaving this empty should be fine.

jlegrone: Ah I see that it's required at the moment... 🤔

schuyler1d: Yeah, I think we should keep it required, but just wanted people not to get blocked like I did until you had that rather good suggestion :-)
Very open to other suggestions.

docs/HOWTO-configure-auth0.md
```diff
@@ -0,0 +1,123 @@
...
+There are two authentication backends supported for Spoke.  The first and default is using
+[https://auth0.com](https://auth0.com). This service allows support for e.g. Google authentication
+and others, and supports things like password resets and account management separate from Spoke.
+
+The alternative and default for development environments is local login, where passwords are hashed locally
+in the database and resets, etc are administered all within Spoke.  While good for development, we
+believe Auth0 still provides better security for production environments.  Below are the steps to configure
+Spoke for Auth0.
+
+## For Development
```
bchrobot: This seems a little confusing as there is no accompanying "For Production" section. I think step 4 is the only difference, so maybe better to annotate step 4 about dev vs. production values (see suggested change below)

docs/HOWTO-configure-auth0.md
```diff
@@ -0,0 +1,123 @@
...
+
+1. First configure the environment variable `PASSPORT_STRATEGY=auth0` in `.env` or wherever to configure Spoke environment
+variables.
+2. Create an [Auth0](https://auth0.com) account. In your Auth0 account, go to [Applications](https://manage.auth0.com/#/applications/), click on `Default App` and then grab your Client ID, Client Secret, and your Auth0 domain (should look like xxx.auth0.com). Add those inside your `.env` file (AUTH0_CLIENT_ID, AUTH0_CLIENT_SECRET, AUTH0_DOMAIN respectively).
+3. Run `yarn dev` to create and populate the tables.
+4. In your Auth0 app settings, add `http://localhost:3000/login-callback` , `http://localhost:3000` and `http://localhost:3000/logout-callback` to "Allowed Callback URLs", "Allowed Web Origins" and  "Allowed Logout URLs" respectively. (If you get an error when logging in later about "OIDC", go to Advanced Settings section, and then OAuth, and turn off 'OIDC Conformant')
```
bchrobot:
```suggestion
4. In your Auth0 app settings, set the following (Note: for development use `http://localhost:3000` instead of `https://yourspokedomain.com`):
    + **Allowed Callback URLs** - `https://yourspokedomain.com/login-callback`
    + **Allowed Web Origins** - `https://yourspokedomain.com`
    + **Allowed Logout URLs** - `https://yourspokedomain.com/logout-callback`
    + **Allowed Origins (CORS)** - `https://*.yourspokedomain.com`
        - _Note: for development this will be `http://localhost:3000` without the `*.`_
    + If you get an error when logging in later about "OIDC", go to Advanced Settings section, then OAuth, and turn off 'OIDC Conformant'.
```

shakalee14 (approved):

Issue #1113, opened 2019-05-21
Require First Name, Last Name for new user on Sign Up


joemcl:
Our log in page now points to spokewf.auth0.com and the new Sign Up link does not require a first or last name, which obviously we need for each new user.

cgalbiati:
I'm going to look into this a bit, but if someone else finds it first, let me know

cgalbiati:
Turns out you can add validation to the custom login form.  I added it to the docs in this PR: https://github.com/MoveOnOrg/Spoke/pull/1115, but someone will have to make the update in the prod auth0 account as well

joemcl:
@cgalbiati - the fix is this, per @bchrobot - ```Have you done step 11 (new-ish) in the readme?
https://github.com/MoveOnOrg/Spoke/blob/main/README.md ``` . 

We're back up and running with our  previous login page ( see 'Sign Up' at spoke.workingfamilies.org ).

But the thing is - we had the customized log in set up previously - since our initial deploy of Spoke on Heroku over a year ago. 

I copied and pasted the code in the link in step 11, above - and re-enabled the Customize Login Page toggle in auth0 settings. 

What changed from previously and I think is the cause of the change is that we moved from the Heroku 16 stack to the Container stack on Heroku. 

So @jlegrone - something about changing from Heroku 16 to Container stack and deploying upstream stage-main appears to disable Customize Login Page toggle in auth0 and remove previous customized code in that.  

cc: @deasterdaywfp .

joemcl:
@bchrobot when I changed from Heroku 16 to Container and deployed stage-main in the container, that disabled the previous auth0 Customize Login Page toggle, with presumably the original Spoke Auth0 Lock code we used when setting up Spoke originally early last year. I turned on the Customize Login Page toggle and used the newer code from step 11 in the readme. 

Some users logged in or signed up as new during the window when the standard credentials were in place, which did not require first name or last name, so I'm adding those first and last to the user records in Spoke. 

Four of the folks who logged in during the window were already Spoke users, but, they used Google to auth, and now we have two user records for them, with same e-mail, but with auth0 ID in `auth0_id` in one user record, and google-oauth2 ID in the other. Which is not a big deal, I'll leave the two user records as they are after updating the newer record with first and last names and cells.

[...]

PR #1115, opened 2019-05-22
Issue 1113: First and last name should be required in auth0 login page

cgalbiati:
https://github.com/MoveOnOrg/Spoke/issues/1113

### DIFF
--- a/docs/HOWTO-configure-auth0.md
+++ b/docs/HOWTO-configure-auth0.md
@@ -93,11 +93,19 @@ callback(null, user, context);
           additionalSignUpFields: [{
             name: 'given_name',
             icon: 'https://upload.wikimedia.org/wikipedia/commons/c/ca/1x1.png',
-            placeholder: 'First Name'
+            placeholder: 'First Name',
+            validator: (name) => ({
+              valid: name.length >= 1,
+              hint: 'Required'
+            })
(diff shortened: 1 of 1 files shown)


joemcl:
thanks @cgalbiati and my Issue title was not clear - First Name and Last Name were not showing *at all* as fields in the Sign Up link, much less as required. I figured out that problem, the fix is in the Issue comments in #1113 .  If this now makes sure those are required, all the better, and thank you.

ibrand:
Hey @cgalbiati! I'm noticing that the issue itself here was resolved, but seems like your PR adds validation that wasn't previously there to ensure that there actually *is* a first and last name. Is that correct? If so, we should merge this regardless of its connection to #1113

ibrand:
@cgalbiati wanted to ping you again on this one cause I'd love to merge it if it's still relevant :)

cgalbiati:
@ibrand My PR does add extra validation, so if that is desired, then it should be merged.  I don't actually have access to do so, so someone else will have to.

ibrand:
@cgalbiati thank you so much for your work on this. Sounds like the team discussed and validation isn't needed here after all so I'm going to close the PR. That said, I really appreciate you jumping in and submitting a PR!

PR #1130, opened 2019-05-31
Scalingport campaign cache

schuyler1d:
This is part of an ongoing project to merge in pieces of #965 

It does the following
1. fix security issue in organization root -- currently this is globally exposed.
2. fix cache staleness in DataLoader (in src/server/models/index.js)   for local data after object saves. This fixes a bug where in the same request `loaders.<object>.load` will have a stale result after a save.  This fixes that by adding clear calls in the right place especially as part of caches.
3. fix a small bug in CACHE_PREFIX which ignores it because `|` (binary add) was used accidentally instead of `||` (logical OR)
4. enable tests to run with cache enabled in travis run
5. use interactionStep cache when available from campaign cache
6. Improve the export pattern in cacheable_queries
 
Test adding
 - [x] get/set/get interaction steps confirming caches are cleared
 - [x] get/set/get campaign data confirming caches are cleared

Notes for reviewers:
* `__test__/` mostly adds server/api/campaign.test.js, but some methods are tweaked test_helpers.js in service of those tests
* `src/containers/` changes are mostly to expose the queries for use in the campaign.test.js tests -- nothing is changed except moving some code and removing console.logs
* `src/components/` deleting unused imports
* `src/lib/interaction-step-helpers.js` adds a function for server/models/cacheable_queries/campaign.js
* `src/server/models/index.js` tweaks `loaders` so it is not dynamically generated each time it is imported.  This allows some circular imports in models/cacheable_queries/ files so that we can clear the DataLoader in-memory cache (even more local than redis, e.g.).  If the DataLoader cache is not cleared, there are some bugs where campaign lookups right after a save will not be fresh.
* `src/server/models/cacheable_queries/*` -- there are some small tweaks to several files that fix two things: 1. cache keys need to prefix with `||` instead of `|` and 2. import structure is a little cleaner
* `src/server/models/cacheable_queries/campaign.js` -- this is the biggest change.  Most importantly, interactionSteps and customFields are saved with the campaign cache to save many inefficient queries per-texter request.


shakalee14 (dismissed):

src/lib/interaction-step-helpers.js
```diff
@@ -91,3 +91,25 @@ export function makeTree(interactionSteps, id = null) {
...
 }
+
+export function assembleAnswerOptions(allInteractionSteps) {
+  // creates recursive array required for the graphQL query with 'answerOptions' key
+  const interactionStepsCopy = allInteractionSteps.map(
+    is => ({...is, answerOptions: []}))
+  allInteractionSteps.forEach(interactionStep => {
+    if (interactionStep.parent_interaction_id) {
+      const [parentStep] = interactionStepsCopy.filter(
+        par => (par.id === interactionStep.parent_interaction_id))
```
ibrand: nitpick, but could this var be `parent` instead of `par` for readability?

src/server/models/cacheable_queries/canned-response.js
```diff
@@ -36,9 +35,11 @@ export const cannedResponseCache = {
       }))
       await r.redis.multi()
         .set(cacheKey(campaignId, userId), JSON.stringify(cacheData))
-        .expire(cacheKey(campaignId, userId), 86400)
+        .expire(cacheKey(campaignId, userId), 43200)
```
ibrand: what's up with this magic number?

src/server/models/cacheable_queries/lib.js
```diff
@@ -0,0 +1,16 @@
+
+
+export const modelWithExtraProps = (obj, Model, props) => {
+  const newObj = { ...obj }
+  const extraProps = {}
+  props.forEach(prop => {
```
ibrand, line 12: Do we need this? Seems like we're exactly coping `newObj` into `extraProps` unless I'm missing something. Couldn't `newObj` be used on line 12 instead of `extraProps`?

src/server/models/cacheable_queries/opt-out.js
```diff
@@ -1,12 +1,13 @@
...
 // STRUCTURE
-// maybe HASH by organization, so optout-<organization_id> has a <cell> key
+// SET by organization, so optout-<organization_id> has a <cell> key
+//  and membership can be tested
 
 const orgCacheKey = (orgId) => (
 !!process.env.OPTOUTS_SHARE_ALL_ORGS
- ? `${process.env.CACHE_PREFIX | ''}optouts`
- : `${process.env.CACHE_PREFIX | ''}optouts-${orgId}`)
+ ? `${process.env.CACHE_PREFIX || ''}optouts`
```
ibrand, line 9: were these `|`s all meant to be ORs before and haven't been working?

src/server/models/cacheable_queries/opt-out.js
```diff
@@ -23,12 +24,12 @@ const loadMany = async (organizationId) => {
     for (let i100 = 0, l100 = Math.ceil(cellOptOuts.length / 100); i100 < l100; i100++) {
       await r.redis.saddAsync(hashKey, cellOptOuts.slice(100 * i100, 100 * i100 + 100))
     }
-    await r.redis.expire(hashKey, 86400)
-    console.log(`CACHE: Loaded optouts for ${organizationId}`)
+    await r.redis.expire(hashKey, 43200)
+    // console.log(`CACHE: Loaded optouts for ${organizationId}`)
```
ibrand: delete commented code?

ibrand (commented):
I'm seeing "export const dataQuery" as a really consistent refactor. Is this a strong preference? Is it possible to enforce this type of behavior via a linter?

Also, any rollout plan needed? Looks like no, but this is also a fairly big changeset

src/server/models/cacheable_queries/opt-out.js
schuyler1d: exactly.  These were accidental bitOR but should have been ORs.

src/server/models/cacheable_queries/canned-response.js
schuyler1d: 12 hours.  will add a comment

src/server/models/cacheable_queries/lib.js
schuyler1d: I'll add some documentation on the function -- note that we delete properties from newObj, so in the spirit of 'functional' style, where you generally want to avoid altering objects inputted into functions.  Thus we do the copy.  This is an odd dance for the use-case of attaching properties to Models but allowing Model validation to take place.

src/server/models/cacheable_queries/opt-out.js
schuyler1d: exactly.  These were accidental bitOR but should have been ORs.

src/components/AssignmentTexter.jsx
```diff
@@ -140,7 +140,7 @@ export class AssignmentTexter extends React.Component {
         .slice(newIndex, newIndex + BATCH_GET)
         .map((c) => c.id)
         .filter((cId) => !force || !this.state.contactCache[cId])
-      console.log('getContactData missing current', newIndex, getIds)
+      // console.log('getContactData missing current', newIndex, getIds)
```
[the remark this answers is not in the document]
schuyler1d: I deleted commented code from the server parts.  I'd like to leave it in on the client side, as we've seen some reports from WFP that some race conditions are still causing issues.  These console.logs are really useful when trying to reproduce some of these scenarios and see what went wrong.

So I'd like to keep them for now.  Before this PR they were *running* so even if we remove them, I would vote that doesn't block this PR, and someone else can lint and purge console.logs else-times.

schuyler1d:
ok, addressed feedback @ibrand .  LMK if this looks good.

ibrand (approved):

Issue #1060, opened 2019-02-19
Add Favicon


shakalee14:
Add spoke logo as favicon.

deasterdaywfp:
Love this! Can you share the Spoke logo @shakalee14 ? I don't think I've seen it before.

joemcl:
@deasterdaywfp see https://opensource.moveon.org/spoke

filafb:
Hey all! Just opened a PR #1135 for this. I got to work on this issue last progressive hacknight and forgot to report here (sorry)

shakalee14:
Fixed with #1148

PR #1135, opened 2019-06-04
Add favicon Issue #1060

filafb:
Hey! Sky assigned [this issue](https://github.com/MoveOnOrg/Spoke/issues/1060) for me two weeks ago (I forgot to comment on it, sorry).

I included a folder called `/public` inside `/server` to hold the favicon asset. My first thought was sync this asset somehow to the `/asset` folder webpack creates during the build process and that holds the `bundle.js` file, but decided to avoid messing with the webpack config in my first PR.

Also, I got >400 errors running `npm run lint`. Not sure if it's a local configuration though.


schuyler1d:
Hi @filafb -- this is a great start!

Can you add some instructions in a docs/ file on how to add a favicon.ico on deploy?

Alternatively, if you want to create a favicon.ico file based on this:
https://s3-us-west-1.amazonaws.com/spoke-public/spoke_logo.svg
maybe we could serve it inline instead of it being customizable -- (I think there might even be a way to import binary data, and serve directly since the file will be so small)

PR #1139, opened 2019-06-12
remove rootQuery contact

schuyler1d:
@shakalee14 maybe we can add this to #1138 ?

This is a security vulnerability:
* anyone with merely a TEXTER privilege can iterate through all the contact details including all the messages sent, etc.

(Note: I don't see it used in any of our frontend)


ibrand (approved):
++ to fixing a security vulnerability!

PR #1147, opened 2019-06-18
fix application of texting hours per campaign in client

lperson:


ibrand:
Do you have any extra context on this one @lperson? Does it fix a documented issue? What's the current state and what's the desired state?

I know I'm asking a bunch of questions, but want to be able to risk assess as I'm merging PRs. Thanks for your work on this!

lperson:
@ibrand  there's not a lot of context here.  This is a bug fix for a stupid bug.  I'm not sure if the issue was documented in a GitHub issue.  It probably came up because WFP reported to me in slack that it didn't work, so I just did a fix.  It is probably already in their production.

The issue was probably "Spoke is allowing testers to send texts even though a campaign has texting hours configured and the text was sent outside the campaign's texting hours."

Do you need context about organization texting hours and campaign texting hours as well?

ibrand:
I don't need any context on organization texting hours and campaign texting hours but thanks for offering! This looks good to merge 🚀

PR #1154, opened 2019-06-22
Switch to Knex migrations

schuyler1d:
This builds on @harpojaeger's excellent work in #692 and should get us out of the hell of race conditions when creating/destroying the db -- especially during testing runs.

Using knex's native migrations support also should improve usability and predictability (esp. for anyone that is already familiar with knex).

One thing that we'll need to be careful of:
Before merging into main, we probably want to cut a new release (3.0?) with some documentation on how to migrate from an older version that still has migrations between current main and when this PR is merged.  In that case, they should update their code to the latest release BEFORE this one, run migrations, and only then, checkout this version (or any future versions).  I can add documentation for that, but since it will be release/version-specific, I thought I'd hold off for guidance on placement.  If we want, we could skip deletion of the old migrations file, and then have some `yarn old-migrations` command that you would run.  However, that might create confusion going forward if new devs find a src/migrations/index.js and incorrectly update that instead.

From Harpo's changes, this does the following:
* updates the schema changes since the PR -- send_before tweaks and campaign.creator_id
* removes Harpo's __test__/server/db/knex.test.js -- this was a useful test while experimenting with the upgrade, but once we switch, the db will be out of date and it makes no sense to maintain the schema both in tests and in migrations (or maybe it does?)
* creates the package.json local command `yarn knex ....` for easy migration
* fixes some bugs introduced: (restoring datawarehouse connection and knex.destroy() was not working to remove the existing tables) -- also removes `dotenv` dependency loading in server/index.js to preserve independence from .env contexts vs. e.g. other testing.
* upgrades knex to 0.17.3

cc: @ibrand and @lperson (since he is most likely to create schema change PRs)


__test__/server/db/utils.js
```diff
@@ -0,0 +1,33 @@
+export const tables = [
+  'log',
+  'message',
+  'user_cell',
+  'job_request',
+  // 'migrations' // Now automatically managed by Knex
```
ibrand: Wondering if we should just delete this line? Cause a few months after this PR this becomes a reference to a past state of the code that's no longer relevant? Could be useful though

knexfile.env.js
```diff
@@ -0,0 +1,10 @@
+// Update with your config settings.
```
ibrand: Is this a leftover comment from knex's docs? Or is the implication here that anyone who pulls this change needs to make some personal change to this file?

migrations/20190207220000_init_db.js
```diff
@@ -0,0 +1,355 @@
+const initialize = async (knex, Promise) => {
```
ibrand, line 1: Is this file an initial migration that folks need to run to get everything set up on the new system? It seems like the knex CLI should be covering us for future migrations. If it is an initial migration that needs to be run, can we label it with a comment block and document it?

src/server/models/thinky.js
```diff
@@ -1,61 +1,13 @@
 import dumbThinky from 'rethink-knex-adapter'
 import redis from 'redis'
 import bluebird from 'bluebird'
+import config from '../knex.js'
 
 bluebird.promisifyAll(redis.RedisClient.prototype)
 bluebird.promisifyAll(redis.Multi.prototype)
 
-// // This was how to connect to rethinkdb:
```
ibrand, line 8: Has all of this been lifted verbatim into `knex.js` as a refactor? Or is there a bigger difference that I'm missing?

src/server/knex.js
```diff
@@ -0,0 +1,73 @@
+// Define a Knex connection. Currently, this is used only to instantiate the
```
ibrand: Wondering if we should should rename this to `knex-config.js` for clarity. From the name I wouldn't have guessed that the whole file is a config

ibrand (commented):
I want us to outline a step by step gameplan for the order of operations with rolling this out re:
> Before merging into main, we probably want to cut a new release (3.0?) with some documentation on how to migrate from an older version that still has migrations between current main and when this PR is merged.

Here are a few guiding questions:
1. This is not backwards compatible, so what are the exact steps that folks would need to take after they git pull to get fully onto this new system
2. Will this influence anyone's current dev envs if they pull assuming they aren't doing a migration right now?
3. How does `migrations/20190207220000_init_db.js` fit into this transition?

Awesome work @schuyler1d and @harpojaeger I'm really excited about this, and want to make sure we merge it as seamlessly as we can!

src/server/models/thinky.js
schuyler1d: Should be a verbatim copy

src/server/knex.js
schuyler1d: maybe knex-connect.js?  the config is in environment variables, but this establishes the connection

schuyler1d:
> Here are a few guiding questions:
> 
>     1. This is not backwards compatible, so what are the exact steps that folks would need to take after they git pull to get fully onto this new system

* For anyone without a database yet (or one they need to keep), then answer is nothing -- it will get auto-created when they start for the first time
* For anyone that started a database with a version of main prior to June 3rd (when the last migrations/index.js change was made) and have NOT updated since then:
```
git checkout 7799081deb #current main
# deploy or run instance which triggers migrations (without SUPPRESS_MIGRATIONS or SUPPRESS_DATABASE_AUTOCREATE variables set)
git checkout main #post knexmigrations branch presumably
# deploy or run instance # this will do nothing if no additional migrations have been made, and run newer migrations if there have been.
```

>     2. Will this influence anyone's current dev envs if they pull assuming they aren't doing a migration right now?

If they had a database created before March (the time the last dev-affecting schema change was made), then they should clear/delete the database and let it autocreate again -- or do the instructions in (1)

>     3. How does `migrations/20190207220000_init_db.js` fit into this transition?
It seems knexmigrations ignores the init migration if it is already connected to a database.  So the important thing here is that it reflects the current state of prod (which it does at the moment) -- first installs/runs will auto create it with the init_db file.

migrations/20190207220000_init_db.js
schuyler1d: it is an initial migration, but does not need to be run manually, because of this:
https://github.com/MoveOnOrg/Spoke/pull/1154/files#diff-6e0d62f54b853b53af874f5965d7adf2R36
i.e. on startup, by default migrations or initial db creation happen automatically.
So knex CLI is mostly useful for creating new migrations or manipulating them for db admin or doing it manually, e.g. in a production setting, where you want to run manually instead of let it 'autorun'

yarn.lock
```diff
@@ -2793,13 +2793,6 @@ bl@~1.1.2:
   dependencies:
     readable-stream "~2.0.5"
 
-block-stream@*:
```
ibrand: Is this file accidentally included? Or included on purpose?

schuyler1d: AIUI, yarn.lock is updated occasionally based on upstream package dependencies.
I can remove the yarn.lock changes in this PR, if desired, but we should definitely update yarn.lock regularly -- and if we don't merge the changes here, then every time people do a `yarn install` the same changes will enter through a different PR.

PR #1156, opened 2019-06-24
WIP: OSDI People for Contacts

joshco:
> I'm posting this for review.  This is my first PR for Spoke, so please guide me process-wise.

This pull request implements the OSDI people collection on top of Spoke Contacts.  
It supports the bulk and single person signup helper to insert people into Spoke.  There is an example python script under dev-tools which supports the OSDI version.

The bulk OSDI extension spec is here:
https://github.com/opensupporter/osdi-docs/pull/335

It also exposes Spoke Contacts via the the OSDI people collection so a client could enumerate and download the contacts from a campaign/org.

The OSDI API Entry Point for Spoke is specific to the organization and campaign like:
http://spoke.example.com/osdi/org/1/campaigns/1/api/v1
So each campaign/org collection has its own AEP.  I took a guess on this architecture, if folks have input on how this should be implemented, lmk.

@joemcl @lperson


joshco:
Closing in favor of https://github.com/MoveOnOrg/Spoke/pull/1166

PR #1174, opened 2019-07-17
WIP: Enable sqlitetests

schuyler1d:
(note this is against knex-migrations-sky branch, since they don't seem to pass on main but do on that branch)

Several things in the app can break with sqlite which is a much easier dev environment to run and helps us stay database-agnostic in the code base.  I think we should enable tests there now/if they are passing.


ibrand (approved):

ibrand:
Hey @schuyler1d where did this end up? Is this worth revisiting now that knex migrations are merged?

schuyler1d:
Ideally someone else could take this up.  The main work left to do, I believe is __test__/server/api/assignment.test.js needs to be database-agnostic (right now it just runs on postgres)

schuyler1d:
closing because #1387 succeeds

PR #1175, opened 2019-07-19
Copy campaign async Issue #1141

filafb:
Issue #1141 

This version make the copy of interactions and canned responses more async proof. 

Tests created to cover:
1. Copying campaign (I didn't find any one covering this process);
2. Make sure the copied campaign has the same canned responses as the original one;
3. Make sure the copied campaign has the same interactions as the original one.

note about tests:
1. I replicated a few steps (created user, campaign, canned responses, and interactions). Couldn't find a ready-to-use in a upper scope. 
2. I'm calling a `beforeEach` hook for every test case. It's not the best case for performance, but it was the best way I found to avoid race condition with promises. The old code failed considering this test setup.

### DIFF
--- a/__test__/backend.test.js
+++ b/__test__/backend.test.js
@@ -7,11 +7,13 @@ import {
   superAdminRequired
 } from '../src/server/api/errors'
 import { graphql } from 'graphql'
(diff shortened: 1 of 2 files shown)


shakalee14 (approved):

shakalee14:
Merged with #1203

PR #1179, opened 2019-07-27
optimize display of interaction steps edit

### LINKED ISSUE #1335: Improve performance of interaction steps editor
As part of RFC #1320 and the project [Bring work in WorkingFamilies/Spoke fork into MoveOnOrg/Spoke](https://github.com/MoveOnOrg/Spoke/projects/14), merge the work in these WorkingFamiles/Spoke PRs:
- [122](https://github.com/WorkingFamilies/Spoke/pull/122) 
- [124](https://github.com/WorkingFamilies/Spoke/pull/124)  

See **_[MoveOn PR 1179](https://github.com/MoveOnOrg/Spoke/pull/1179)_**

lperson:
Fixes #1335.

Two things are changed here.

### Interaction step edit loading on safari

A user reported "I need to step through every interaction step before I can do anything."

Upon investigation we realized it only happens on Safari. That's why I never saw it.

It seems that Safari is causing onFocus to fire for the script in every interaction step after the first render. It's not clear to me why that was happening.

We were handling onFocus to cause the dialog to open and present the script editor, rather than handling it on the click. That seems unnecessary, because is anybody really going to tab from script to script and expect the dialog to open? Once it opens, they can't tab to the next step anyway.

So I changed it to cause the script editor to open only when the field is tapped. That's a hack that works around the root cause, but as I hoped I explained above, this solves the problem and hopefully won't break anyone else's workflow.

This fix has been tested on WFP production.

### Slow loading of interaction step edit on mobile devices

This happens when a campaign has a lot of interaction steps and the user is on a mobile device.  

The fix was to do the render twice -- once with just the first interaction step node, and then with all the nodes.

My reasoning for this change is that it takes react longer to create the DOM than to update it.


ibrand:
@lperson I'm seeing the test suite failing with an error that looks related to this PR when I check this out:

```
Summary of all failing tests
 FAIL  __test__/lambda.test.js
  ● Test suite failed to run

    Cannot find module 'react-dot-fragment' from 'CampaignInteractionStepsForm.jsx'

      45 |     interactionSteps: this.props.formValues.interactionSteps[0] ? this.props.formValues.interactionSteps : [{ id: 'newId', parentInteractionId: null, questionText: '', answerOption: '', script: '', answerActions: '', isDeleted: false }],
      46 |     displayAllSteps: false
    > 47 |   }
         |    ^
      48 |
      49 |   componentDidUpdate = (prevProps, prevState) => {
      50 |     if (!this.state.displayAllSteps) {

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:221:17)
      at Object.<anonymous> (src/components/CampaignInteractionStepsForm.jsx:47:25)


Test Suites: 1 failed, 21 passed, 22 total
```

lperson:
I'll look @ibrand.  Sorry for the delay.

ibrand:
Did you end up looking over these tests @lperson ?

schuyler1d:
Can you merge `main` and add a stub test for CampaignInteractionStepsForm -- doesn't even have to cover Fragment stuff -- just one or two things that confirm the component doesn't fail instantiation

lperson:
@schuyler1d I added some basic tests for CampaignInteractionStepsForm.

src/components/CampaignInteractionStepsForm.jsx
```diff
@@ -50,16 +43,31 @@ export default class CampaignInteractionStepsForm extends React.Component {
...
+  };
+
+  /*
+    this.statye.displayAllSteps is used to cause only the root interaction
+    node to render when the component first mounts.  ComponentDidMount sets
+    displayAllSteps to true, which forces render to run again, this time
+    including all interaction steps. This cuts half the time required to
+    render the interaction steps after clicking the header to expand the
+    interaction steps card.
+  */
```
schuyler1d, line 66: can we add a 'FUTURE: Re-evaluate after React 16 upgrade' comment

schuyler1d (approved):
:+1:

src/components/CampaignInteractionStepsForm.jsx
ibrand: @lperson I merged this into `stage-main` but maybe you could even put this comment into a future PR that's unrelated?

lperson: #1342

PR #1180, opened 2019-07-30
fix server side err -> id to ids

shakalee14:
Erik saw this error behavior in production after the latest deploy. I fixed the reference error.

### DIFF
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -666,11 +666,9 @@ const rootMutations = {
       )))
 
       campaigns.forEach(campaign => { campaign.is_archived = true })
-      await Promise.all(campaigns.map(campaign => {
-        campaign.save()
-        for (let i=0,l=ids.length; i<l; i++) {
-          cacheableData.campaign.clear(ids[i])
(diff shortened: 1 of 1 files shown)


src/server/api/schema.js
```diff
@@ -668,7 +668,7 @@ const rootMutations = {
       campaigns.forEach(campaign => { campaign.is_archived = true })
       await Promise.all(campaigns.map(campaign => {
         campaign.save()
-        cacheableData.campaign.clear(id)
+        cacheableData.campaign.clear(ids)
```
schuyler1d: I think if ids is an array, we need to do a `for (let i=0,l=ids.length; i<l; i++) { cacheableData.campaign.clear(id) }`

schuyler1d: https://github.com/MoveOnOrg/Spoke/blob/6f82c3f94f1b654653d9f7093bc69a7712e74f66/src/server/models/cacheable_queries/campaign.js#L50

schuyler1d (approved):
:+1:

PR #1193, opened 2019-08-05
Add prettier formatting to Spoke

ibrand:
# Context
I asked the Progressive Coders Network #spoke channel on 7/19/19
> **Are you in favor of running Prettier on the Spoke repo and adding a pre-commit hook to ensure that all new code is “prettified” before being committed.**

as a way of collecting community agreement around having a consistent code-wide style and the question was received with a unanimous 9 👍 and no down votes.

With code style consistency, we wont run into community doubts, PR comments, or new contributor confusion over the basics of how to format our code and can discuss bigger things from a stronger foundation.

# How this PR is formatted
I broke this PR into three commits:
1. Applying the lint-staged installation from [the prettier docs](https://prettier.io/docs/en/precommit.html) to get the pre-commit config set up for all commits going forward
2. Removing the eslint parts of the default config since we haven't discussed that as a community yet
3. Run prettier on the repo specifically with the line: `prettier --write "**/*.js"` and then committing which allows the pre-commit hook to run on it as well.

This way, it'll be easy to read the changes I made before creating a diff the size of nearly the entire codebase by reviewing the config files and earlier commits.

# Testing
- The test suite passes with the same warnings seen on `main`
- The app runs as expected
- The commit hook fired when I committed the prettified files


lperson:
Hi @ibrand.  I'm glad we're doing this!

I notice this PR doesn't prettify the jsx files.  Let's do those, too!

Also, prettier is adding dangling commas, which, because of our current eslint configuration, is causing eslint to complain. 

This is the eslint setting for dangling commas: `"comma-dangle": ["error", "never"],`

I may tweak `.eslintrc` so that eslint doesn't complain about prettier's formatting (as a separate PR).

lperson:
@ibrand I noticed that when you ran prettier, it added semicolon and changed quotes to double quotes.

I have questions in two dimensions here.

First, do we want semicolons and double quotes?

Second, when I run prettier (using the command above), I'm *not* getting semicolons and double quotes.  ~Did you use options on the command line, or do you have a .prettierrc laying around?~  Oh damn!  I have a .prettierrc laying around in my home directory!

But even when I hide that, I'm getting different results than you did.  So I guess I'm curious whether you used options on the command line or a .prettierrc that is not committed.

ibrand:
Hey @lperson thanks for looking this over!

1. I agree. Let's do the jsx too. I wasn't sure if prettier would work for them, but now that you're mentioning it I honestly don't know why I had doubts there.
2. Thanks for mentioning that eslint change! I hadn't even realized we already had eslint and didn't see it complaining when I was running the app which is... interesting.
3. I ran that command and don't have any `.prettierrc` files. That said, after running that command, I committed and the commit hook ran. It's possible that the commit hook has options that are hidden. I'm seeing: 
```
    "*.{js,css,json,md}": [
      "prettier --write",
      "git add"
    ],
```
as the prettier setup in the config which to me implies defaults... I wonder what's going on with that.

lperson:
@ibrand do you maybe have a .prettierrc further up the directory tree? It looks all the way up to your home directory. 

Are you certain the eslint pre commit hook was not present when you ran prettier and committed? I saw you took out the eslint hook in the second commit but was it out when you ran prettier?

I'm just obsessing a little because I can't get the same output from prettier which is making the merge painful.

ibrand:
@lperson I just ran `find / .prettierrc` and got `find: .prettierrc: No such file or directory` to verify so it looks like there are no `.prettierrc` files on my system.

One possible plan would be to remove the actual prettifying step and then have you checkout the config, stage all the js files in your repo and commit so that the commit hook runs on all of them? That might make for more consistency?

shakalee14 (approved):

PR #1206, opened 2019-08-18
Update issue templates to include an RFC template

ibrand:


shakalee14 (dismissed):
Title confused me but this is for RFC (not RFPs).

ibrand:
Oh that's totally my bad. I can edit the commit msg too for the log. This is for an RFC. You're totally right.

shakalee14 (approved):

PR #1211, opened 2019-08-28
enable testing with real Redis service and fix a broken test

schuyler1d:
# Fixes redis caching server

## Description
While we ran test-cache in travis which used 'fake' redis (implemented in JS) using real redis is even better.  To wit, enabling it showed that a test broke in that scenario

# Checklist:
- [na] I have manually tested my changes on desktop and mobile


__test__/lambda.test.js
```diff
@@ -1,10 +1,10 @@
 import { handler } from "../lambda.js";
 import { setupTest, cleanupTest } from "./test_helpers";
 
-beforeAll(
-  async () => await setupTest(),
-  global.DATABASE_SETUP_TEARDOWN_TIMEOUT
-);
+beforeAll(async () => {
+  await cleanupTest();
```
ibrand, line 5: Learning question, not a critique: I'm guessing this is now needed because we're actually using a db and not a fake db, is that right?

__test__/server/api/campaign.test.js
```diff
@@ -375,8 +375,20 @@ describe("Reassignments", async () => {
...
+    // start campaign
+    // send 5 texts with texter1
+    // reassign 20 to texter2
+    // texter2 sends 5 which will get autoreplies (needsResponse)
+    // texter2 replies to 3
+    // reassign 10 more to texter2 from texter1
+    // use reassignCampaignContactsMutation (Message Center)
+    //     to reassign a messaged contact from texter1 to texter2
+    // use bulkReassignmentCampaign to reassign texter2 needsResponse => texter1
+    console.time('fullfunction');
```
ibrand, line 388: Why all these `console.log`s?

__test__/setup.js
```diff
@@ -2,3 +2,6 @@ import { configure } from "enzyme";
 import Adapter from "enzyme-adapter-react-15";
 
 configure({ adapter: new Adapter() });
+
+// server/api/campaign.test.js has some long tests so we increase from 5sec default
```
ibrand, line 6: 🎉 thanks for this comment! Helps with clarity

ibrand (approved):
Questions about `console.log`s but otherwise looking good :shipit:

__test__/lambda.test.js
schuyler1d: my guess is we load db info a little 'earlier' in the spoke app loading now (probably from knex migrations changes) -- and so whereas before this would run/load cleanly without a db cleanup, it now needs to load correctly.

PR #1236, opened 2019-09-25
Add explicit reference to `PGSSLMODE=require`

joemcl:
# Fixes # (issue)

## Description

# Checklist:


schuyler1d (approved):
:+1: 
If we had multiple Heroku buttons, maybe the 'paid' link would have PGSSLMODE=require by default?

joemcl:
thanks @schuyler1d - actually yeah think we make a minimal paid as the default Deploy to Heroku option with PGSSLMODE=require built in - with yeah another Deploy to Heroku button for 'hobby'/free deploy... and maybe even with a cheap default Twilio and Auth0 account sign up built in to the default paid Deploy script...maybe not doable, I get, but would be handy/quicker...

joemcl:
and there's an Auth0 "add on" available in Heroku, though not a Twilio "add on". Don't know if tweaking Deploy to Heroku script to add Auth0 add on makes sense - and yeah probably a $ premium for adding Auth0 that way - but, just like Bucketeer makes deploy simpler, though costlier than adding own S3 bucket connection directly, and Mailgun easier, maybe really required - perhaps adding in Auth0 add on helps. See https://elements.heroku.com/addons/auth0 .

PR #1245, opened 2019-10-09
remove faker which is 86M and not really used at all in live code

schuyler1d:
# Fixes build failing on deploy to AWS Lambda

## Description

The instructions in DEPLOYING_AWS_LAMBDA.txt don't work with current dependencies which are too large.  I remove `faker` which was used only in old files to generate a phone number easily 'faked' manually.

# Checklist:

- [na] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/server/api/lib/nexmo.js
+++ b/src/server/api/lib/nexmo.js
@@ -3,7 +3,6 @@ import { getFormattedPhoneNumber } from "../../../lib/phone-format";
 import { Message, PendingMessagePart } from "../../models";
 import { getLastMessage } from "./message-sending";
 import { log } from "../../../lib";
-import faker from "faker";
 
 let nexmo = null;
 const MAX_SEND_ATTEMPTS = 5;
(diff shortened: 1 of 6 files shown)


src/server/api/lib/nexmo.js
```diff
@@ -45,8 +44,12 @@ async function convertMessagePartsToMessage(messageParts) {
 
 async function findNewCell() {
   if (!nexmo) {
+    const num = "1234"
+      .split("")
+      .map(() => parseInt(Math.random() * 10))
+      .join("");
     return {
-      numbers: [{ msisdn: getFormattedPhoneNumber(faker.phone.phoneNumber()) }]
+      numbers: [{ msisdn: getFormattedPhoneNumber(`+1212555${num}`) }]
```
ibrand, line 52: The majority of this number isn't randomized anymore, what is `nexmo.js` used for? Will that be a problem at all?

ibrand (commented):
Want to make sure we're not gonna run into problems with this lack of randomization (I'm guessing it's fine, but want to confirm before approval)

schuyler1d:
Basically `rentNewCell` is completely dead code at the moment.  In both cases, the randomization is only done if an actual twilio account doesn't exist -- so that would be during testing.
So:
* dead code
* only used in testing
* if it were used in testing, and needed to get a new number, the result on the rest of the app is non-existent because we no longer send/pin by a cell number anyway

ibrand (approved):

PR #1255, opened 2019-10-20
clear cache to force contact to be refreshed from database

### LINKED ISSUE #1250: Bug: "Skip Reply" button missing
**Describe the bug**
On the "Send Replies" screen, texters do not see the Skip Reply button, at least initially. This means they end up with a backlog of "needs response" conversations they need to page through in order to get to the newest replies.

After some time (a few hours) the Skip Reply button does usually appear. I have also confirmed that restarting the server immediately resolves the issue.

This _feels_ like a caching issue to me, where the frontend is not seeing the updated message status, but I'm not well versed enough in the Spoke architecture yet to definitively figure out what's going on.

**To Reproduce**
1. Send some first texts
2. Wait for responses
3. Click Send Replies
4. See that the Skip Reply button is missing

**Screenshots**
2019-10-11 at 7 24 PM

 - OS: Mac and iOS
 - Browser: Chrome and Safari
 - Desktop or Mobile? Both
 - Version 3.1

**Additional context**
I've replicated this issue both on our production instance on Heroku and a test instance running on my local machine.

### LINKED ISSUE #1251: Bug: Reassigned contacts won't load
**Describe the bug**
Most of the time (but not always), when I reassign contacts from the Message Review page, the newly assigned user is not able to act on those contacts. When the new texter clicks on the Send First Texts or Send Replies button, they get a spinner and the message interface never loads.

I strongly suspect this is related to #1250, but since I'm not sure, I'm filing it as a separate issue. Like with #1250, after enough time passes or if I restart the server, the issue resolves itself.

**To Reproduce**
1. Assign contacts to Texter 1, send first messages and wait for replies
2. On the Message Review screen, filter to Needs Texter Response and reassign those contacts to Texter 2
3. Log in as Texter 2 and click Send Replies
4. You will get a spinner that never ends.

**Screenshots**
2019-10-12 at 1 57 PM
2019-10-12 at 1 58 PM

 - OS: Mac and iOS
 - Browser: Chrome and Safari
 - Desktop or Mobile? Both
 - Version 3.1

**Additional context**
I've replicated this issue both on our production instance on Heroku and a test instance running on my local machine.

lperson:
# Fixes #1250
# Fixes #1251
Note that #1251 can't be reproduced in `main`.  It's dependent on a fix in `stage-main` to be able to reassign.

## Description

This fix clears from the in-memory [DataLoader cache](https://github.com/graphql/dataloader) CampaignContacts being returned in GetAssignmentContact to force them to be reread from the database.  This is necessary because the campaign contact in the database might have been updated by `saveNewIncomingMessage` in a different process (`incomingmessagehandler`).  That addresses #1250.  It's a weakness that there could be other places in the future that rely on the cache being updated.  This fix address the problem at the point the cache is read, not the point the cache is written to, so there's an underlying problem still not being addressed.  We might want to consider removing the DataLoader caching or using custom caching with DataLoader in a way that the cache can be updated between processes.  Worth noting that tests in `texter.test.js` pass because they use `fakeservice`.  `fakeservice` does update the DataLoader cache, whereas `twillio` does not.  `fakeservice` works in tests because CampaignContact is being uploaded in the same process, but it fails in local installations running a separate `incomingmessagehandler` because the cache update is not reflected in the server process.

The fix also clears contacts from the DataLoader cache when reassigning them in `reassignConversations`.  That's necessary because `assignment_id` is changed in the database and not reflected in the cache.  That addresses #1251.

An earlier iteration of this fix considering dropping DataLoader caching altogether for CampaignContact.  I decided against doing that because there are several places that reference CampaignContact for static data and I did not want to add latency to them.

# Checklist:

- [N/A] If my change is a UI change, I have attached a screenshot to the description section of this pull request
- [N/A] I have made any necessary changes to the documentation
- [N/A] My PR is labeled [WIP] if it is in progress


lperson:
Good catch on this @jeffm2001!  TBH I'm surprised this hasn't come up sooner.

@ibrand and @schuyler1d Is the change in this PR a code smell to you?

jeffm2001:
@lperson I'm surprised it hasn't come up sooner too. I tried looking through the recent commit history to see if I could find where the bug was introduced, but came up empty.

I do have a thought on this proposed fix. I'm not that familiar yet with the Spoke architecture, so take this feedback for what it is, i.e. I could be completely wrong.

That said, as a general rule, it makes more sense to expire cached values when the data is updated rather than right before it's read. In this case, I think that would mean clearing the cache whenever a message is sent or received, and when a contact is reassigned or opted-out.

joemcl:
Hey @joshco can you merge this into your `osdi_key` branch and I'll test this on the site I'm running tonight, along with your other updates to that branch?

schuyler1d (dismissed):
I'm approving the code change however the tests aren't currently passing so that will block merging.

I agree that this is the 'least-worst' approach.  I think once we get campaignContact (redis) caches on a full footing, then we might consider just disabling campaignContact's DataLoader() wrapper (and just implement it as a dict of raw functions: `load` and an empty `clear`.  But for now this looks fine.

lperson:
@schuyler1d The tests pass now.  I found having one test per file works around limitations related to jest's require cache [as described by @jparkrr](https://github.com/MoveOnOrg/Spoke/blob/main/__test__/test_helpers.js#L29).  (This is not the case for all tests, just for tests that involve using graphQL queries in multiple jsx files.)

> I'm approving the code change however the tests aren't currently passing so that will block merging.

schuyler1d (approved):
:+1: 
Yeah, fwiw, I started using a different pattern that doesn't hit the query cache, e.g. in __test__/server/api/campaign.test.js -- you export the specific GQL query from the component rather than 'loading' it through the introspection mechanism.
I kinda like that we have both types of tests in the suite so that at least some things are tested both ways, but I think the campaign.test.js method is 'more scalable' rather than dividing up too much per-file.

Thanks!

PR #1270, opened 2019-11-07
use dotenv in dev and test only

jabyess:
# Closes issue #753 and PR #805  
This is code is all thanks to @harpojaeger, I ran into the same problems getting this set up on my local machine. 

## Description
Instead of fixing up/rebasing/updating #805 (so far behind), I just took the code changes and manually included them here. See referenced issues for more detail.

# Checklist:


ibrand:
Hey @jabyess ! I'm seeing the travis build failing and it looks like this line at the end is the culprit: https://travis-ci.org/MoveOnOrg/Spoke/builds/608524441#L3128

Any thoughts on why that might have broken?

jabyess:
Hmm that's very interesting. There's no REDIS_URL value in the .env file, so is this something that just travis uses?

ibrand:
@jabyess  I reading your question made me realize that I recognized Redis being related to tests from somewhere... We recently merged: #1211 and i think specifically [this line](https://github.com/MoveOnOrg/Spoke/pull/1211/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R13) is related to the problem? Is anything standing out to you here?

schuyler1d:
I think it would be both better code and possibly more flexible if the test was instead `if (process.env.NODE_ENV !== "production") { ... }`
I think the default node_env is development and won't necessarily be set at all

src/server/index.js
```diff
@@ -1,3 +1,8 @@
+// dotenv is used only in development & test
```
schuyler1d, line 1: why is this addition (in this file) needed at all?

It would be better to keep dotenv stuff out of src/ and more in build/startup contexts.

ibrand:
@jabyess The test failures were a fluke! I reran Travis and they passed

jabyess:
Omg! Super bizarre @ibrand. Gremlins at work.

src/server/index.js
jabyess: In this case, it's a runtime issue. The config needs to be loaded here because it enables debug mode in this file.

schuyler1d: It shouldn't.  I know it's a little labyrinthine, but if you run `yarn dev` that runs (see package.json) dev-tools/Procfile.dev which runs `./dev-tools/babel-run ./webpack/server.js` and  webpack/server.js loads webpack/config.js which *does* load dotenv.

The current setup actually will never load dotenv to begin with (and that's intentional) in production, because production runs server/index.js directly (or a built version of it).

So actually, is this PR required at all? 

(I see the issue might be that through the Docker file .env isn't loaded but then the current Dockerfile also sets everything with NODE_ENV=production anyway)

ibrand: Harpo had this as his fix a long time ago to address #753 we still see that problem when trying to set up people now. Are you sure that that full flow is actually executed when running `yarn dev`?

schuyler1d: If it's not, then we should solve the core problem rather than adding additional code that make it even more confusing.

schuyler1d (changes requested):
Let's solve the core problem rather than introduce more places that can load .env -- even multiple times.

If we are still seeing #753 -- then let's trace where the `passport` library gets loaded and ?move that?
Second, in dev, this should at least be a lot less urgent now that we have `local` passport auth for dev.

schuyler1d:
closing this PR since I'm not sure the issue is still present -- but if it is, then we'll need to do it differently as per comments above.

PR #1273, opened 2019-11-09
Message.error_code 

### LINKED ISSUE #1018: Add error_code and other important fields to Log table for better reporting
We found some pain points on reporting error results based on the current table structures.  The current structure has a few tables:
* `campaign_contact`
* `message` (joinable by assignment_id,contact_number => campaign_contact)
* `log` (joinable by message_sid to message) -- this is 

The key moments the data from an error comes back are here:
`handleDeliveryReport`: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/lib/twilio.js#L214-L234
`message.service_response`: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/lib/twilio.js#L169-L183

We want to do a few things -- add a contact_number/cell column to log which is more easily joinable direclty to campaign_contact, along with an error_code column -- these will make easier reporting/searching directly on the log table.  We also should carry over that error_code directly onto campaign_contact table in order to make fast reporting -- possibly directly from the campaign report dashboard for the campaign admins.

Finally, (as originally argued by GetUp when they created the log table, and I was against) -- service_response is just too big a blob to carry around on the message table.  We want to remove service_response column on message and move any data that comes back from the API response(s) into JUST the log table, and not update the message record.

- [ ] Log record should be created instead of adding content to message.service_response
- [ ] message.service_response should not be filled -- too much space, but content SHOULD go into log -- delete this column.
- [ ] error_code, contact_number and other useful fields should be separate columns in `log` for easy query
- [ ] add a column to campaign_contact for error result (e.g. for tracking undeliverability, etc. and so we don't need to query the message table). 

Complications:
* the campaign_contact.error_code should be the LAST error that came back, so any error found in a response should automatically update the campaign_contact.  While this might confuse reports on whether the first message was sent, knowing that can be done based on other queries (e.g. message_status) rather than the error_code.
* Note that in a (hopefully) [pending PR](https://github.com/MoveOnOrg/Spoke/pull/965): message.campaign_contact_id is added for faster join on campaign_contact (and better for other future features).  This shouldn't affect this work too much, but worth being conscious of it -- especially when updating the contact record -- though we already do this for campaign_contact.message_status, so it's just bundling that further.

schuyler1d:
# Fixes #1018

## Summary
Many errors come back from Twilio and other message services that imply issues with the phone number pool, or carrier ratelimiting (sometimes based on domain) and issues with the number type (disconnected or landline).  All these are currently just 'lost'.  This PR adds error_code to message, log, and campaign_contact field to track message errors and make it easier to track errors better.  It also removes message.service_response which fills a lot of db space and probably makes saving/querying harder.

## Details
* adds message.error_code, campaign_contact.error_code, log.{error_code, from_num, to_num} -- to track error codes and make them more easily queryable and reportable per-campaign (reporting is the ultimate goal for adding it to campaign_contact -- which will be the last error_code that was received)
* removes `message.service_response` which was a blob-append of json data of the entire requests.  This fills up the database and probably causes some issues with performance on the `message` table (even if just during retrieval bandwidth)
* The one operational use of message.service_response was to count how many errors/tries were used when sending a message (and thus allowing retries by not updating the message status away from 'SENDING' -- We instead overload the error_code field and use values `-1`-`-6` for message send failures (mostly due to network errors)
* Tests are added around postMessageSend behavior for error_code scenarios.

## Future work (after merging this)
* Reporting error counts per-campaign and maybe per number/region
* Possibly doing specific behaviors based on specific error codes (e.g. putting landline numbers in a separate table (optout might be too overloaded) to avoid other attempts)
* Mostly unrelated, but I also anticipate more schema changes on the `message` table to include the campaign_contact_id (and use that instead of cell+assignment_id for lookup) along with the messageservice_id for a particular twilio message service per-organization
* More tests around lib/twilio.js (would make a `good first issue`)

# Checklist:

- [na] I have manually tested my changes on desktop and mobile
  - excluding whitespace/linting changes that were triggered by `prettier`


docs/REFERENCE-environment_variables.md
```diff
@@ -1,81 +1,82 @@
...
+| DEV_APP_PORT                                     | Port for development Webpack server. Required for development.                                                                                                                                                                                                                                                                                                                              |
+| DISABLE_DB_LOG                                   | Disable saving error log info to `log` table                                                                                                                                                                                                                                                                                                                                                |
```
schuyler1d, line 27: this is the one-line change in this file -- the rest is changed by `prettier`

src/server/api/lib/twilio.js
```diff
@@ -194,72 +201,90 @@ async function sendMessage(message, contact, trx) {
...
+  if (hasError) {
+    if (
+      messageToSave.error_code < -MAX_SEND_ATTEMPTS &&
+      messageToSave.error_code > -100
+    ) {
+      messageToSave.send_status = "ERROR";
+    }
+    // decrement error code starting from zero
+    messageToSave.error_code = Number(response.error_code || 0) - 1;
+    let contactUpdateQuery = Promise.resolve(1);
```
schuyler1d, line 247: `postMessageSend` is mostly just pulled out of `sendMessage` for testing purposes.  There are a couple changes:
1. message.service_response refs are removed
2. message.error_code references are added along with logic that accomodates a similar use of service_response which sees how many times we try to send before giving up.
3. `contactUpdateQuery` updates campaign_contact.error_code with the error state.

jeffm2001:
This will be really helpful. I was using some ugly queries to extract error codes from the log table. One note, the other thing I've been extracting from the `message.service_response` blob is media urls — which I pipe out to a CSV and mass download with wget.

I know there's a PR somewhere that deals with handling MMS responses, so this probably isn't a big deal, and handling multimedia responses in the app will be better all around. Just wanted to flag, since it might be a good idea to double check that there's nothing else useful in the service_response that is not being captured.

strangeways (dismissed):
👍

schuyler1d:
> This will be really helpful. I was using some ugly queries to extract error codes from the log table. One note, the other thing I've been extracting from the `message.service_response` blob is media urls — which I pipe out to a CSV and mass download with wget.

Interesting, I guess the issue with the current PR here is if you still need the MediaUrl, then this drops it completely.  I had generally envisioned MediaUrl getting added to the `message.text` field in some formatted way.  Maybe what I'll do is put it in the `log` table, which you can use instead.  I'm loathe to add more text fields to the `message` table, as it gets quite large.

jeffm2001:
Putting it in `log` makes sense to me. Probably also fine to punt on it for now and figure it out in #663

schuyler1d:
> Putting it in `log` makes sense to me. Probably also fine to punt on it for now and figure it out in #663

Cool.  The recent commit will allow you to query `FROM log WHERE error_code = -101` for (incoming) media urls.

strangeways (dismissed):

strangeways (dismissed):

docs/REFERENCE-environment_variables.md
ibrand: Why would we want this as an env var? Curious why anyone would want to turn this feature off

schuyler1d: This adds a lot of data (and writes) into the database that is arguably better examined in actual log data (vs. a db table).  I think at some point we might even want this as a default, but Jeff's use-case with MediaUrls should probably be addressed in some way or would be an exception (we could skip the condition based on DISABLE_DB_LOG when the MediaUrl is present)

ibrand (dismissed):

Issue #448, opened 2017-12-20
Campaign primary color doesn't require enough (any) contrast with text color


sreynen:
As an extreme example, it's possible to set the primary color to exactly the same as the text color, making the text totally unreadable:

<img width="398" alt="screen shot 2017-12-20 at 8 53 37 am" src="https://user-images.githubusercontent.com/108645/34215671-6882079a-e563-11e7-994c-237afdfe3d50.png">

But more likely, less extreme low-contrast color options are bad for accessibility. See: https://www.w3.org/TR/WCAG20/#visual-audio-contrast-contrast

ibrand:
@rifkegribenes and @sreynen  I'm cleaning up our `good first issues`s and want to know if this one was completed (since it's pretty old.) @rifkegribenes did you ever find an answer to your question?

SAnschutz:
I'd love to work on this issue if no one else has claimed it yet!

ibrand:
Awesome @SAnschutz ! I added you as the assigned person for this issue!

ibrand:
This was completed and merged!

PR #1277, opened 2019-11-12
Address color contrast for user-selected primary color

SAnschutz:
Fixes issue #448 

Added code to detect the brightness of the user-selected background color, then render the text in either white or dark gray, to provide enough contrast so the text is legible and for improved accessibility.

Before: 
Screen Shot 2019-11-11 at 10 33 33 PM
 
After: 
Screen Shot 2019-11-11 at 10 35 37 PM

- [ ] [My change is 300 lines of code or less]

### DIFF
--- a/src/components/AssignmentSummary.jsx
+++ b/src/components/AssignmentSummary.jsx
@@ -10,6 +10,7 @@ import moment from "moment";
 import Divider from "material-ui/Divider";
 import { withRouter } from "react-router";
 import { dataTest } from "../lib/attributes";
+import theme from "../styles/theme";
 
 const inlineStyles = {
   badge: {
@@ -106,7 +107,7 @@ export class AssignmentSummary extends Component {
     }
   }
 
-  detectHexBrightness(color) {
+  calcHexBrightness(color) {
     let hexcolor = color.replace("#", "");
     if (hexcolor.length === 3) {
       hexcolor = hexcolor.replace(/(.)/g, "$1$1");
@@ -118,10 +119,12 @@ export class AssignmentSummary extends Component {
     const brightness = Math.floor((r * 299 + g * 587 + b * 114) / 1000);
     const threshhold = 130;
 
-    return brightness > threshhold ? "light" : "dark";
+    const contrastingColor =
+      brightness > threshhold ? theme.colors.darkGray : theme.colors.white;
+    return contrastingColor;
   }
 
-  detectRgbBrightness(color) {
+  calcRgbBrightness(color) {
     const colorValuesArray = color
       .slice(color.indexOf("(") + 1, color.indexOf(")"))
       .split(",");
@@ -136,10 +139,14 @@ export class AssignmentSummary extends Component {
     const brightness = Math.floor((r * 299 + g * 587 + b * 114) / 1000);
     const threshhold = 130;
 
-    return brightness > threshhold || alpha < 50 ? "light" : "dark";
+    const contrastingColor =
+      brightness > threshhold || alpha < 50
+        ? theme.colors.darkGray
+        : theme.colors.white;
+    return contrastingColor;
   }
 
-  detectHslBrightness(color) {
+  calcHslBrightness(color) {
     let colorValuesArray = color
       .slice(color.indexOf("(") + 1, color.indexOf(")"))
       .replace(/(%)/g, "")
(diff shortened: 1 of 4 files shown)


src/components/AssignmentSummary.jsx
```diff
@@ -106,6 +107,77 @@ export class AssignmentSummary extends Component {
...
+  calcHexBrightness(color) {
+    let hexcolor = color.replace("#", "");
+    if (hexcolor.length === 3) {
+      hexcolor = hexcolor.replace(/(.)/g, "$1$1");
+    }
+    const r = parseInt(hexcolor.substr(0, 2), 16);
+    const g = parseInt(hexcolor.substr(2, 2), 16);
+    const b = parseInt(hexcolor.substr(4, 2), 16);
+
+    const brightness = Math.floor((r * 299 + g * 587 + b * 114) / 1000);
```
ibrand: What are the numbers `299`, `587` and `114` used for? Could those be variables instead to make it clearer why we're doing that multiplication?

src/components/AssignmentSummary.jsx
```diff
@@ -106,6 +107,77 @@ export class AssignmentSummary extends Component {
...
+      .split(",");
+
+    const r = parseInt(colorValuesArray[0]);
+    const g = parseInt(colorValuesArray[1]);
+    const b = parseInt(colorValuesArray[2]);
+    const alpha = colorValuesArray[3]
+      ? parseInt(colorValuesArray[3] * 100)
+      : 100;
+
+    const brightness = Math.floor((r * 299 + g * 587 + b * 114) / 1000);
```
ibrand: Same as the above comment. Makes me wonder if it would make sense to combine these two functions into one with a switch case so that the variables would only need to be declared once? But that is totally up to you :)

ibrand (changes requested):
A couple of readability ideas, but great progress here! The only comment here blocking approval is around those numbers that I was asking for clarity on.

I could also see this being extracted into a file with a name like `color-contrast-helper.js` used really similarly to how we use `zip-format.js`. I think that would be nice and allow us to use color contrasting in other parts of the app if we needed it, but that's totally up to you if it's a change that you think makes sense.

Thank you for your work on this! 🎉

SAnschutz:
Thanks for these comments!  I've made some revisions -- happy to make more if necessary.

The biggest change I made was extracting out the color-contrast functionality into a separate file (and writing tests for that file.  :-) )  I also reformatted to utilize a switch statement, as you suggested -- I definitely like the way that reads better!

As for the numbers you asked for clarity on, that's just a formula for evaluating color that came from W3C (https://www.w3.org/TR/AERT/#color-contrast)  I must admit the numbers and calculations themselves are a little beyond my understanding of color theory and the manipulation of numbers on the color wheel. :-/  What I did to try to address your concern was to clearly label it as a formula (via variable name, which also has an indication of where it comes from).  Is that any more readable/acceptable?  If I still need to delve deeper into that, feel free to let me know!

schuyler1d:
I appreciate this work which will make the UI much better!

I think we might be able to simplify this by a lot.  I might not know the subtleties of the W3CBrightness function, so if my suggestion doesn't makes sense, then please just document the brightness formula with some comments, but I wonder if we can just use the `L` in HSL, which is, after all, Lightness (separated from hue and saturation) -- Instead of converting HSL => RGB and then calculating the foreground color from the RGB, can we just split 3 ways between different Lightness values (white, grey, black)?

Also, even if we can't, it looks like [tinycolor2](https://www.npmjs.com/package/tinycolor2#getbrightness) is a pre-existing subdependency of the color-picker already present -- let's use that library's conversion/brightness methods instead of writing our own.

DUR: it looks like TinyColor has a getBrightness method -- so let's use that and/or the Lightness value.

SAnschutz:
Thanks for your comments!  Yeah, in my first attempt I had written this using just the L value (and had also considered using some combination of S and L).  Thing is, in testing it a bit more, it isn't particularly accurate, because different hues have different perceived brightness...a dark yellow is still going to appear lighter than a dark red or a dark blue, regardless of the lightness value.

(For example: If you compare yellow hsl(60, 100%, 50%) with blue hsl(250, 100%, 50%) -- identical lightness and saturation values, but the yellow appears light enough you'd probably want dark text against it, and the blue appears dark enough you'd probably want white text against it.)

So I fleshed out the formula for HSL in this commit because I felt like I was being lazy by not addressing that!  But maybe in this case using just the L value falls into the category of "close enough"?  I'm certainly happy to go back to my former version that uses just the L value.

Or I'm happy to just use TinyColor!  I hadn't found that as a pre-existing subdependency until you pointed it out -- that probably makes the most sense.  I'll take a look at getBrightness and see what I can do.

SAnschutz:
OK, I just pushed a refactored version which utilizes tinycolor2.  Let me know if you think this looks better, @schuyler1d ?

In addition to getBrightness as you suggested, I also used getAlpha, since getBrightness doesn't seem to account for that value, and isValid to handle invalid manual inputs. The test file has also been updated accordingly.

Please forgive what I'm sure is a blatantly newbie question, but I honestly don't know if tinycolor2 should also be added to the list of dependencies in package.json?  I presume not, since it will automatically be installed as a sub-dependency of material-ui-color-picker?  But is there a reason to list it...in case, say, down the line we'd decide to change to a different color-picker, at which point this file would probably break?

schuyler1d (approved):
:+1:

schuyler1d:
> Please forgive what I'm sure is a blatantly newbie question, but I honestly don't know if tinycolor2 should also be added to the list of dependencies in package.json? I presume not, since it will automatically be installed as a sub-dependency of material-ui-color-picker? But is there a reason to list it...in case, say, down the line we'd decide to change to a different color-picker, at which point this file would probably break?

@SAnschutz -- yeah, I agree, let's add it to package.json explicitly.  Much simpler, thank you!

ibrand (approved):

Issue #1198, opened 2019-08-14
Bug: Spoke Not Scrubbing Opt Outs From Larger Organizations


shakalee14:
**Describe the bug**
When I upload a list with opt outs into an organization with a large amount of contacts, Spoke will upload the list and create new campaign contacts for the opted out contacts. Spoke will not allow texters to send texts to the opted contacts, but texters will see these contacts in their todo list. This causes confusion since these contacts are not contactable.

**To Reproduce**
1. Go to an instance with an organization with a lot of contacts (cannot reproduce on a smaller organization, within the same instance)
2. Click on create a campaign, and upload a csv with opted out contacts. 
3. Click 'upload contacts'
4. In our instance - we see an unfinished progress bar, and all of the contacts count appear in the texter assignment admin panel. When we click 'cancel job', it turns the contact section green, but doesn't scrub them out.
5. Assign texts to texter - login to texter account - and attempt to send texts (they shouldnt send, and these contacts should be still marked as 'needsMessage' in the `campaign_contact` table.

**Expected behavior**
Opted out contacts should be scrubbed out on the initial upload and the number of deleted should appear in a message.

**Screenshots**
The initial scrub  should report here:
<img width="829" alt="Screen Shot 2019-08-14 at 2 56 30 PM" src="https://user-images.githubusercontent.com/16676323/63048090-c70d6000-bea3-11e9-8e92-821ebfb01e13.png">

**Desktop (please complete the following information):**
 - OS: n/a
 - Browser n/a
 - Version n/a

**Smartphone (please complete the following information):**
 - Device: n/a
 - OS: n/a
 - Browser n/a
 - Version n/a

**Additional context**
I cant reproduce this issue locally, on staging or in the same instance but in the test organization. This only occurs in our production organization.

schuyler1d:
Some guidance about code points:
Optouts should be excluded in this code: https://github.com/MoveOnOrg/Spoke/blob/main/src/workers/jobs.js#L262-L269
You'll notice that this calls [getOptoutSubquery which is defined further up in the file](https://github.com/MoveOnOrg/Spoke/blob/main/src/workers/jobs.js#L52-L84) -- further that it and several optout functions are duplicated (a dittographic error) -- which we should probably remove while fixing this bug.

The bug is likely in this area and if it's identified it should also be fixed here for datawarehouse loading:
https://github.com/MoveOnOrg/Spoke/blob/main/src/workers/jobs.js#L430-L440

[...]

PR #1288, opened 2019-11-27
WIP: Adding debug statements for bug #1198

strangeways:
# Fixes # 1198

## Description

Adding debug statement to help track down this bug.

# Checklist:

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -49,22 +49,6 @@ const defensivelyDeleteJob = async job => {
 
 const zipMemoization = {};
 let warehouseConnection = null;
-function optOutsByOrgId(orgId) {
-  return r.knex
-    .select("cell")
-    .from("opt_out")
-    .where("organization_id", orgId);
-}
-
-function optOutsByInstance() {
-  return r.knex.select("cell").from("opt_out");
-}
-
-function getOptOutSubQuery(orgId) {
-  return !!process.env.OPTOUTS_SHARE_ALL_ORGS
-    ? optOutsByInstance()
-    : optOutsByOrgId(orgId);
-}
 
 function optOutsByOrgId(orgId) {
   return r.knex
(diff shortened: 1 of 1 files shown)


src/workers/jobs.js
```diff
@@ -267,6 +271,7 @@ export async function uploadContacts(job) {
     .then(result => {
```
schuyler1d, line 333: I wonder if the next step is to add a `.catch(err => { console.log('DEBUG: uploadContacts error', campaignId, err)})`
after the .then() -- maybe the error is happening there and we're not getting it -- alternatively or in addtion to, we could try...catch the top await call to get the count.

strangeways: Okay, added some error catching. Hopefully it makes sense!

src/workers/jobs.js
```diff
@@ -254,15 +238,37 @@ export async function uploadContacts(job) {
...
-  const optOutCellCount = await r
+  const optOutCellResults = await r
     .knex("campaign_contact")
     .whereIn("cell", function optouts() {
       this.select("cell")
         .from("opt_out")
         .where("organization_id", campaign.organization_id);
     });
-  console.log('DEBUG: uploadContacts 3', campaignId, optOutCellCount);
+  const optOutCellCount = optOutCellResults.length;
```
schuyler1d: This seems a bit risky -- if optOutCellResults is undefined/null then this will error.
There's a better pattern for just getting the count anyway (for the final version):
[r.getCount](https://github.com/MoveOnOrg/Spoke/blob/main/docs/EXPLANATION-development-guidelines.md#db-queries)

strangeways: Thanks for the pointer! I've updated it to use `await r.getCount` instead.

strangeways:
Superseded by PR https://github.com/MoveOnOrg/Spoke/pull/1348

PR #1289, opened 2019-11-27
Update the bug template label

ibrand:
Update the bug template label to match the new bug label because we changed the label from `bug` to `C-bug`


azuzunaga (approved):
👍

PR #1290, opened 2019-11-27
Update feature label in feature issue template

ibrand:
Match the new `C-enhancement` label in the feature issue template. This way issues opened as features will automatically label themselves as `C-enhancement`


azuzunaga (approved):
👍

PR #1293, opened 2019-12-04
Fix opt-out from message review UI

### LINKED ISSUE #1291: Bug: OPT OUT button in Message Review not working
When using message review to see texts that still need a response from a texters, when an admin clicks "view conversation" the admin has the ability to opt out the contact. This button is not working.

EXAMPLE: 
1. Enter 'Message Review"
2. Under "Contact Message Status" select "Needs Texter Response"
3. Under "campaign, enter ID "8189"
4. Select "View Conversation" for any of the open conversations that populate
5. In "View Conversation" mode, try clicking "Opt Out"

lperson:
# Fixes #1291

## Description

This PR provides the ability for message-review users to opt out a contact from message review.  From the UI perspective, this involves passing organization ID to the createOptOut mutation.

Two components were affected.  In addition, I modified components from which these two components were instantiated to pass additional props as necessary, added tests, and did some minor refactoring.

This PR is longer than 300 lines because I added a lot of test code.  Also, we moved a couple of files based on [this comment](https://github.com/MoveOnOrg/Spoke/pull/1293#pullrequestreview-333327222).

Although tests pass, attempting to test opt-out from the UI won't work unless #1297 is merged.  This PR, and #1297, should be merged and tested at the same time.

**ConversationPreviewModal**
* added organization id to the createOptOut mutation
* moved the createOptOut mutation to src/lib/client/createOptOut.js so we can use it from multiple places
* fixed handling when the createOptOut mutation raises an error
* added tests

**AssignmentTexterContact**
* added organization id to the createOptOut mutation
* moved the createOptOut mutation to src/lib/client/createOptOut.js so we can use it from multiple places
* added tests

# Checklist:

- [N/A] ~If my change is a UI change, I have attached a screenshot to the description section of this pull request~ (the UI didn't change)


src/components/IncomingMessageList/ConversationPreviewModal.jsx
```diff
@@ -161,34 +179,18 @@ class ConversationPreviewModal extends Component {
...
-    variables: {
-      optOut,
-      campaignContactId
-    }
-  })
+  createOptOut: createOptOutMutation
 });
 
-export default loadData(wrapMutations(ConversationPreviewModal), {
+export default loadData(wrapMutations(_ConversationPreviewModal), {
```
schuyler1d: if this item loads data, it should be a container and in the src/containers directory.  Please move.

src/components/IncomingMessageList/ConversationPreviewModal.jsx
```diff
@@ -92,7 +101,7 @@ ConversationPreviewBody.propTypes = {
   conversation: PropTypes.object
 };
 
-class ConversationPreviewModal extends Component {
+export class _ConversationPreviewModal extends Component {
```
schuyler1d: why the `_` prefix?  This seems like something to avoid, especially on a variable we're exporting.  If this is the only exported, then we should `export default ConversationPreviewModal` at the bottom and then imports can go by the file name.

schuyler1d (changes requested):
Can you merge #1297 in here, so we can see the tests passing together.  I understand this exceeds 300 lines, but more important to see tests passing and make sure the work can be merged independently.

src/server/api/errors.js
```diff
@@ -30,6 +30,21 @@ export async function accessRequired(
   }
 }
 
+export async function assignmentOrAdminRequired(
```
lperson: In the review for #1297 @schuyler1d commented:

> schuyler1d 3 days ago  Collaborator
> Can we rename this to assignmentOrAdminRoleRequired -- so it's not confused with is_superadmin

src/server/api/errors.js
```diff
@@ -30,6 +30,21 @@ export async function accessRequired(
...
+  user,
+  orgId,
+  assignmentId,
+  assignment
+) {
+  authRequired(user);
+  const isAdmin = await cacheableData.user.userHasRole(user, orgId, "ADMIN");
+  if (isAdmin || user.is_superAdmin) {
```
lperson: In their review for #1297, @schuyler1d commented:

> schuyler1d 3 days ago  Collaborator
> this needs to be user.is_superadmin (it's a db field, and isn't case-changed)
> 
> Alternatively, you can delete it and just test on isAdmin, since the function below also tests superadmin.

schuyler1d (approved):
:tada:

schuyler1d:
After deploying this PR as part of [stage-main](https://github.com/MoveOnOrg/Spoke/pull/1340) we received complaints from texters that while the opt-out message was sent to the contact, the opt_out record was not being saved (and thus they were able to revisit the conversation and continue texting them).

I haven't traced the bug (yet), but reporting the issue on the PR that's likely (but not guaranteed to be) related.

lperson:
I testing opting out a contact in stage-main.  A record was added to the opt-out table when a contact was opted out from texter reply as well as from message review.  So basically, I confirmed that it still works.  Is there more detail about how to reproduce what the testers observed?

schuyler1d:
Frustratingly, I haven't been able to reproduce.  There's a possibility it was happening mostly/more to certain texters rather than others (the one worst-affected was on Firefox, fwiw).

However, I observed even when it was failing that the createOptOut call was being made on the backend.  All texters both those that successfully did opt outs and those that didn't were all at TEXTER level.

I beleive @ibrand is going to organize some QA volunteers to try reproducing the issue soon.

lperson:
Do you know if there was anything in the server log?  Do you know if the opt outs were sent outside texting hours?

lperson:
Do you think it's possible something went awry in the caching layer?

src/server/api/schema.js
```diff
@@ -901,11 +906,15 @@ const rootMutations = {
 
     createOptOut: async (
       _,
-      { optOut, campaignContactId },
+      { organizationId, optOut, campaignContactId },
```
schuyler1d, line 909: While re-reviewing this after we encountered the optout (inconsistent) failure, I realized that we do `const campaign = await loaders.campaign.load(contact.campaign_id);` below this call.

It seems better to bring that call up higher, and then get organizationId from campaign.organization_id -- That way we avoid an edge-case privilege escalation in a multi-tenant situation, where an ADMIN of another organization could technically optout any contact from any other organization.

It also seems like it could reduce some of the complexity -- we don't need to add the organizationId to the api call, mutation, etc.

This isn't high priority relative to the bug, but if it helps to simplify the PR to reduce the bug surface area, then this seems like an opportunity.

ibrand:
Note on this that we had 2 QAs test with 400 test contacts on our staging instance and neither of them could repro this :/

ibrand:
@lperson @bdatkins wondering if either of you have seen any opt out failures on prod? Would you be able to run this query to determine if there are any opt outs that have slipped through the cracks for debugging?

```
SELECT u.first_name, u.last_name, c.assignment_id, max(o.created_at), min(m.id), count(*), sum(CASE WHEN o.id is null THEN 1 ELSE 0 END) as lostoptouts from message m JOIN campaign_contact c ON (m.campaign_contact_id = c.id) JOIN assignment a ON (c.assignment_id = a.id) JOIN public.user u ON (u.id = a.user_id) LEFT JOIN opt_out o ON (o.cell = m.contact_number) where m.id > 88199283 and text = '<YOUR DEFAULT OPT OUT TEXT HERE>.' group by 1,2,3 order by 4 DESC;
```

bdatkins:
@ibrand We have a fair number of optout failures, it could be combination of a code issue & texters not doing the right thing. (Note to self, need to clean them up)

src/server/api/errors.js
```diff
@@ -30,6 +30,21 @@ export async function accessRequired(
...
+  assignmentId,
+  assignment
+) {
+  authRequired(user);
+  const isAdmin = await cacheableData.user.userHasRole(user, orgId, "ADMIN");
+  if (isAdmin || user.is_superadmin) {
+    return true;
+  }
+
+  return assignmentRequired(user, assignmentId, assignment);
```
schuyler1d, line 45: should this be awaited?

lperson:
Closed this in favor of #1412.

PR #1295, opened 2019-12-07
AWS async

schuyler1d:
# Fixes aws database migration command was failing

## Description

AWS Lambda has changed processing models from 6.10, but `./lambda.js` has not been updated since then.  When running the aws database migration command there were database failure/connection errors. This change removes the connection error issues and runs the command correctly. It hopefully also puts us on better footing to match the current AWS Lambda node 10.x process model.

# Checklist:

- [na] I have added tests that prove my fix is effective or that my feature works


lambda.js
```diff
@@ -53,7 +54,12 @@ exports.handler = (event, context, handleCallback) => {
     invocationEvent = event;
     invocationContext = context;
     cleanHeaders(event);
-    const webResponse = awsServerlessExpress.proxy(server, event, context);
+    const webResponse = awsServerlessExpress.proxy(
+      server,
+      event,
+      context,
+      "PROMISE"
```
ibrand, line 61: Why are we passing a string called `"PROMISE"` here? It looks like we never use it? Is it something specific and internal to AWS?

ibrand (approved):
Looks good, one question in there though!

lambda.js
schuyler1d: Got it from here:
https://github.com/awslabs/aws-serverless-express/issues/234#issuecomment-492529316
(or a similar place in aws-serverless-express documentation that I have since lost).
Basically, it has a weird API, but that's how you 'ask' for a promise-based call.

ibrand: 👍

PR #1308, opened 2019-12-16
guard against passport auth switch

schuyler1d:
# Fixes switching from auth0 => local auth with a local cookie
guard against NaN content, which e.g. could be from previous auth0 auth, but could also be used for testing guesses of decrypted values (security issue)

## Description

When a server was originally using auth0 and then switches to `PASSPORT_STRATEGY=local` people who have visited and logged in to the site would visit, the server would respond with 'Error on server' -- and the backend would complain that "NaN" is an invalid SQL value.  This is from the changed lines in the PR, where `id` was something like `auth0|1234abcdef678` -- parseInt would fail and then NaN would be passed as an integer arg to the database.

This PR

# Checklist:

### DIFF
--- a/__test__/server/auth-passport.test.js
+++ b/__test__/server/auth-passport.test.js
@@ -1,0 +1,61 @@
+import passport from "passport";
+import passportSetup from "../../src/server/auth-passport";
+
+import {
+  setupTest,
+  cleanupTest,
+  createUser,
+  createOrganization
+} from "../test_helpers";
+
+beforeEach(async () => {
+  await setupTest();
+}, global.DATABASE_SETUP_TEARDOWN_TIMEOUT);
+
+afterEach(async () => {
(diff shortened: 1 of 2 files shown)


ibrand (approved):
Awesome, and nice thorough tests too!

joemcl:
OK @ibrand @schuyler1d I clicked View deployment, above. I created a new user and a new org. I was able to log in successfully. I created a dummy campaign. So far so good. 
But it seems like the only way to actually test that this works as desired is to hook up Auth0 to it, create users, then switch the `PASSPORT_STRATEGY=` value to `local`.

ibrand:
@joemcl that makes sense.. this is a hard one to QA through the dummy deploy. I reviewed the code and so we're good to merge this and we'll be able to test it on our staging instance. Thank you!

PR #1315, opened 2019-12-21
Allow required headers to be snake_case in the contacts csv

### LINKED ISSUE #1272: Feature Request: allow csv required headers to be first_name, etc instead of firstName...
Many csv generators from CRMs, etc create headers list `first_name`, `last_name` and some do not support uppercase characters at all.  In order to avoid the need to manually edit csv headers before upload.

**Describe the solution you'd like**
Instead of blocking CSV uploads that have first_name, last_name instead of firstName, etc columns, either header-style should succeed in uploads.

This is entirely on the front-end -- the related code is here:
https://github.com/MoveOnOrg/Spoke/blob/main/src/lib/index.js#L127

Code will probably need to change the validation part a bit and then the function at the bottom `convertRowToContact`
And then, of course, we'll need to change the documentation in src/components/CampaignContactsForm.jsx

spoke_uploadcontacts

lperson:
# Fixes #1272 

## Description

Accept and correctly handle contacts csv files with `first_name` instead of `firstName` and `last_name` instead of `lastName` for their respective column headers.

Updated instructions in contact upload:

Screen Shot 2019-12-21 at 2 17 23 PM

What changed: upgraded papa-parse to version 5 and provided a `transformHeaders` method to implement this.

# Checklist:

- N/A ~[ ] I have made any necessary changes to the documentation~
- N/A ~[ ] My PR is labeled [WIP] if it is in progress~

### DIFF
--- a/src/lib/parse_csv.js
+++ b/src/lib/parse_csv.js
@@ -1,0 +1,101 @@
+import Papa from "papaparse";
+import _ from "lodash";
+import { getFormattedPhoneNumber, getFormattedZip } from "../lib";
+
+const requiredUploadFields = ["firstName", "lastName", "cell"];
+const topLevelUploadFields = [
+  "firstName",
+  "lastName",
+  "cell",
+  "zip",
+  "external_id"
+];
+
+const getValidatedData = (data, optOuts) => {
+  const optOutCells = optOuts.map(optOut => optOut.cell);
+  let validatedData;
+  let result;
+  // For some reason destructuring is not working here
+  result = _.partition(data, row => !!row.cell);
+  validatedData = result[0];
+  const missingCellRows = result[1];
+
+  validatedData = _.map(validatedData, row =>
+    _.extend(row, {
+      cell: getFormattedPhoneNumber(
+        row.cell,
+        process.env.PHONE_NUMBER_COUNTRY || "US"
+      )
+    })
+  );
+  result = _.partition(validatedData, row => !!row.cell);
+  validatedData = result[0];
+  const invalidCellRows = result[1];
+
+  const count = validatedData.length;
+  validatedData = _.uniqBy(validatedData, row => row.cell);
+  const dupeCount = count - validatedData.length;
+
+  result = _.partition(
+    validatedData,
+    row => optOutCells.indexOf(row.cell) === -1
+  );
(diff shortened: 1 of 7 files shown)


src/lib/parse_csv.js
```diff
@@ -0,0 +1,115 @@
...
+    return camelizedColumnHeader;
+  }
+
+  return columnHeader;
+};
+
+export const parseCSV = (file, optOuts, callback) => {
+  Papa.parse(file, {
+    header: true,
+    transformHeader: ensureCamelCaseRequiredHeaders,
```
schuyler1d, line 92: INFO (no critique): This `transformHeader` option and the ensureCamelCaseRequiredHeaders seems to be the new code, aside from the code movement into a separate file.

src/lib/parse_csv.js
```diff
@@ -0,0 +1,115 @@
...
+      optOutCount: optOutRows.length,
+      invalidCellCount: invalidCellRows.length,
+      missingCellCount: missingCellRows.length,
+      zipCount
+    }
+  };
+};
+
+const ensureCamelCaseRequiredHeaders = columnHeader => {
+  const camelizedColumnHeader = humps.camelize(columnHeader);
```
schuyler1d, line 77: INFO (no critique): humps is also already used in e.g. src/server/api/lib/utils.js so is already a dependency, though this probably adds the library to the frontend now.

src/lib/parse_csv.js
```diff
@@ -0,0 +1,115 @@
...
+
+  if (
```
schuyler1d, line 79: Can you add a comment here that just says something like "This preserves the value of `vendor_id` and custom fields/columns"

schuyler1d (dismissed):
:+1: 
This is great!  Made one request to add a comment, but either way, let's merge.

filafb (dismissed):

src/lib/parse_csv.js
lperson: @schuyler1d I added the comment.

src/lib/parse_csv.js
lperson: Noted!  Thanks!

schuyler1d (approved):
:+1:

Issue #1016, opened 2018-11-07
Might automatic opt outs be possible?


suegibson:
When a contact replies with single word STOP, can it be arranged for Spoke to automatically opt out?
@lperson @joemcl @codygordon

DinahSanders:
Recommendation: trigger only on replies which consist only of the single word 'stop'.

(We can't count on all-caps as a signifier because sometimes we get replies like "Count me in! We have got to STOP this terrible bill from passing!!")

suegibson:
> Recommendation: trigger only on replies which consist only of the single word 'stop'.
> 
> (We can't count on all-caps as a signifier because sometimes we get replies like "Count me in! We have got to STOP this terrible bill from passing!!")

suegibson:
Agreed!

bdatkins:
@suegibson Followup question - should we silently opt them out?

bdatkins:
**I'm grabbing this issue to work on!**  Opting folks out automatically if they use just STOP will increase our data quality and decrease the possibility of annoying those we text!  Especially for new texters, they don't always do the right thing and opt folks out as they should.

DinahSanders:
To @bdatkins's question, I say: No, for this we should send a confirmation.

Our organization's policy is:

> We always send the 'Opt Out' message _except_ under three conditions:
>>The contact mentions texts costing them money;
>>The contact explicitly says, “Do not EVER contact me again”;
>>The contact explicitly says, “Don't text back”.

So, if we're triggering only off the single word 'stop' those don't apply.

Suggested default message: "I'm opting you out of our texts immediately. Have a great day."

bdatkins:
Thanks for that feedback @DinahSanders.   The configured Opt out message be used (either default or the one set at the Organization level). 

I can see starting with a plain 'STOP' and perhaps making it more elegant in a later change, possibly dovetailing with the Profanity check work. (https://github.com/MoveOnOrg/Spoke/pull/938)

suegibson:
Yes, opt out without sending confirmation message, please, @bdatkins. Thank you for thinking to ask!

ibrand:
@bdatkins @suegibson I'm cleaning up our good first issuess and want to know if this one was completed (since it's pretty old) or if it's open for new contributors to claim!

bdatkins:
@ibrand This has not been completed, so someone can grab it!

joemcl:
[...]

Issue #359, opened 2017-10-27
flagged words/profanity should put incoming messages in a review queue


schuyler1d:
Texter volunteers shouldn't need to see responses with slurs or other offensive language.
Before a texter sees the reply, if a message matches the blockword regex (a configured environment variable), then it should not be shown to the texter and the texter should not get the contact (yet) in the list of replies. Instead, the contact and message should be available to an campaign ADMIN where they can choose three 'fates' for the contact/message:
1. optout the contact (with reason 'profane')
2. leave the contact's message in the profane queue but do nothing
3. allow the message (in which case the message gets passed to the contact as a normal reply)

Furthermore, when a texter *sends* a message that matches the blockword regex, the server should respond with an error message and the texter should be prompted to 'rephrase their message'.  -- i.e. texter messages that match the blockword regex should be blocked from being saved or sent.

 - [ ] STOPs (see url for other words) texts should auto-optout (for compliance with US texting laws/best-practices. Note: by default Twilio does this already, but accounts can request turning this feature off, so you can track which contacts requested optouts, for e.g. syncing optouts to other systems. STOP/STARTing should follow the behavior described here: https://support.twilio.com/hc/en-us/articles/223134027-Twilio-support-for-opt-out-keywords-SMS-STOP-filtering-
 - [ ] A configurable regex string to find bad words. 
 - [ ] Ideally the regex string would be transliterated so people looking at the config wouldn't need to see a bunch of bad words. ( maybe something like flipping a=>z and z=>a `regex_string.replace(/[a-z]/g, ((x) => (String.fromCharCode('a'.charCodeAt()+'z'.charCodeAt()-x.charCodeAt()))))`
 - [ ] A command (e.g. `npm run profaneconfig ...` to convert a regular regex to a transliterated regex.
 - [ ] Consider adding a `message.send_status` of PROFANE -- depending on consequences.  Pros: might exclude in a good way from messages shown in-thread. Cons: might 
 - [ ] STOP requests should get a specific 'reason_code' in the opt-out table (say `stop`). and profane messages should get `profane` in the opt_out.reason_code.  
 - [ ] add profane field marker to allmessages api (https://github.com/MoveOnOrg/Spoke/blob/main/src/server/index.js#L127) so that contacts that got marked 'profane optout' are visible (will need to add data out after joining to unless we change the send_status as mentioned above)

Open questions:
* Should we have a new send_status?

schuyler1d:
[...]

PR #1319, opened 2019-12-23
Incoming message handler framework and refactoring

### LINKED ISSUE #1317: RFC: incoming message handlers
## Proposal:
Build a lightweight framework that allows adding incoming message processors.  An incoming message processor allows a Spoke deployment to add optional processing for each incoming message.

## Motivation:
There are at least three use cases for this that I am aware of.
* **Profanity filter** -- if a message from a contact contains profanity or abusive language, reassign the conversation to an administrator or apply a tag. See #359.
* **Automatic opt out** -- if a message from a contact contains the word STOP, opt out the contact. See #1016.
* **Automatic tagging** -- if a message appears to be in Spanish, apply a tag (for example).

## Detailed Design:
* Allow a list of incoming message handlers to be specified in the MESSAGE_HANDLERS environment variable.  If that variable is not specified it will default to `['save-message']`.
* `save-message` is the existing functionality in `src/server/api/lib/message-sending.js saveIncomingMessage`.
* The server will dynamically load each of the specified handlers, which must be placed in the `src/server/message_handlers` directory.  
* File names must match the names of the handlers  specified in MESSAGE_HANDLERS with a `.js` extension.  
* Each file must export a function called `processMessage` with takes the new incoming message instance and returns the message instance or something falsey.  
* If a handler returns something falsey, no other handles are called.  
* A handler can return a modified copy of the new incoming message if its function is to do some transformation on incoming messages.

## Additional Context:
A handler can do things other than returning something falsey or transforming an incoming message.  For example, a handler could reassign the contact associated with the incoming message, or a handler could apply a tag to an incoming message.

lperson:
# Fixes #1317 

## Description

This is a framework to allow modular and optional handling of incoming messages.  Applications include profanity filter (#359), automatic opt-out (#1016), and automatic tagging.

This PR refactors message saving (previously in `the saveIncomingMessage` function in `src/server/api/lib/message-sending.js saveIncomingMessage`) into an incoming message handler called `save-message`.

This PR also introduces a throw-away incoming message handler called `example` for purposes of illustration.

A Spoke installation can specify incoming message handlers to use by providing a list in the `MESSAGE_HANDLERS` environment variable.  If that variable is not specified it will default to `["save-message"]`. Incoming message handlers will be processed in the order specified in `MESSAGE_HANDLERS`.

Incoming message handlers must exist in the `src/server/incoming_message_handlers` directory.  Each file must have a name corresponding to the name of one of the handlers listed in `MESSAGE_HANDLERS` with the extension `.js`.  Each file must export `processMessage`, a function taking one parameter, the incoming message object, and returning a message object (possibly transformed) or something falsey.

If a handler returns something falsey, no further handlers will be executed.

The app will crash at startup if MESSAGE_HANDLERS specifies a message handler that doesn't exist or exists and doesn't export the `processMessage` function.

# Checklist:

- N/A ~If my change is a UI change, I have attached a screenshot to the description section of this pull request~


schuyler1d:
So I generally like the structure and flexibility of this PR -- i.e.making save-message a handler that can be before and after save-message.  I wonder if we can use the framework though for each file to mark whether it needs to be before or after save-message so it can also be enforced at the code level.

This PR is going to interact with my [message.campaign_contact_id PR](https://github.com/MoveOnOrg/Spoke/blob/00b0e1825b95e1d4a9db78c8a5a926693bf1b29f/src/server/api/lib/twilio.js#L56-L73) and create some challenging merging.  I think the most interesting part to think through is where we run getLastMessage.  It seems like historically, it's been somewhat hackily in convertMessagePartsToMessage (at least in twilio).

Looking at your code it looks like that will still run before incoming message handlers, which I think is the right call -- i.e. when these two PRs are together, we'll also know the campaign_contact_id of the message.  Some of these handlers, might need to lookup the contact data and/or the organization -- it would be good if there were a way to pass/keep the context loaded from each, so subsequent needs of the same data would only be done once rather than for each handler.

schuyler1d:
this work was merged in-spirit in v7.0 and was foundational to MESSAGE_HANDLERS in that release.

PR #1346, opened 2019-12-31
add job to load messages in from a csv and fix some crucial bugs in twilio.js

schuyler1d:
# Fixes # (issue)

## Description
This adds a method to run `yarn process-message-csv SOMEMESSAGEFILE.csv` on a file and load them message responses into the database all at once -- this is a useful tool when processing messages from outside the twilio api and/or for testing.

In writing this, some critical bugs were found in the current twilio processing that may be causing deadlocks on indexes and other problems.  This fixes those minimally, awaiting more fixes in e.g. 

# Checklist:

### DIFF
--- a/src/server/api/lib/message-sending.js
+++ b/src/server/api/lib/message-sending.js
@@ -18,15 +18,15 @@ export async function getLastMessage({ contactNumber, service }) {
 
 export async function saveNewIncomingMessage(messageInstance) {
   if (messageInstance.service_id) {
-    const [countResult] = await r
+    const [duplicateMessage] = await r
       .knex("message")
       .where("service_id", messageInstance.service_id)
       .select("id")
       .limit(1);
-    if (countResult) {
-      console.error("DUPLICATE MESSAGE", countResult, messageInstance);
+    if (duplicateMessage) {
+      console.error("DUPLICATE MESSAGE", duplicateMessage, messageInstance);
+      return;
     }
-    return;
   }
   await messageInstance.save();
 
--- a/src/server/api/lib/message-sending.js
+++ b/src/server/api/lib/message-sending.js
@@ -18,16 +18,15 @@ export async function getLastMessage({ contactNumber, service }) {
 
 export async function saveNewIncomingMessage(messageInstance) {
   if (messageInstance.service_id) {
-    const countResult = await r.getCount(
-      r.knex("message").where("service_id", messageInstance.service_id)
-    );
+    const [countResult] = await r
+      .knex("message")
+      .where("service_id", messageInstance.service_id)
+      .select("id")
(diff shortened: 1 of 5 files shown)


src/server/api/lib/message-sending.js
```diff
@@ -18,16 +18,15 @@ export async function getLastMessage({ contactNumber, service }) {
 
 export async function saveNewIncomingMessage(messageInstance) {
   if (messageInstance.service_id) {
-    const countResult = await r.getCount(
-      r.knex("message").where("service_id", messageInstance.service_id)
-    );
+    const [countResult] = await r
```
schuyler1d: A theory about the failure was that this was getting a count rather than a single message -- it therefore deadlocked on two of the same messages coming in, because it didn't need to find just one, but get a full count.

src/server/api/lib/twilio.js
```diff
@@ -56,13 +56,16 @@ async function convertMessagePartsToMessage(messageParts) {
   const lastMessage = await getLastMessage({
     contactNumber
   });
+  if (!lastMessage) {
```
schuyler1d, line 59: Often SMS spam comes in.  If we failed to find a lastMessage, then this failed rather than returning and completing normally -- leaving a db connection open -- and possibly a transaction.

schuyler1d (commented):
Would like to test this out on production to see if it resolves our issues.

src/server/api/lib/message-sending.js
strangeways: This makes sense, though it would be good to rename this variable, since it is no longer the count (unless I'm misunderstanding the change).

src/server/api/lib/message-sending.js
```diff
@@ -18,16 +18,15 @@ export async function getLastMessage({ contactNumber, service }) {
...
+      .limit(1);
     if (countResult) {
-      console.error(
-        "DUPLICATE MESSAGE SAVED",
-        countResult.count,
-        messageInstance
-      );
+      console.error("DUPLICATE MESSAGE", countResult, messageInstance);
     }
+    return;
```
strangeways: Can you explain the `return` here? Is this meant to be inside the `if (countResult) {` block?

schuyler1d: dur, the `return` should have been inside the if-clause.  I've moved it inside and also renamed the variable duplicateMessage to be clearer.

strangeways (approved):

schuyler1d:
This was tested on MoveOn's production instance and solved a data-race issue for DB changes described in https://github.com/MoveOnOrg/Spoke/pull/1311#issuecomment-568786302

ibrand (approved):

Issue #1005, opened 2018-10-30
High priority re: reassigning 


suegibson:
When texters are making egregious errors, I need to be able to take all the conversations they have not yet replied in and leave them with only corrections to make. I can’t afford to let them have the opportunity to keep repeating their errors. 
@lperson @joemcl @codygordon

joemcl:
@suegibson I believe this can be accomplished already using the Needs Texter Response option and Reassign To option. You can reassign any convos to yourself or a different texter, taking those away from the original texter.

suegibson:
No, @joemcl, because some that need texter response already have errors in them. 
cc: @lperson

joemcl:
@suegibson just thinking about the 'workflow' for this.  Now that we can add tags/labels to a conversation in Message Review - and filter for convos with any or no tag -  if @lperson hardcodes a new tag/label  - called 'needs correction' maybe? - so you can tag/label/flag the convos you want to leave with the texter, then you could reassign all the *other* texter convos to yourself, or another texter. I think. Thoughts @suegibson @deasterdaywfp @lperson ? cc: @ibrand .

joemcl:
Just tested that workflow, and yeah, it works. Just need a new tag added.

suegibson:
I don’t see the ability to tag in message review.

joemcl:
@suegibson - yup - it's Assign Tag to Selected. 

Before:
![SpokeTagBefore](https://user-images.githubusercontent.com/3912434/64134860-911f1580-cd97-11e9-9454-f00baee8744c.jpeg)

After:
![SpokeAddTagAfter](https://user-images.githubusercontent.com/3912434/64134864-95e3c980-cd97-11e9-8e25-fd7e2f8fb491.jpeg)

suegibson:
I see. I was looking within a conversation, which would be handier.

joemcl:
> I see. I was looking within a conversation, which would be handier.

@suegibson agree, would be handier. @lperson whaddya think?

ibrand:
Hey @suegibson , @joemcl and @lperson this repo (MoveOn/main) doesn't have tags on it, so this solution doesn't work on. Should we rewrite this ticket to be a little more specific about a spec for solving this problem?

lperson:
[...]

PR #1350, opened 2020-01-07
Suspend texters

### LINKED ISSUE #1323: Suspended texter status
As part of RFC #1320 and the project [Bring work in WorkingFamilies/Spoke fork into MoveOnOrg/Spoke](https://github.com/MoveOnOrg/Spoke/projects/14), merge the work in these WorkingFamiles/Spoke PRs:
- [98](https://github.com/WorkingFamilies/Spoke/pull/98) 
- [136](https://github.com/WorkingFamilies/Spoke/pull/136)

This allows an administrator to mark a user as suspended in the people page and suspended users can't send initial texts or replies.

lperson:
# Fixes #1323 

### Motivation
WFP asked for a way to quickly stop testers from texting.  This may be necessary if they're not following instructions, ignoring responses, going off script, sending offensive messages, etc.  This may be something useful to other organizations as well, so I'm doing this PR to merge it into the MoveOn repository.

### What's in this PR
* Add `SUSPENDED` texter status.
* Admins can change a use's status to `SUSPENDED` on the people page.
* Testers with `SUSPENDED` status cannot send messages or opt out contacts.
  * When they log in they will land on `/app/<organizationId>/suspended`
  * If they become suspended while they are logged in they will be redirected to `/app/<organizationId>/suspended` (this will happen either when the UI tries to retrieve more contacts or when the user hits `send` before the UI had a chance to retrieve more contacts)

### This is `WIP` because 
- [ ] Tests are failing 😭 
- [ ] It needs lots of new tests.
- [ ] The way we determine that a user is suspended if they become suspended while they are logged in is too broad.  Right now we're assuming any GraphQL error means suspended.  We need to use a specific GraphQL error and only redirect to `/app/<organizationId>/suspended` if the error is happening because they're suspended.
- [ ] The suspended message should be customizable.  Organizations that don't have slack in their workflow will want a different message.

---

Screen Shot 2020-01-06 at 7 42 56 PM

---

### Possible enhancements
- Suspended texters should not be eligible for assignment.
- Indicate in message review that a texter is suspended when their name appears in filter by texter?
- Indicate in the message review grid that a texter is suspended?
- Indicate in texter assignment that a texter is suspended? (if we don't make them ineligible for assignment as noted above)
- When a texter is suspended, unassign them from all contacts? Or should that remain a manual step in message review?

# Checklist:

### DIFF
--- a/src/containers/AssignmentTexterContact.jsx
+++ b/src/containers/AssignmentTexterContact.jsx
@@ -407,7 +407,15 @@ export class AssignmentTexterContact extends React.Component {
       }
       this.setState({ disabled: true });
       console.log("sendMessage", contact.id);
-      await this.props.mutations.sendMessage(message, contact.id);
+      const sendMessageResult = await this.props.mutations.sendMessage(
+        message,
+        contact.id
+      );
+      if (sendMessageResult.errors && this.props.campaign.organization.id) {
+        this.props.router.push(
+          `/app/${this.props.campaign.organization.id}/suspended`
+        );
+      }
 
       await this.handleSubmitSurveys();
       this.props.onFinishContact(contact.id);
@@ -478,7 +486,15 @@ export class AssignmentTexterContact extends React.Component {
     this.setState({ disabled: true });
     try {
       if (optOutMessageText.length) {
-        await this.props.mutations.sendMessage(message, contact.id);
+        const sendMessageResult = await this.props.mutations.sendMessage(
+          message,
+          contact.id
+        );
+        if (sendMessageResult.errors && this.props.campaign.organization.id) {
+          this.props.router.push(
+            `/app/${this.props.campaign.organization.id}/suspended`
+          );
+        }
       }
 
       const optOut = {
--- a/src/server/api/errors.js
+++ b/src/server/api/errors.js
@@ -56,6 +56,28 @@ export async function assignmentRequired(user, assignmentId, assignment) {
   return true;
 }
 
+export async function assignmentAndNotSuspended(
+  organizationId,
+  user,
+  assignmentId,
+  assignment,
+  allowSupervolunteer = true
+) {
+  try {
+    await accessRequired(user, organizationId, "TEXTER");
+    await assignmentRequired(user, assignmentId, assignment);
+  } catch (e) {
+    console.log(typeof e);
+    if (e instanceof GraphQLError && allowSupervolunteer) {
+      await accessRequired(user, organizationId, "SUPERVOLUNTEER", true);
+    } else {
+      throw e;
+    }
+  }
+
+  return true;
+}
+
 export function superAdminRequired(user) {
   authRequired(user);
 
--- a/src/server/api/schema.js
+++ b/src/server/api/schema.js
@@ -40,7 +40,12 @@ import {
   reassignConversations,
   resolvers as conversationsResolver
 } from "./conversations";
-import { accessRequired, assignmentRequired, authRequired } from "./errors";
+import {
+  accessRequired,
+  assignmentRequired,
+  assignmentAndNotSuspended,
+  authRequired
+} from "./errors";
 import { resolvers as interactionStepResolvers } from "./interaction-step";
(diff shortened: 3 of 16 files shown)


src/server/api/errors.js
```diff
@@ -56,6 +56,28 @@ export async function assignmentRequired(user, assignmentId, assignment) {
   return true;
 }
 
+export async function assignmentAndNotSuspended(
```
schuyler1d, line 59: I think this should be built into the `assignmentRequired` function rather than have a separate function call.

just add the `accessRequired(user, organizationId, "TEXTER")` into there. The question is then where to get organizationId

 -- I guess in the past we've added it to the mutation as an argument (problematic, because then the texter could technically 'shop' sending a diff orgId to the server)
-- we could make sure looking up the org of the assignmentId is a fast cache-request, and just look it up (when unavailable in the higher context anyway) from the assignmentId
-- This would be worse for multi-tenant instances, but we could do something where suspension in one org basically suspends you from all orgs (in that case, the user-org affils are in-cache already, and we just have a isSuspended(user) function

src/server/api/schema.js
```diff
@@ -1155,11 +1165,15 @@ const rootResolvers = {
         return user;
       }
     },
+    currentUserWithAccess: async (_, { organizationId, role }, { user }) => {
```
schuyler1d, line 1168: I don't thiink we need this -- we can just query:
`currentUser { roles(organizationId:$orgId) }`

src/server/api/schema.js
```diff
@@ -1155,11 +1165,15 @@ const rootResolvers = {
...
+      await accessRequired(user, organizationId, role, false);
+      return user;
+    },
     organizations: async (_, { id }, { user }) => {
       if (user.is_superadmin) {
         return r.table("organization");
       } else {
-        return await cacheableData.user.userOrgs(user.id, "TEXTER");
+        return await cacheableData.user.userOrgs(user.id, "SUSPENDED");
```
schuyler1d, line 1176: INFO (no critique): I think this will (or could help) fix a long-standing bug where sometimes people get a "you are not affiliated with any organizations"

src/server/models/cacheable_queries/user.js
```diff
@@ -31,8 +31,6 @@ const userRoleKey = userId =>
 const userAuthKey = authId =>
   `${process.env.CACHE_PREFIX || ""}texterauth-${authId}`;
 
-export const accessHierarchy = ["TEXTER", "SUPERVOLUNTEER", "ADMIN", "OWNER"];
```
schuyler1d, line 34: INFO (positive critique): I love this -- one constant to rule them all and in the darkness bind them!

src/containers/AssignmentTexterContact.jsx
```diff
@@ -478,7 +486,15 @@ export class AssignmentTexterContact extends React.Component {
     this.setState({ disabled: true });
     try {
       if (optOutMessageText.length) {
-        await this.props.mutations.sendMessage(message, contact.id);
+        const sendMessageResult = await this.props.mutations.sendMessage(
+          message,
+          contact.id
+        );
+        if (sendMessageResult.errors && this.props.campaign.organization.id) {
```
schuyler1d, line 493: I'm a little remiss to put this condition here -- especially without a specific error tested for.

Can we either test for a specific error OR can an error trigger a refresh on UserMenu or some higher component of currentUser -- where an update on *that* will test for SUSPENDED role and redirect?
We could even put it in the `hoc` super component and test for any response having a particular suspend error and then redirect there.

Basically, I don't want to need this logic in every component that tests for suspension.

schuyler1d (changes requested):
This will be very useful and looks good.  I have some comments on changes, and then would like at least one test for a request on a suspended texter and confirmation that the right error/behavior comes back.

ibrand:
Note: this fixes #1005

ibrand:
Closing this to clear up confusion. Work being done for #1487 will cover this PR. This PR can be used as a reference if that's helpful!

Issue #1358, opened 2020-01-09
Feature Request: Display the Account view with the user's current profile info in the appropriate form fields (instead of showing a blank form)


arena:
**Problem**
The account information form gets cleared upon viewing (not sure why) but it's confusing to try to remember which email and phone you used. 

**Solution**
When the Account view is displayed, populate the form with the user's current profile info.

On the front end the fields are labeled First Name, Texting Alias (if it exists), Last Name, Cell Number, Email, and Password. The password is not displayed, of course. 

On the backend the fields are *placeholder for the names of these fields on the backend*

**Related Features**
Layout of the form: #1357
Texting Alias feature: #1109

arena:
Anybody know what these variables are named? Please suggest an edit to the description above.

arena:
@lperson is this a duplicate of an issue you created?

arena:
This is fixed now.
I don't know who fixed it, but THANK YOU!!
Was it you, @filafb or @lperson ?
@ibrand , ok if I close it?

arena:
I think @lperson had a similar issue that was redundant to this one. I'll try to find it later... but I think you can verify that the form gets populated and close this. Or take my word for it, and close this issue.

ibrand:
Yay thanks for letting me know it's fixed @arena ! Closing this

Issue #1357, opened 2020-01-09
Feature Request: Streamline, add field labels and improve usability of the Account view


arena:
**Problems**
On mobile the account form doesn't always display the hint text so the fields are hard to decipher. Horizontal forms (what you currently see on desktop) are inefficient and error prone.

**Solution**
- Adjust the layout to match this mockup (attached)
- add the Alias field between first and last name (related to issue #1109 --- if the feature is not complete, ok to leave this field out until it is complete)
- add labels above the fields so that they can be deciphered without the hint text

**Mockup**
![Account-View](https://user-images.githubusercontent.com/411070/72046375-29109c80-326d-11ea-8bdf-7648f1d2971b.png)

**Related Features**
Populate and display form fields: #1358
Texting Alias field: #1109

arena:
for reference this is how the account page currently displays for an existing user (Jan 2020)
(Note how the form is not populated)
![Group 11](https://user-images.githubusercontent.com/411070/72046746-e3080880-326d-11ea-9795-b8cc80e49293.png)

ibrand:
@arena How do you want to handle "add the Alias field between first and last name (related to issue #1109)" ? Should we call this blocked until #1109 is done? Or should we encourage submissions without that field for now?

arena:
@ibrand good point. I edited the solution above. I say, leave it out if the feature is not complete, then when the alias feature exists this mockup shows where the field should be placed.

arena:
Whoever picks this up is most welcome to break it into smaller pieces. I can definitely see the keeping the form populated being its own little puzzle that is separate from the layout change. @ibrand if you agree, maybe we should break it down now?

ibrand:
If we break it out into 2 issues, then I think each of those mini issues will be great `good first issues` so let's do it! Thanks!

arena:
Cool, I'll extract the populate one out to a new issue and edit this one. 🙌

arena:
@ibrand please tag this one a good first issue as well (if you agree) 😄

aschneit:
Working on this

aschneit:
[...]

PR #1365, opened 2020-01-13
Add alias name

### LINKED ISSUE #1109: add texterAliasfirstName as an option for sending
Add an option for folks, especially women, to add a 'texterAliasfirstName' to their user record to optionally send as - like 'Frank' (alias) rather than 'Lydia' (actual First Name). Often, texters IDing as women receive harassing replies back, perhaps as a reaction to common 'gender association' of their first name.  
Would also necessitate a preferredTexterFirstName field to use for actual  message creation (like {preferredTexterfirstName} token), with user firstName as default, but alias as an option.

filafb:
# Fixes #1109 

## Description

This feature allows user to send messages using an alias. If there is no alias for that texter, the message will include texter's first name when the interactions is _{ texterAliasOrFirstName }_

I had to change the resolver implementation for texter [here](https://github.com/filafb/Spoke/blob/63cfbac5bc423bfe09fb354d17df26f291069f10/src/server/api/assignment.js#L116). The previous implementation wasn't bringing me a new instance for texter after adding/changing an user alias. Not sure if it was a caching strategy, but now I'm finding the updated user instance direct from the db. 

I also fixed an issue (might be the issue #1358) that was avoiding the Dialog Box Edit User to be populated with user data. Not when user is editing a person data, this dialog box brings their current information.

# Checklist:

_* Added test for the new DB schema. I don't have much experience testing GraphQL. Happy to follow directions if anybody wants to point it to me._ 

interactions
list of people
edit user


arena:
🎉🎉🎉

joemcl:
thanks @filafb and @ibrand , and I hope I can make time to review this tonight. cc: @deasterdaywfp @lperson @bdatkins @suegibson .

__test__/backend.test.js
```diff
@@ -1,7 +1,7 @@
-import { resolvers } from "../src/server/api/schema";
```
bdatkins: General comment: Can you please use the prettier defaults (such as double quotes) so it's easier to see what has changed?

src/server/api/user.js
```diff
@@ -174,10 +174,10 @@ export const resolvers = {
   },
   User: {
     ...mapFieldsToModel(
-      ["id", "firstName", "lastName", "email", "cell", "assignedCell", "terms"],
+      ["id", "firstName", "lastName", "alias", "email", "cell", "assignedCell", "terms"],
       User
     ),
-    displayName: user => `${user.first_name} ${user.last_name}`,
+    displayName: user => `${user.first_name}${user.alias ? ` (${user.alias}) ` : " "}${user.last_name}`,
```
bdatkins, line 180: Should this be first name or alias in this case?

bdatkins (changes requested):
Looks good!  A really great addition to Spoke. Just a couple minor points from me.

src/containers/UserEdit.jsx
```diff
@@ -168,6 +169,7 @@ class UserEdit extends React.Component {
                 name="lastName"
                 {...dataTest("lastName")}
               />
+              <Form.Field label="Alias" name="alias" {...dataTest("alias")} />
```
schuyler1d: Interested in thoughts from others -- I feel like "Alias" doesn't have much context especially when the user is just creating their account.  Maybe label (or if there's a hint/description property) could be "Alias for texting" -- and make it clear someone can leave it blank?

src/containers/UserEdit.jsx
```diff
@@ -252,7 +257,8 @@ UserEdit.propTypes = {
   nextUrl: PropTypes.string,
   style: PropTypes.string,
   handleClose: PropTypes.func,
-  openSuccessDialog: PropTypes.func
+  openSuccessDialog: PropTypes.func,
+  query: PropTypes.function
```
schuyler1d: what/where is this query property?  the query result will end up in the `data` property above (I believe).  I think this is unnecessary.

src/server/api/assignment.js
```diff
@@ -113,10 +113,7 @@ export function getContacts(
 export const resolvers = {
   Assignment: {
     ...mapFieldsToModel(["id", "maxContacts"], Assignment),
-    texter: async (assignment, _, { loaders }) =>
-      assignment.texter
-        ? assignment.texter
-        : loaders.user.load(assignment.user_id),
+    texter: async (assignment) => await r.table("user").get(assignment.user_id),
```
schuyler1d: Yes, please revert this -- we definitely want this data to come from the cache.
The issue you observed IS something we should fix.  If you want to do it in this PR, then it looks like `editUser()` correctly calls ` await cacheableData.user.clearUser(member.id, member.auth0_id);`
The problem is that clearUser doesn't currently clear the loaders (local memory) cache.  What you want to do is copy the pattern found in:
src/server/models/cacheable_queries/organization.js (or campaign.js)
where you import `loaders` from "../../models" and then at the end of clearUser (in user.js in the same dir) you run `loaders.user.clear(Number(id))` and do the same thing with type string (to be extra sure).

Caching is a super-crucial part of scaling Spoke as it's a very 'chatty' client which makes the database the bottleneck, so it's important to retain (and in the future extend) the data that's cached -- especially for anything texter-related.

src/server/api/schema.js
```diff
@@ -424,6 +424,7 @@ const rootMutations = {
         const newUserData = {
           first_name: capitalizeWord(userData.firstName),
           last_name: capitalizeWord(userData.lastName),
+          alias: capitalizeWord(userData.alias),
```
schuyler1d: Since alias can easily override firstName -- should we chomp off whitespace before and after (probably good to do for first/last name, if you want, too)

src/server/models/user.js
```diff
@@ -1,9 +1,9 @@
-import thinky from "./thinky";
+import thinky from './thinky';
```
schuyler1d: I think the changes in this file are lint-churn.  `prettier` really prefers double-quotes here, so at least please make sure you're using the githook in package.json or running `prettier --write` on changed files on your own.

schuyler1d (changes requested):
This is a great complete implementation of the alias feature!
I've requested some minor changes here or there, but I think this is very close.
Thanks!

src/containers/UserEdit.jsx
arena: In my mockup I believe I used "Texting Alias"

arena:
Not sure if this is relevant, but this shows the form field with a label of 
"Texing Alias (optional)"
https://github.com/MoveOnOrg/Spoke/issues/1357

__test__/backend.test.js
filafb: Sure! Will do that. For some reason, husky & lint-staged are not running for me when I commit my stuff.

src/server/api/user.js
filafb: The code is doing something like `FirstName (alias) lasName` when alias is provided. If not, it will display `firstName lasName`. You can see that on this [print screen](https://user-images.githubusercontent.com/38737958/72227990-9e7fb400-3570-11ea-8617-049efca1f045.png)

src/containers/UserEdit.jsx
filafb: Is everybody fine if I change it to `Texting Alias`?

src/containers/UserEdit.jsx
filafb: I think you are right. Don't think we are using it inside the component. Will review it.

src/server/api/assignment.js
filafb: Alright. Will take a look at it.

src/server/models/user.js
filafb: Will do it!

src/server/api/schema.js
filafb: Sorry, I didn't fully understand the suggestion here.

filafb:
Thanks for all comments people! I will probably work on them this weekend! @schuyler1d @bdatkins

src/containers/UserEdit.jsx
arena: That is most intuitive
"Texting Alias (optional)"
Nobody has voiced any opposition and it's nice and descriptive, especially with the "(optional)" added to it. 

Yay!!!

bdatkins (dismissed):
Nice work!

filafb:
Hey @schuyler1d and @bdatkins! Just pushed new code that fixes the stuff you have commented on.

arena:
🙌🎉✨🎉🎉❤️🎉🎉✨🎉✨

ibrand:
@filafb some things I noticed when testing this out on the [heroku deployment of this branch](https://spoke-review-add-aliasn-h6crwu.herokuapp.com/):

1. When I sign up with an alias, it doesn't seem to save that alias vs when I go to my account details and edit it there, it works.

2. For a live campaign, if I change my name in the middle, I'm still texting with my old name

3. If the `texterName` variable was used, it doesn't use my alias, so I have to make sure as an admin to ignore that one and use `texterNameOrAlias`

4. Strangely when i used the `texterName` var it used the first name from my contact list as my name instead. Not sure what to make of that

Thank you so much for working on this! It's going to mean a lot for texters

ibrand (changes requested):
I want to make sure to get answers to the QA questions I pose in my comment before merging. thank you for your work so far!

arena:
not to answer for @filafb ... but I see your questions @ibrand and I have some thoughts
1. I don't think I should be prompted for an alias on sign up... just a way to add one after sign up
2. if you change any part of your name it won't have an effect on an existing assignment... so that seems expected (even if it's surprising to a user)
3. & 4. (over my head, no opinion)

src/server/api/schema.js
schuyler1d: I'm worried about the case where someone accidentally adds `__` (spaces) into the optional Alias field.  Since it overrides first_name in places, I think it's important to either remove front and leading spaces (so that 'only spaces' alias turns into an empty string and will then evaluate to false), and/or test that in this place.

Since we'd be stripping whitespace, I was thinking that might be good to do, even for the fields first/last name, as well, but that has less of a fatal result.

filafb:
Hi @ibrand. Bellow my answers:
1. I will check it. I haven't worked in anything related to the sign up page, but since the option is available, I think it should save the alias.
2. You mean the name or you also tried to change the alias?
3. That is correct. It was part of my question [here](https://github.com/MoveOnOrg/Spoke/issues/1109#issuecomment-533139062). It also gives the admin the power to use texter name when required.
4. I didn't get it. What is the difference between your name and your first name from my contact list?

src/server/api/schema.js
filafb: Got it. Will update the PR, trimming all three fields here.

filafb: @schuyler1d  Just pushed a new version. Check it out and let me know!

filafb:
@ibrand Just pushed a new version that should fix item number 1 in your list. Check it out.

ibrand:
~@filafb  I'm now seemingly unable to save my alias if I try and change it from the profile screen [in the test deploy](https://spoke-review-add-aliasn-h6crwu.herokuapp.com/)?~ This looks good on my end! Thanks for answering all the questions and for the revision @filafb ! ⭐️

ibrand (dismissed):

src/containers/UserEdit.jsx
```diff
@@ -40,7 +40,7 @@ class UserEdit extends React.Component {
 
   async componentWillMount() {
     if (!this.props.authType) {
-      await this.props.mutations.editUser(null);
+      await this.props.mutations.editUser(this.props.data.currentUser);
```
schuyler1d: do we need to test if `this.props.data` is available?  It's possible the component mounts without data available until the network connection is returned.

Can we add a comment here (once we figure out the need).
This needs testing/QA for:
* signup screen
* texter editing themselves screen
* admin editing a texter screen.

I worry that putting this.props.data.currentUser will only work for the second case (and possibly break the others)
Will defer to @ibrand on what we can QA vs. ensure we have a test for, for these use cases.

filafb: We can have here a function that returns either `this.props.data.currentUser` or null, if it catches an error reading `this.props.data`. What do you think? @schuyler1d

filafb: Ok, I will wait the QA to change it if needed. I was thinking about a `try/catch` block to define what `editUser` will be called with, but that should work too. And if that might be the case (this.props.data being undefined for some cases) I think it's a safe measure. I honestly don't know a good strategy to test it, specially if it is a race condition. Maybe changing the lifecycle method to `componentDidMount` could help?

schuyler1d: I think before QA, the logic I mentioned above needs to be in -- or are you arguing that we can leave it just as `props.data.currentUser`? even in the `!props.userId` case?

filafb: Oh, my bad! Sorry. I will update it later and do some tests. Thanks!!!

filafb: Hey @schuyler1d! Just tested the scenario you listed above and you were right. Editing a user in the list would bring the current logged in user info to Edit Page, not the user I clicked in. I reverted my changes and seems it is working now. Before, when I click to edit a user, the Edit Page was empty, but now it is bringing all the correct info. I think there was another change in the code that fixed that behavior I had seen.

schuyler1d (dismissed):
Ok, I think this is good to go -- this will be so helpful for texters!!!! :tada:

arena:
🎉 🙌

migrations/20190916145342_add_user_alias.js
```diff
@@ -0,0 +1,10 @@
+
+exports.up = function (knex) {
+  return knex.schema.table("user", function (table) {
+    table.string("alias").defaultTo("");
```
schuyler1d: I just realized we should default this to be null so the schema change is non-locking

schuyler1d (approved):
:tada:

PR #1351, opened 2020-01-07
Google Import script on AWS Lambda

schuyler1d:
# Fixes deployment of Google Script Import on AWS Lambda

## Description

AWS Lambda has a max environment variable size of 4K and GOOGLE_SECRET is typically large and most environments with other variables will go over that limit.  This PR adds functionality to load config for GOOGLE_SECRET through a deployable CONFIG_FILE.json -- there is a little bit of a dance to accomodate what claudia.js will upload, while making sure CONFIG_FILE.json is in `.gitignore`

It also stubs out a library that we may eventually start nudging other environment variables to, to avoid problems between global.* (useful for tests) and process.env.* (common for dev and production envs) -- adding `getConfig` and `hasConfig`

Finally, with this infrastructure, it hides the Script Import section if the feature is unavailable on the platform.

# Checklist:

- [x] I have manually tested my changes on desktop and mobile
- [x] The test suite passes locally with my changes
- [na] If my change is a UI change, I have attached a screenshot to the description section of this pull request
- [x] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#your-first-code-contribution), or has a documented reason in the description why it’s longer
- [x] I have made any necessary changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [na] My PR is labeled [WIP] if it is in progress

PR #1344, opened 2019-12-29
Fix: user edit dialog fields not populated

lperson:
# Fixes #1343 

## Description

The `editUser` mutation is used to retrieve user information, and in that use case it was referencing a null variable, causing it to fail.  This PR adds a guard against referencing that variable when it's null, and it adds a test.

# Checklist:

- [x] I have manually tested my changes on desktop and mobile
- [x] The test suite passes locally with my changes
- [N/A] ~If my change is a UI change, I have attached a screenshot to the description section of this pull request~
- [x] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#your-first-code-contribution), or has a documented reason in the description why it’s longer
- [N/A] ~I have made any necessary changes to the documentation~
- [x] I have added tests that prove my fix is effective or that my feature works
- [N/A] ~My PR is labeled [WIP] if it is in progress~

PR #1340, opened 2019-12-27
Stage main

ibrand:
Would love a 👍 from both @lperson and @schuyler1d because I had a merge conflict specifically on the function params in [await sendMessage](https://github.com/MoveOnOrg/Spoke/compare/stage-main?diff=unified&expand=1#diff-8691de01bf75ec536b4d3ec22ffb7a64R69) and on a refactor of the var [serviceName](https://github.com/MoveOnOrg/Spoke/compare/stage-main?diff=unified&expand=1#diff-5d005dc78135c88ea39522684234badcR115-R120) that I think I resolved correctly, but would love you both to inspect.

@lperson 
#1178 
#1293 
#1179 
#1344 

@SAnschutz 
#1277 

@jeffm2001 
#1306 

@schuyler1d 
#1311
#1309 
#1346 
#1351

PR #1306, opened 2019-12-16
Allow alternate schema for postgres db

jeffm2001:
## Description

This is a very simple change that allow you to specify a `searchPath` in the knex config. See: http://knexjs.org/#Installation-client

This allows you to share a single postgres database across multiple Spoke instances, but keep data separate, e.g. using `heroku addons:attach`. It's probably mostly useful on Heroku, because Heroku allows you to [attach a database addon to multiple apps](https://devcenter.heroku.com/articles/heroku-postgresql#sharing-heroku-postgres-between-applications), but doesn't allow you to create multiple databases.

If the searchPath is left blank, than knex will just ignore it, see: https://github.com/knex/knex/blob/master/lib/dialects/postgres/index.js#L21

# Checklist:

- [x] I have manually tested my changes on desktop and mobile
- [x] The test suite passes locally with my changes
- [ ] If my change is a UI change, I have attached a screenshot to the description section of this pull request
- [x] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#your-first-code-contribution), or has a documented reason in the description why it’s longer
- [x] I have made any necessary changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] My PR is labeled [WIP] if it is in progress

PR #1367, opened 2020-01-13
Stage main2

schuyler1d:
Includes everything in #1340 except https://github.com/MoveOnOrg/Spoke/pull/1293/ which had a bug that requires further investigation

This should be the PR for Spoke 5.0.  It still needs release notes which should include migration instructions here (for larger instances): https://github.com/MoveOnOrg/Spoke/pull/1311#issuecomment-568786302 (or at least a link to that comment).

@lperson
#1178
#1179
#1344

@SAnschutz
#1277

@jeffm2001
#1306

@schuyler1d
#1311
#1309
#1346
#1351


ibrand (approved):
🎉

PR #1371, opened 2020-01-14
Ingest framework PR 2: csv-upload

### LINKED ISSUE #1370: RFC: Abstracted Contact Ingest methods
## Proposal:

Move the logic for CSV contact uploading (and data warehouse contact ingest) into an abstracted framework that can include other ways and implementations to load contacts for a particular campaign.

## Motivation:
One of the challenges of other groups using spoke, especially, is its lack of support (in our branch) for things like the VAN, ActionNetwork, NationBuilder, and other CRMs and systems.

Even we have had this challenge with importing from our Datawarehouse -- a somewhat stunted project in 2017, hacking the frontend interface to allow a separate way to load contacts felt clunky.

For those orgs/projects that have forked spoke, they’ve often created a second issue -- that most integrations have required deep changes across the entire codebase: frontend, backend, server loading, job queue choices, …. And so PRs that implement support are daunting to review, and force us to weigh the complexity added with the value to us and/or other orgs.  We have NOT, so far, added any of these PRs but that has caused orgs to fork and the changes throughout the codebase make merges also challenging.

## Detailed Design:
Spoke already has a framework that can load in optional code called “action handlers” which is used to process responses like ActionKit RSVPs or SMS subscription ‘yes’s.  With these there is a directory in the codebase and then each file in the directory has a structured set of functions which it must implement.   Then the application calls them if it’s configured to be enabled.

In code you can see the API methods to implement here:
https://github.com/MoveOnOrg/Spoke/pull/1369/files#diff-97aee2b4cd87698738080584a707bef3
Also feel free to comment on this Google Doc:
https://docs.google.com/document/d/1KApTC06XMJtM-tS75GVSJpXYlGUok3KDqa6yKpilpNo/edit?folder=0AGqdtHsEddKSUk9PVA#bookmark=id.wdou0kzdvpvb

These mirror not much more than what is necessary to implement CSV uploading which is the baseline template.

Added infrastructure: similar to action_handlers, in the appropriate places in the code, it would decide which ingest-contact-loaders would be

schuyler1d:
# Fixes #1370 

## Description

This PR uses the framework built in #1369 and moves the csv-upload functions into that framework. The directory is now src/ingest-contact-loaders/csv-upload.

# Checklist:


src/ingest-contact-loaders/csv-upload/index.js
```diff
@@ -0,0 +1,136 @@
...
+  return {
+    "environmentVariables": [],
+    "description": "CSV Upload a list of contacts",
+    "setupInstructions": "Nothing is necessary to setup since this is default functionality"
+  }
+}
+
+export async function available(organization, user) {
+  /// return an object with two keys: result: true/false
+  /// if the ingest-contact-loader is usable and has
```
ibrand:
```suggestion
  /// these keys indicate if the ingest-contact-loader is usable
```

I think this is the sentence you meant? This is a nit, it's just not a fully formed comment

src/ingest-contact-loaders/csv-upload/index.js
```diff
@@ -0,0 +1,136 @@
...
+  /// * "Request of Doom" scenarios -- queries or jobs too big to complete
+  const campaignId = job.campaign_id;
+  const jobMessages = [];
+
+  await r
+    .knex("campaign_contact")
+    .where("campaign_id", campaignId)
+    .delete();
+
+  // TODO
```
ibrand: What's the `TODO` here? Is it just residual from the comment-based instructions? or is there something left in here to do?

src/ingest-contact-loaders/csv-upload/index.js
```diff
@@ -0,0 +1,136 @@
...
+  let contacts;
+  if (job.payload[0] === '{') {
+    contacts = JSON.parse(job.payload).contacts;
```
ibrand, line 95: if `JSON.parse` fails, it will return an error. Should this be a `try catch` block? Could that `try catch` even replace the kind of hacky if statement checking for `{`? Not sure how often we've seen problems with `JSON.parse` failing, but it'd be nice to use the `catch` to surface a user facing message

src/ingest-contact-loaders/csv-upload/index.js
```diff
@@ -0,0 +1,136 @@
...
+  } else {
+    contacts = (await unzipPayload(job)).contacts;
+  }
+  if (maxContacts) {
+    // note: maxContacts == 0 means no maximum
+    contacts = contacts.slice(0, maxContacts);
+  }
+  const chunkSize = 1000;
```
ibrand, line 103: any reason for this to be configurable based on infrastructure? I'm guessing no, but wanted to pose the question

src/ingest-contact-loaders/csv-upload/index.js
```diff
@@ -0,0 +1,136 @@
...
+    const datum = contacts[index];
+    if (datum.zip) {
+      // using memoization and large ranges of homogenous zips
+      datum.timezone_offset = await getTimezoneByZip(datum.zip);
+    }
+    datum.campaign_id = campaignId;
+  }
+  for (let index = 0; index < numChunks; index++) {
+    /*
+    await updateJob(job, Math.round((maxPercentage / numChunks) * index))
```
ibrand: why is this commented out? I know I always say to cut commented code, but I really do think that it's a lot easier to read if there isn't commented code left in. If we really need it, can we explain why it's there with another comment?

src/ingest-contact-loaders/csv-upload/react-component.js
```diff
@@ -119,38 +77,62 @@ export default class CampaignContactsForm extends React.Component {
...
     });
     const contactCollection = {
       contactsCount: contacts.length,
-      contactSql: null,
       customFields,
       contacts
     };
-    this.props.onChange(contactCollection);
+    const self = this;
+    // this.props.onChange(JSON.stringify(contactCollection));
```
ibrand: ✂️

ibrand (commented):
A few comments but nothing too major. My only real thing is around that `JSON.parse` function. If it will fail silently, then I think we should fix that before shipping this. Thank you for writing such thorough tests!

src/ingest-contact-loaders/csv-upload/index.js
schuyler1d: If JSON.parse can fail, then the tests will fail.  It should be impossible.  Also note that this is partially guarded already because we confirm tahat the first char is a `{` (a good sign of JSON, anyway)-- I think it's ok to make this an assumption, and if it throws/errors, then the application error log will be visible.

src/ingest-contact-loaders/csv-upload/index.js
schuyler1d: dur, will remove

src/ingest-contact-loaders/csv-upload/index.js
schuyler1d: I don't think so -- testing across both local and small/big DBs suggest this is a good constant -- it's basically related to how much work a single transaction should do -- amount of data processed between `await` context switches.

src/ingest-contact-loaders/csv-upload/index.js
schuyler1d: updateJob was necessary during a slow, but large contact upload.  When we deploy on AWS Lambda, there's more a hard-maximum related to data-size and then the max-size is processable pretty quickly.

I have an idea for better statuses in ingest-contact-loaders, if we come back to this work, but even then, it would not rely on the same structure -- statuses reported through JobRequests is inherently flawed, because it's all deleted as soon as the job is complete.

ibrand (approved):

ibrand:
Closing this since it was merged in the last stage-main as part of a bigger PR

PR #1379, opened 2020-01-23
enable tests for editUser in backend.test.js

schuyler1d:
# Fixes skipped tests
For some reason `describe.only` at the bottom was making jest skip a bunch more tests (18 in total) in the same file

# Checklist:

### DIFF
--- a/__test__/backend.test.js
+++ b/__test__/backend.test.js
@@ -709,7 +709,7 @@ describe("Campaign", () => {
   });
 });
 
-describe.only("editUser mutation", () => {
+describe("editUser mutation", () => {
   let testAdminUser;
   let testTexter;
   let testOrganization;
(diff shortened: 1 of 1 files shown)


__test__/backend.test.js
```diff
@@ -740,10 +740,10 @@ describe.only("editUser mutation", () => {
 
   it("returns the user if it is called with a userId by no userData", async () => {
     const result = await runGql(editUserMutation, variables, testAdminUser);
-    expect(result).toEqual({
+    expect(result).toMatchObject({
       data: {
         editUser: {
-          id: "2",
+          // id: "2", // id might be diff
```
ibrand, line 746: Shouldn't we just delete this line then?

schuyler1d: I thought it would be good to document why we are not comparing by equality (and in case there are other tests, a place where it's shown how to fix)

ibrand (approved):
Yay!

Issue #410, opened 2017-11-28
Test and document worker processes


jmcarp:
The [workers](https://github.com/MoveOnOrg/Spoke/tree/master/src/workers) module doesn't have any unit or integration tests yet. We should write some, and continue to update tests when worker behavior changes. This might be a good task for the hackathon!

I'm also a little uncertain about exactly what the worker code does, and how it's organized. Better documentation might be helpful here, and might make it easier to write tests. Sequence diagrams could be especially useful--I like [mermaid](https://mermaidjs.github.io/sequenceDiagram.html) for this, but there are lots of great tools we can use to throw together a diagram in a few minutes.

Frydafly:
Following up here since it's been some time! There's some description of the function of the workers module on [here](https://moveonorg.github.io/Spoke/#/EXPLANATION-development-guidelines?id=asynchronous-tasks-workers)

@ibrand @jmcarp any ideas on how the existing documentation could be improved?

PR #1380, opened 2020-01-23
wip: test writing documentation

schuyler1d:
# Fixes no full documentation on how to write tests

## Description

Let's document testing patterns, gotchas and best practices.

- [ ] Figure out when we really need mount() vs. shallow() in React components
- [ ] put a `sleep` in ./lib or somewhere else appropriate (and document)

and generally need more details.  This outline is meant to be fleshed out -- please help expand!

### DIFF
--- a/docs/HOWTO-write_tests.md
+++ b/docs/HOWTO-write_tests.md
@@ -1,0 +1,99 @@
+# What we want to test
+
+- Component behavior
+- Proper Component loading
+-
+
+# React Tests
+
+We want to test component behavior, make sure components load based on the full range of
+properties they can be passed, make sure the data from properties and state successfully
+renders.
+
+For networked components (i.e. containers) that get properties based on GraphQL calls,
+we want the queries included in components to be tested, but those tests should be in
+**test**/server/api/ directory (rather than in the components/containers test dirs)
+
+## Setup
+
+## Standard changes to a container for testability
+
+- export the underlying class and then
+  `import { Foo } from "../../src/containers/foo"` (underlying component) vs
+  `import Foo from "../../src/containers/foo"` loaded component
+- export the `` gql`...` `` with `export const dataQuery = gql...`
+  Name it explicitly `dataQuery` unless there are multiple queries that need to be exported
+
+## Shallow copy testing
+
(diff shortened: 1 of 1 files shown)


ibrand:
This is so helpful!! I can try and help add stuff on too. Do you think we could merge it as is but included that we need more contributions to it? Just to make folks are actually able to find what's already here?

Frydafly:
Does this cover this issue? https://github.com/MoveOnOrg/Spoke/issues/410

schuyler1d:
> Does this cover this issue? #410

No -- worker processes are background 'cron' jobs -- this is for how to run tests, which we would need to update, if we merged this -- e.g. we should at least link to Cypress test documentation I believe we have.

Frydafly:
@schuyler1d is this the documentation you're referring to? https://github.com/MoveOnOrg/Spoke/blob/ddd32f474461da98d6377b59e8a2f76099370543/docs/HOWTO-run_tests.md#integration-testing

Frydafly:
hey @schuyler1d wondering if you might want to add this as a change to this existing document? https://github.com/MoveOnOrg/Spoke/blob/main/docs/HOWTO-run_tests.md

schuyler1d:
I think we were thinking of it as a separate document just to separate the 'I made a change and want to run the tests' vs. 'I am writing a test' which is more a subset of code guidelines.  But agree that either way it would work -- it's more the task of completing this document (and then putting it somewhere)

Frydafly:
hm! in that case @schuyler1d do you think it would be beneficial to merge as-is, perhaps with a caveat text at the bottom that this document could use more work, so that there is more visibility on this doc and folks might be inspired to add to it?

Frydafly:
@schuyler1d I think it would be great to merge as-is and get folks adding to this/contributing more! what do you think?

Frydafly (approved):
🥳 -- actually, @schuyler1d would be great to add this doc to the sidebar so that this shows up on the microsite, maybe even under the help-needed section!

Frydafly (changes requested):
Can you add this doc to the [sidebar.md file](_sidebar.md) and then we can merge?

PR #1405, opened 2020-02-05
Fix supervolunteer access to campaign data

schuyler1d:
This fixed an issue that came up in production when testing stage-main.
Supervolunteers don't have access to ingestMethods, but they should still be able to load campaign data.

### DIFF
--- a/__test__/server/api/campaign.test.js
+++ b/__test__/server/api/campaign.test.js
@@ -36,6 +36,7 @@ let testAdminUser;
 let testInvite;
 let testOrganization;
 let testCampaign;
+let testSuperVolunteerUser;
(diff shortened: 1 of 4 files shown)


__test__/server/api/campaign.test.js
```diff
@@ -67,6 +79,20 @@ afterEach(async () => {
   if (r.redis) r.redis.flushdb();
 }, global.DATABASE_SETUP_TEARDOWN_TIMEOUT);
 
+it("allow supervolunteer to retrieve campaign data", async () => {
```
ibrand, line 82: YAY TESTS

ibrand (approved):

PR #1414, opened 2020-02-09
Use GraphQLErrorWithStatus to avoid TypeErrors caused by throwing GraphQLError created with an object

### LINKED ISSUE #1413: Throwing GraphQLError results in TypeError when it its constructor is called with an object
This is easily reproduced in the node repl:

Calling the `GraphQLError` constructor with no parameters or a string is ok.

```
> const { GraphQLError } = require("graphql/error");

> throw new GraphQLError();
undefined
> throw new GraphQLError("foo");
foo
```

But calling it with an object is not.

```
> throw new GraphQLError({message: "oh crap", "status": 401});
TypeError: Cannot convert object to primitive value
    at Array.join (native)
```

The constructor is defined in ` node_modules/graphql/error/GraphQLError.js` at line 27.

We have 10 instances of this:
- src/server/api/mutations/sendMessage.js|25 col 11| throw new GraphQLError({
- src/server/api/mutations/sendMessage.js|48 col 11| throw new GraphQLError({
- src/server/api/mutations/sendMessage.js|65 col 14| throw new GraphQLError({
- src/server/api/mutations/sendMessage.js|74 col 11| throw new GraphQLError({
- src/server/api/mutations/sendMessage.js|110 col 11| throw new GraphQLError({
- src/server/api/mutations/bulkSendMessages.js|14 col 11| throw new GraphQLError({
- src/server/api/schema.js|298 col 18| throw new GraphQLError({
- src/server/api/schema.js|521 col 15| throw new GraphQLError({
- src/server/api/schema.js|759 col 15| throw new GraphQLError({
- src/server/api/schema.js|817 col 15| throw new GraphQLError({

lperson:
Allows us to create a GraphQL error with a status without passing an
object to GraphQLError, which causes other problems.

# Fixes #1413 

## Description

# Checklist:


schuyler1d (dismissed):
Needs a rebase, but approving the code.

matteosb:
I'd like to hold off on this and instead switch to using ApolloErrors with error codes after upgrading to apollo-server 2.x: https://www.apollographql.com/docs/apollo-server/data/errors/#codes

schuyler1d:
Going to close this in anticipation of Apollo server upgrade.

PR #1417, opened 2020-02-12
Prettierrc file

ibrand:
# Description
We've seen multiple PR's recently that have prettified our files based on the contributor's global system prettier. This will add a local `.prettierrc` to our repo so that when the linters run they always apply the prettier defaults. This should be the same as running prettier if you don't have any prettierrc on your system at all.

Huge shoutout to @filafb both for pointing out this problem and experimenting on the config to figure out how to fix it. This is his work, I'm just making the PR.

# Checklist:


.prettierrc
```diff
@@ -0,0 +1,18 @@
+{
+  “arrowParens”: “avoid”,
```
schuyler1d: this looks like it has a bunch of smart quotes.....  Can you convert it to JSON?

Also, in the PR, can you confirm that after you have this, running prettier on all files doesn't change anything?  If it does, then let's maybe re-PR updates on those (in a separate PR mid-stage)

schuyler1d (changes requested):
json

ibrand:
ahhh smart quotes. Fixed those. So I ran `prettier --write "**/*.js"` like what I ran to initially prettify things and there were a few things that got changed but they all appear to be ingest contact loader related and every change was a line width change.

For example:
```diff
-import {
-  sendMessage
-} from "./sendMessage";
-import {
-  bulkSendMessages
-} from "./bulkSendMessages";
-import {
-  findNewCampaignContact
-} from "./findNewCampaignContact";
+import { sendMessage } from "./sendMessage";
+import { bulkSendMessages } from "./bulkSendMessages";
+import { findNewCampaignContact } from "./findNewCampaignContact";
```

Do you have a local `.prettierrc` that has a smaller line width than standard?

Here are all the files changed:
```
	modified:   src/integrations/contact-loaders/index.js
	modified:   src/server/api/lib/import-script.js
	modified:   src/server/api/mutations/findNewCampaignContact.js
	modified:   src/server/api/mutations/index.js
	modified:   src/server/middleware/render-index.js
```

schuyler1d:
I don't have a prettierrc file, but I wouldn't be surprised if the version or something leads to different results -- in any case, let's merge this -- and then post stable stage-main, let's run it once more (or i can if you like) so that it's all locked in going forward

schuyler1d (dismissed):
:+1:

filafb (commented):
A few things I think we should add to make the eslint / prettier process more robust:

1. Remove the styles rules from the [eslint config file](https://github.com/MoveOnOrg/Spoke/blob/prettierrc-file/.eslintrc), since we are delegating those rules to prettier;

2.  Add the [eslint-config-prettier library](https://www.npmjs.com/package/eslint-config-prettier) and extend the eslint config file to use it - add it as the last element in an array, so this way any other style rule from the eslint plugin we are using will be override by prettier;

3. Add a command to run eslint on the linstaged config. The way it is configured now, eslint is only showing the error messages in the files. Adding this command, it will avoid any commit breaking eslint rules.

Doing so, we ensure every commit will style the code base in the right way and no one will commit anything breaking eslint rules.

bdatkins (dismissed):

filafb (dismissed):
Great!

schuyler1d (approved):
:+1:

PR #1430, opened 2020-02-21
revert eslint on commit-hook because we have not eslint'd everything,…

schuyler1d:
eslint is doing too much, and since we haven't pre-linted the files, this will cause too much churn in PRs. Let's eslint as separate PRs (doing no changes besides). Since eslints don't have a '--write' like prettier, this will be piece-meal work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -189,9 +189,6 @@
     }
   },
   "lint-staged": {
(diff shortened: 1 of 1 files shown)


ibrand (approved):
Thank you for catching this!

PR #1448, opened 2020-03-14
Refactor parse-csv

lperson:
# Description
* This PR refactors parseCSV and code related to it
    * It moves:
        * code that will be used by other contact-loader integrations out of csv-upload/react-component.js and into lib/parse_csv.js
        * code that is only used by csv-upload/react-component.js out of lib/parse_csv.js
    * It introduces a third parameter to parseCSV, an options object that currently supports 2 optional keys:
        * rowTransformer -- a function that gets called on each row in the file after it is parsed.  It takes 2 parameters, an array of fields and the object that results from parsing the row.  It returns an object after transformation.  The function can do lookups, field mappings, remove fields, add fields, etc.  If it adds fields, it should push them onto the fields array in the first parameter.
        * headerTransformer -- a function that gets called once after the header row is parsed.  It takes one parameter, the header name, and returns the header that should be used for the column.  An example would be to transform first_name to firstName, which is a required field in Spoke.
    * It adds tests.

It's longer than 300 lines because I added a lot of tests.

# Checklist:

- [N/A] ~If my change is a UI change, I have attached a screenshot to the description section of this pull request~
- [N/A] ~My PR is labeled [WIP] if it is in progress~

### DIFF
--- a/src/lib/parse_csv.js
+++ b/src/lib/parse_csv.js
@@ -1,9 +1,8 @@
 import Papa from "papaparse";
 import _ from "lodash";
 import { getFormattedPhoneNumber, getFormattedZip } from "../lib";
-import humps from "humps";
 
-const requiredUploadFields = ["firstName", "lastName", "cell"];
+export const requiredUploadFields = ["firstName", "lastName", "cell"];
 const topLevelUploadFields = [
   "firstName",
   "lastName",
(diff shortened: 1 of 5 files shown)


src/lib/parse_csv.js
```diff
@@ -54,37 +53,68 @@ const getValidatedData = data => {
...
+  //   field in Spoke.
+  const { rowTransformer, headerTransformer } = options || {};
   Papa.parse(file, {
     header: true,
-    transformHeader: ensureCamelCaseRequiredHeaders,
+    ...(headerTransformer && { transformHeader: headerTransformer }),
+    skipEmptyLines: true,
     // eslint-disable-next-line no-shadow, no-unused-vars
-    complete: ({ data, meta, errors }, file) => {
+    complete: ({ data: parserData, meta, errors }, file) => {
```
schuyler1d, line 95: nit, but why do we rename data => parserData just to set it to data, lower-down?

schuyler1d (approved):
:heart: love it.

src/lib/parse_csv.js
lperson: I did that to avoid modifying the function parameter.

PR #1450, opened 2020-03-16
completeContactLoad sets fields to 0 instead of null

lperson:
## Description

completeContactLoad was setting the columns deleted_optouts_count and duplicate_contacts_count in campaign_admin to null when they should have been set to 0.

Also added a happy-path test for completeContactLoad.

# Checklist:

- [N/A] ~I have manually tested my changes on desktop and mobile~
- [N/A] ~If my change is a UI change, I have attached a screenshot to the description section of this pull request~
- [N/A] ~My PR is labeled [WIP] if it is in progress~

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -247,8 +247,8 @@ export async function completeContactLoad(
   const campaign = await Campaign.get(campaignId);
   const organization = await Organization.get(campaign.organization_id);
 
-  let deleteOptOutCells;
-  let deleteDuplicateCells;
+  let deleteOptOutCells = null;
+  let deleteDuplicateCells = null;
   const knexOptOutDeleteResult = await r
     .knex("campaign_contact")
     .whereIn("cell", getOptOutSubQuery(campaign.organization_id))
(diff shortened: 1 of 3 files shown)


src/workers/jobs.js
```diff
@@ -295,8 +295,12 @@ export async function completeContactLoad(
     .knex("campaign_admin")
     .where("campaign_id", campaignId)
     .update({
-      deleted_optouts_count: deleteOptOutCells || null,
-      duplicate_contacts_count: deleteDuplicateCells || null,
+      deleted_optouts_count:
+        deleteOptOutCells || deleteOptOutCells === 0 ? deleteOptOutCells : null,
```
schuyler1d: maybe we should just set them to `null` higher up where they're declared in `let deleteOptOutCells`  and then they can be fully set here without `||`'s, etc?

schuyler1d (approved):
Thanks for fix.  I had a suggestion on how to make it even simpler, but this is good, too.

src/workers/jobs.js
lperson: Done!

PR #1452, opened 2020-03-18
Heroku button paid

ibrand:
The heroku button should deploy a Spoke that is usable in a prod scenario. Lots of orgs click it but then don't have great next steps once they very quickly max out the DB.

Talking to @jeffm2001 and @schuyler1d we determined that this is a fine starter place for a prod deploy

PR #1459, opened 2020-03-24
Create a quick link for deploying a production set up to Heroku

ibrand:
# Description
Taking notes from #1452 I've prepped a link for deploying Spoke straight to prod. I've also included a change to app.json that free and paid deploys will need that makes `DB_USE_SSL` true. I think that the new generation of Heroku DBs require SSL connections.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,10 +1,9 @@
 ---
 name: Feature request
 about: Suggest an idea for Spoke
-title: 'Feature Request: <your-title-here>'
+title: "Feature Request: <your-title-here>"
 labels: C-enhancement
-assignees: ''
-
+assignees: ""
 ---
 
 **Problem**
--- a/README.md
+++ b/README.md
@@ -11,9 +11,13 @@ The latest version is [5.2](https://github.com/MoveOnOrg/Spoke/tree/v5.2) (see [
 
 ## Deploy to Heroku
 
+Use the Heroku Button to deploy a version of Spoke suitable for testing. This won't cost any money and will not support production usages. It's a great way to practice deploying Spoke or see it in action.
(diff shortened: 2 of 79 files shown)


jeffm2001:
This looks great @ibrand.

One question, is it `DB_USE_SSL` that needs to be set, or `PGSSLMODE`? Just wondering because in HOWTO_HEROKU_DEPLOY.md it references PGSSLMODE. In my deploy I have both set.

My one other flag is in the README, where it says "ballpark for 100k texts" is somewhat vague. Is that intended to mean per month? total? I actually have no idea what the correct answer is, and in reality it's probably more determined by number of simultaneous users rather than number of messages. Perhaps it would be better to say something line "should be suitable for production-level usage for most organizations."

jeffm2001:
Oh, and @schuyler1d, should the `heroku-redis:hobby-dev` add-on be included for the standard (free) Deploy To Heroku button as well?

schuyler1d (dismissed):
:+1:

ibrand:
Gonna answer a few questions here:

> Oh, and @schuyler1d, should the `heroku-redis:hobby-dev` add-on be included for the standard (free) Deploy To Heroku button as well?

I suppose it doesn't matter too much with the free deploy since redis is used for scaling purposes and the free version never needs to be concerned with scale.

> Perhaps it would be better to say something line "should be suitable for production-level usage for most organizations."

I agree! I'll change that now.

> One question, is it DB_USE_SSL that needs to be set, or PGSSLMODE?

From my testing, it needs to be `DB_USE_SSL` and [this piece of code](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/knex-connect.js#L25-L26) confirms that that's the right variable. When searching the entire codebase I only see `PGSSLMODE` referenced in docs. I'm pretty sure that indicates that it's either since been deprecated or was never the right var and that I should edit it out of the docs in a PR too.

schuyler1d (approved):
:+1:

jeffm2001:
Looks good!

PR #1471, opened 2020-04-01
A GSForm wrapper for material-ui/AutoComplete

lperson:
## Description

This PR introduces a component that wraps material-ui/AutoComplete so it can be included in a GSForm.  For example:

```
          <GSForm
            schema={formSchema}
            value={{ colors: colors[1] }}
            onChange={() => {}}
          >
            <Form.Field
              name="colors"
              type="autocomplete"
              choices={colors}
              hintText="What's your favorite color?"
            />
```

I intend to use this in the PR for syncing Spoke answers to canvass results in VAN.

# Checklist:


src/components/App.jsx
```diff
@@ -22,7 +23,8 @@ Form.addInputTypes({
   email: GSTextField,
   script: GSScriptField,
   select: GSSelectField,
-  password: GSPasswordField
+  password: GSPasswordField,
+  autocomplete: GSAutoComplete
```
schuyler1d, line 27: Is it possible to do this without including it in input types here?  (instead, using it by directly importing `GSAutoComplete`)?

My understanding is that addInputTypes is deprecated in newer React-formal versions -- so the others here will have to be changed the same way, but it would be good not to add to the migration burden.

See https://github.com/jquense/react-formal/issues/167

package.json
```diff
@@ -172,7 +172,7 @@
     "json2csv": "^3.6.2",
     "lint-staged": "^9.2.0",
     "mockdate": "^2.0.2",
-    "nodemon": "^1.9.2",
+    "nodemon": "^2.0.2",
```
schuyler1d: is this related to this change?  If not, let's not mix things together.

src/components/App.jsx
schuyler1d: @lperson pointed out to me that until we upgrade react-formal, we can't do it any other way -- so it's chicken/egg, and we shouldn't block on this.

lperson: In light of this, would you be so kind as to approve 🙏

lperson:
@ibrand why did you close this without merging?

ibrand:
That's weird... certainly didn't mean to close it! Thanks for seeing that. Reopening! Got in a weird state while making stage-main just now

lperson:
@schuyler1d and @ibrand would it be possible to merge this into stage main.  I'm using it for the action handler work and it would be useful to have it not show up in the diffs for that.  Thank you!

schuyler1d:
I think this PR still has yarn.lock churn that's unrelated to this change.

If `main` is merged into this, I think there's already a twilio upgrade there, so hopefully it will merge ok, and then there won't be a diff.

That might be ok anyway, but please please please -- let's keep these PRs contained to the needed changes.  There's a lot of community activity and we need to stay in our own focused areas.

PR #1475, opened 2020-04-04
retryable timeoutable fetch

lperson:
## Description

Because VAN integration relies on API calls to VAN, it would be nice to have a way to retry and timeout those API calls.  `isomorphic-fetch` doesn't have retry behavior and timeout support is BYO.

This PR introduces 
* `nock` (which allows us to mock http requests in tests -- 170K in dev only, and I can't believe we lived without this!) 
* `abort-controller` (a small -- 76K -- implementation of `AbortController` so we can  bring our own timeout to `fetch` which is BYO retry)
* an updated version of `node-fetch` which is a dependency of `isomorphic-fetch`. This does not add a new package, but rather provides an updated version of an existing package. We need the update to get support for the `signal` property, which enables us to implement a timeout.

# Checklist:

- ~[ ] I have manually tested my changes on desktop and mobile~
- ~[ ] If my change is a UI change, I have attached a screenshot to the description section of this pull request~
- ~[ ] I have made any necessary changes to the documentation~
- ~[ ] My PR is labeled [WIP] if it is in progress~


src/server/lib/http-request.js
```diff
@@ -0,0 +1,123 @@
...
+
+  const requestId = uuid();
+  const propsRetries = originalProps.retries;
+  const retries =
+    propsRetries !== null && propsRetries !== undefined ? propsRetries : 2;
+  const timeout = originalProps.timeout || 2000;
+
+  const retryDelay = () => {
+    const baseDelay = 50;
+    const randomDelay = Math.floor(Math.random() * (baseDelay / 2));
```
ibrand, line 19: Why incorporate randomness here?

lperson: We found it to be a best practice at my day job.

src/server/lib/http-request.js
```diff
@@ -0,0 +1,123 @@
...
+import { v4 as uuid } from "uuid";
+
+const requestWithRetry = async (url, props) => {
+  const originalProps = props || {};
+  const remainingProps = {
+    ...(originalProps || {})
+  };
+  delete remainingProps.validateStatus;
+  delete remainingProps.retries;
+  delete remainingProps.timeout;
```
schuyler1d: Everything above this (and a little below) seems like it would be much easier to read with ES6 default deconstructed args, e.g. `const requestWithRetry = async (url, { timeout=2000, retries=2, validateStatus=false }) =>`...etc

src/server/lib/http-request.js
```diff
@@ -0,0 +1,123 @@
...
+
+    return RetryReturnError.ERROR;
+  };
+
+  for (let attempt = 0; attempt < retries + 1; attempt++) {
+    let error;
+    let response;
+
+    const controller = new AbortController();
+    const controllerTimeout = setInterval(() => {
```
schuyler1d: Shouldn't this be `setTimeout`?  if not, then you don't need to 'clear' it below, and this is inside the loop, so it's just once, right?  If not, then please document this extensively.

schuyler1d (changes requested):
Some tweaks.  Also, in the VAN PR that uses this, can we add some detailed inline notes on WHY we need retries?  Does the request frequently fail?  Are we essentially polling an async process?

In VAN, this will be blocking the load of the campaign editors often enough, so we need to balance that experience with having VAN if it wasn't.

src/server/lib/http-request.js
lperson: Yeah, good catch.

lperson: `clearTimeout` is necessary for when the request completes before the timer timeout fires. It's inside the loop because a request can fail, and be retried, and the retry could time out.

lperson: SHA 04fd3889d999dcaf618a21b749fa95b772df50f6

src/server/lib/http-request.js
lperson: SHA 3dffe9ec9ec37691d42b25610d4deca98ad15c94

schuyler1d (approved):
Approving, but both for simplicity and what we generally want, please set `retries=0` as the default, and any function that uses this has to be explicit about retries in the call.

lperson:
> Approving, but both for simplicity and what we generally want, please set `retries=0` as the default, and any function that uses this has to be explicit about retries in the call.

@schuyler1d I made this change.

PR #1478, opened 2020-04-07
Separate Twilio accounts per organization

### LINKED ISSUE #580: Allow multiple Twilio credentials per-organization
It should be possible for each organization to have different twilio credentials and accounts.  This is especially useful to more federated instances, where organizations are completely separate. Tasks to accomplish this include:

- [ ] Adding a column to organization with the message service sid we can lookup and thus locate which org it should go to.  (add to src/migrations/index.js and src/server/models/organization.js)
- [ ] An interface that is visible to superadmins and also organization OWNERs if an environment variable TWILIO_ORG_CONFIG is set, that allows you to enter Twilio credentials for the org.  This includes an additional component on the frontend and probably a new graphql api point on the backend.
- [ ] updating src/server/api/lib/twilio.js to lookup the organization sid that comes in with messages
- [ ] updating sendMessage in the same file to be sent with the appropriate twilio creds for the organization attached to the message.

jeffm2001:
# Fixes #580 

## Description

This allows each organization to configure their own Twilio credentials including: API Key, Auth Token, and default (currently only) Message Service SID. If the organization doesn't configure these values it will fall back to the global credentials set in environment variables (if they exist).

For security, Twilio Auth Tokens are encrypted using the SESSION_SECRET environment variable before being stored in the database. This is to prevent leakage through, e.g. database backups.

To make this encryption process simpler, I have also added a Twilio Credentials section to the organization settings form.

2020-04-08 at 5 56 PM

# Checklist:


src/server/api/lib/crypto.js
```diff
@@ -0,0 +1,36 @@
+import { getConfig } from "./config";
+const crypto = require("crypto");
+
```
schuyler1d, line 10: Can you document at the top of this file where you got these functions.

I'm not sure there's much beyond the base crypto library that's dependable, but maybe linking to a major company's library that implements the same?

src/server/api/lib/crypto.js
```diff
@@ -0,0 +1,36 @@
...
+  const iv = crypto.randomBytes(16);
+  const cipher = crypto.createCipheriv(algorithm, key, iv);
+  let encrypted = cipher.update(value);
+  encrypted = Buffer.concat([encrypted, cipher.final()]);
+  return iv.toString('hex') + ':' + encrypted.toString('hex');
+};
+
+const symmetricDecrypt = encrypted => {
+  let parts = encrypted.split(':');
+  const iv = Buffer.from(parts.shift(), 'hex');
```
schuyler1d, line 39: any reason to use hex vs. base64?

src/server/api/lib/twilio.js
```diff
@@ -7,45 +7,53 @@ import {
...
-function webhook() {
-  log.warn("twilio webhook call"); // sky: doesn't run this
-  if (twilio) {
-    return Twilio.webhook();
-  } else {
-    log.warn("NO TWILIO WEB VALIDATION");
-    return function(req, res, next) {
-      next();
-    };
+const getTwilioAuth = async organization => {
```
schuyler1d: So, I think all this logic should be in the organization.getTwilioAuth -- no reason to split up the logic into two separate files.

src/server/models/cacheable_queries/organization.js
```diff
@@ -18,22 +18,10 @@ const organizationCache = {
...
-    }
+  getTwilioAuth: async organization => {
     return {
-      messagingServiceSid: getConfig(
-        "TWILIO_MESSAGE_SERVICE_SID",
-        organization
-      ),
-      authToken: getConfig("TWILIO_API_KEY", organization),
-      apiKey: getConfig("TWILIO_AUTH_TOKEN", organization)
+      authToken: getConfig("TWILIO_AUTH_TOKEN", organization),
```
schuyler1d: I think what we should maybe do is have different values that are encrypted vs. not encrypted -- so first we should test for `hasConfig("TWILIO_AUTH_DATA_ENCYPRTED")` and if so, then decrypt things here and pass it back (or throw if decryption fails).  If not, then we fallback and do getConfig("TWILIO_AUTH_TOKEN") -- and then it can come from environment variables.

src/containers/Settings.jsx
```diff
@@ -217,6 +284,7 @@ class Settings extends React.Component {
           </CardActions>
         </Card>
         <div>{this.renderTextingHoursForm()}</div>
+        <div>{this.renderTwilioAuthForm()}</div>
```
schuyler1d: Can we condition this on an global environment variable allowing this?

src/server/api/schema.js
```diff
@@ -632,6 +633,31 @@ const rootMutations = {
...
+        twilioMessageServiceSid
+      }, {
+        user
+      }
+    ) => {
+      await accessRequired(user, organizationId, "OWNER");
+
+      const organization = await Organization.get(organizationId);
+      const featuresJSON = JSON.parse(organization.features || "{}");
+      featuresJSON.TWILIO_API_KEY = twilioApiKey;
```
schuyler1d: for all of these values, can you accept a max-length.  i.e. `featuresJSON.TWILIO_AUTH_TOKEN = twilioAuthToken.substr(0, 1024)`

schuyler1d (commented):
cool!  added some comments/details.

src/server/api/lib/crypto.js
jeffm2001: This is using the base crypto library. I took the structure from [politics-rewired](https://github.com/politics-rewired/Spoke/blob/master/src/server/api/lib/crypto.js), except I updated it to use the more secure/non-deprecated versions of cipher functions.

The one catch here security-wise is I had to pad the SESSION_SECRET to ensure it's 32 bytes, since the default in .env.example is shorter. But in any production environment that value will be long and randomly generated.

src/server/api/lib/crypto.js
jeffm2001: I think the IV needs to be hex because it needs to stay exactly 16 bytes, but the encrypted text could probably be base64, if there's a reason to prefer that.

src/containers/Settings.jsx
jeffm2001: I actually don't see an easy way to do that. Global variables don't seem to be available in the component, so then the only way would be to load that with the Organization. Although I could be missing something.

src/server/api/schema.js
jeffm2001: I put a max length on the frontend. That work?
e.g. `apiKey: yup.string().nullable().max(64)` in Settings.jsx

src/containers/Settings.jsx
jeffm2001: nevermind, I figured it out.

src/server/api/schema.js
schuyler1d: right -- let's ALSO have a max on the backend.  64 is even a better max (though are you sure that's sufficient?  How small are auth tokens?) -- it both communicates in the backend code what is expected and also stops some (admittedly obscure, but nonetheless) vulnerabilities.

src/server/api/lib/crypto.js
schuyler1d: Ok, how about you cite:
https://github.com/nodejs/node/blob/master/test/parallel/test-crypto-cipheriv-decipheriv.js
and (while keeping algorithm = "aes-256-cbc"), adapt the code to be similar.

My main point is that I want someone reading this code to see a good source that is likely vetted.  PoliticsRewired doesn't count -- a company that can afford security reviews, basically.

src/server/api/lib/twilio.js
```diff
@@ -16,36 +16,27 @@ import { saveNewIncomingMessage } from "./message-sending";
...
-
-function webhook() {
-  log.warn("twilio webhook call"); // sky: doesn't run this
-  if (twilio) {
-    return Twilio.webhook();
-  } else {
-    log.warn("NO TWILIO WEB VALIDATION");
-    return function(req, res, next) {
-      next();
+const SKIP_TWILIO_VALIDATION = process.env.SKIP_TWILIO_VALIDATION || global.SKIP_TWILIO_VALIDATION;
```
schuyler1d: this is a nit, but can you rename this so it's `TWILIO_SKIP_VALIDATION` -- i.e. all TWILIO env args should start with TWILIO (I know we're not good/consistent with this)

__test__/server/api/lib/twilio.test.js
```diff
@@ -257,6 +266,21 @@ it("postMessageSend+erroredMessageSender network error should decrement on err/f
   }
 });
 
+it("orgs should have separate twilio credentials",
```
schuyler1d, line 269: can you add some tests for encrypt/decrypt methods (pure) as well?

schuyler1d (commented):
So besides my comments so far (sorry, I should have done it all in a 'review'), I think in the case of TWILIO_MULTI_ORG (where there's no fallback), we need some way to block starting a campaign until they have set the credentials -- otherwise texting will unexpectedly and strangely fail way after they started.

This is a bit challenging because there are so many interacting variables in play with fallbacks (if service isn't twilio for the org or globally, e.g. then it  .  I'm tempted to suggest we have a `campaign` api method (in src/server/api/campaign.js) that says 'startBlockers: { twilioCredentials: true }` or something, which if it hasn't been set, then you can't click 'start campaign' but you have a message near the button saying why.  -- I'm open to other structures/places -- but basically -- how do we make sure folks fill that out before starting?

src/server/models/cacheable_queries/organization.js
```diff
@@ -18,23 +19,16 @@ const organizationCache = {
...
-        "TWILIO_MESSAGE_SERVICE_SID",
-        organization
-      ),
-      authToken: getConfig("TWILIO_API_KEY", organization),
-      apiKey: getConfig("TWILIO_AUTH_TOKEN", organization)
-    };
+  getTwilioAuth: async organization => {
+    const multiOrg = getConfig("TWILIO_MULTI_ORG");
+    const hasOrgToken = hasConfig("TWILIO_AUTH_TOKEN_ENCRYPTED", organization);
+    const authToken = multiOrg && hasOrgToken
```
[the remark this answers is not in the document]
jeffm2001: That's actually how I had it originally, but then I changed it, with the thinking that if someone disables TWILIO_MULTI_ORG, the intuitive behavior would be to revert to always using thing global credentials.

That's probably an edge case though and I'm fine with changing it to what you suggest.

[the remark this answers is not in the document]
jeffm2001: Actually, one question for you. Should it be:
`const authToken = hasOrgToken 
? symmetricDecrypt(getConfig("TWILIO_AUTH_TOKEN_ENCRYPTED", organization))
: getConfig("TWILIO_AUTH_TOKEN", organization)`

Or:
`const authToken = hasOrgToken 
? symmetricDecrypt(getConfig("TWILIO_AUTH_TOKEN_ENCRYPTED", organization))
: getConfig("TWILIO_AUTH_TOKEN")`

The former would allow folks to store unencrypted auth tokens in the database, which I think is bad practice, but perhaps they don't care (and they would have to manually insert that value in the database).

The latter would always fallback to the environment variable if the encrypted token wasn't present.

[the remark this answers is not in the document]
schuyler1d: I suppose I'm ok with either.  I think if we do the latter (forbidding non-encrypted tokens in the db), then we should really write up some good documentation on how to use your dev-tools/ commands and add something to the database.

jeffm2001:
> I think in the case of TWILIO_MULTI_ORG (where there's no fallback), we need some way to block starting a campaign until they have set the credentials -- otherwise texting will unexpectedly and strangely fail way after they started.

I took a shot at this. Not sure if it's the best way to do it, but maybe good enough for now? I think this might be more naturally addressed with the automatic number provisioning work (if there's no credentials, no numbers can be obtained). Also, I noticed that the `disabled` attribute on the button was not actually preventing the campaign from starting, so tried to fix that too.

Other than that, I think I've addressed everything you flagged (though I welcome more feedback on any/all of it).

schuyler1d:
This is great -- can you do a pull from MoveOnOrg:scaling-feb20-3-message and resolve conflicts?

Thanks for putting in the extra work to make this mergeable.

jeffm2001:
All set!

schuyler1d (dismissed):
This looks good!

src/containers/Settings.jsx
```diff
@@ -143,6 +148,85 @@ class Settings extends React.Component {
...
+    messageServiceSid
+  }) => {
+    await this.props.mutations.updateTwilioAuth(
+      apiKey,
+      authToken === '<Encrypted>' ? false : authToken,
+      messageServiceSid
+    );
+  };
+
+  renderTwilioAuthForm() {
```
matteosb, line 173: I'm concerned that allowing all fields to be updated after the organization is set up leaves a lot of room for misconfiguration. Maybe add a confirmation step if someone updates a per-existing Twilio Account SID or Messaging Service since updating either will break all existing campaigns? I think it's probably worth throwing in a simple twilio request in the update mutation to validate that the settings are correct too.

Another approach to make changing configuration easier would be to preserve the previous settings for existing campaigns. We did that for Messaging Services in the Warren fork to facilitate migrating from global to org-wide and finally to campaign-specific messaging services. The sid is written to the campaign table in [start_campaign](https://github.com/Elizabeth-Warren/Spoke/blob/master/src/server/workers/start-campaign.js#L63-L78) and used in [send message](https://github.com/Elizabeth-Warren/Spoke/blob/master/src/server/api/lib/twilio.js#L94-L105). I think we should do the same when implementing per campaign messaging services (#1495) cc/ @hiemanshu. It's probably overkill to store auth settings since I would expect the account sid to change infrequently once the org is set up.

matteosb (commented):
Should this be merged into `main` rather than `stage-main` @schuyler1d?

src/api/organization.js
```diff
@@ -11,5 +11,9 @@ export const schema = `
     textingHoursEnforced: Boolean
     textingHoursStart: Int
     textingHoursEnd: Int
+    twilioApiKey: String
```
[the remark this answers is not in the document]
jeffm2001: I was trying to maintain consistency with the `TWILIO_API_KEY` environment variable, although I totally agree that it's misnamed.

I would propose that if we change it, we change it across the board. `getTwilioAuth()` could check for both variable names to preserve backward compatibility.

src/containers/Settings.jsx
jeffm2001: Another option is the way they're doing it in the rewired fork — they have a `messaging_service` table and look up the account SID/token based on the campaign-configured messaging service.

That felt like overkill to me, and isn't easily backwards compatible. I think a warning is sufficient, and a good idea.

src/api/organization.js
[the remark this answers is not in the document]
matteosb: that sounds good to me!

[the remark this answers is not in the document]
schuyler1d: I want to preserve backwards compatibility, but if we want to rename the values loaded/saved in encrypted contexts, that might be fine.

I agree this is confusing, but outside of maintaining backward compatibility, I'm open to anything that makes sense to both of you.  Either way, let's document-document-document.

src/containers/Settings.jsx
schuyler1d: Defer to @matteosb on the table (and what to do here, vs later changes). Whatever we do, let's note that it IS possible to rotate credentials, so there's at least some case where you might want to update creds (at least the SECRET) on an existing messageservicesid

schuyler1d:
> Should this be merged into `main` rather than `stage-main` @schuyler1d?

So PRs should always be against `main` -- stage-main is an administrative branch that each cycle we destroy and recreate, so unless we specifically ask (e.g. during fixes during a QA/testing cycle) everything should be to `main`.

Also, Jeff, now that the scaling-feb20-3-message branch was merged to main, I switched the PR.  please merge current main (Spoke 5.4) and resolve conflicts.

src/containers/Settings.jsx
matteosb: I'm happy with just having a warning. We can make it so that changing the messaging service doesn't break existing campaigns once @hiemanshu's changes are in.

@schuyler1d yes, absolutely, the changing the auth token should be fully supported and I don't think it requires a warning.

@jeffm2001 agree, i think something like the politics-rewired messaging service table would be a good solution if we want to support switching twilio accounts without breaking campaigns but it's a bit overkill and makes migration more complicated.

src/api/organization.js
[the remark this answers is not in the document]
matteosb: Ok, i think renaming variables and aliasing the env var/org setting can be done later as a chore. For this commit I do think the user-facing copy should be updated to say "Twilio Account SID" to make it easier for admins.

src/containers/Settings.jsx
```diff
@@ -217,9 +249,16 @@ class Settings extends React.Component {
...
-                type="submit"
                 label={this.props.saveLabel || "Save Twilio Credentials"}
+                onClick={this.handleOpenTwilioDialog}
               />
+              <Dialog
+                actions={dialogActions}
+                modal={true}
+                open={this.state.twilioDialogOpen}
+              >
+                Changing Twilio credentials will break any campaigns that are currently running. Do you want to contunue?
```
matteosb: clarify that changing Account SID or Messaging Service SID, but not Auth Token, will break existing campaigns.

src/api/organization.js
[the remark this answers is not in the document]
jeffm2001: I actually made the change before I saw this comment. I can revert that commit if you think it's too much for this PR.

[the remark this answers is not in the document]
matteosb: oh cool, the env var change looks good to me

jeffm2001:
I think I addressed all the feedback, and resolved the merge conflicts with `main`. Let me know if you see anything else!

schuyler1d:
Made a branch PR: https://github.com/aflcio/Spoke/pull/1 with some tweaks to the api -- basically just limiting to org OWNER for the twilio creds and never sending the authToken, even encrypted.

schuyler1d:
This was merged into `main` yesterday for 5.5 but I think the commit order meant that this PR didn't close automatically, but it's IN!

PR #1482, opened 2020-04-10
readme.md - change example invite hashes to UUID

### LINKED ISSUE #1479: "Getting the App Running" steps incorrect example SQL
In the `readme.md`'s "Getting the App Running" steps (https://github.com/MoveOnOrg/Spoke#getting-the-app-running), a sample SQL snippet implies any string can be used for invite codes:

```
echo "INSERT INTO invite (hash,is_valid) VALUES ('abc', 1);" |sqlite3 mydb.sqlite
# Note: When doing this with PostgreSQL, you would replace the `1` with `true`
```

While I was deploying to production with `passport_strategy=local`, I was getting error messages in the login API call headers saying invite code was invalid. I dived into the code and [found that the hash should be a UUIDv4 in order to initially register a login for Spoke](https://github.com/MoveOnOrg/Spoke/blob/8642cd3413101c2df5e5c4be5f550ebe63d10ac7/src/server/local-auth-helpers.js#L49-L59). Should the docs be updated to say this?

Thanks for your work developing Spoke,
Kyle

kylehotchkiss:
# Fixes #1479

## Description

While I was deploying to production with an ENV of `passport_strategy=local`, I was getting error messages in the login API call headers saying invite code was invalid. I dived into the code and found that the hash should be a UUIDv4 in order to initially register a login for Spoke. I have updated the readme's "Getting the App Running" section to reflect that the invite hash should be a UUIDv4

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,10 +71,10 @@ At this point, you should be ready to start your app in development mode.
 3. As long as you leave `SUPPRESS_SELF_INVITE=` blank in your `.env` you should be able to invite yourself from the homepage.
    - If you DO set that variable, then spoke will be invite-only and you'll need to generate an invite. Run:
      ```
-     echo "INSERT INTO invite (hash,is_valid) VALUES ('abc', 1);" |sqlite3 mydb.sqlite
+     echo "INSERT INTO invite (hash,is_valid) VALUES ('E4502B28-301E-4E63-9A97-ACA14E8160C8', 1);" |sqlite3 mydb.sqlite
      # Note: When doing this with PostgreSQL, you would replace the `1` with `true`
      ```
-   - Then use the generated key to visit an invite link, e.g.: http://localhost:3000/invite/abc. This should redirect you to the login screen. Use the "Sign Up" option to create your account.
+   - Then use the generated key to visit an invite link, e.g.: http://localhost:3000/invite/E4502B28-301E-4E63-9A97-ACA14E8160C8. This should redirect you to the login screen. Use the "Sign Up" option to create your account.
 4. You should then be prompted to create an organization. Create it.
 5. See the [Admin and Texter demos](https://opensource.moveon.org/spoke-p2p#block-yui_3_17_2_25_1509554076500_36334) to learn about how Spoke works.
 6. See [Getting Started with Development](#more-documentation) below.


README.md
```diff
@@ -71,10 +71,10 @@ At this point, you should be ready to start your app in development mode.
 3. As long as you leave `SUPPRESS_SELF_INVITE=` blank in your `.env` you should be able to invite yourself from the homepage.
    - If you DO set that variable, then spoke will be invite-only and you'll need to generate an invite. Run:
```
ibrand, line 72:
```suggestion
   - If you DO set that variable, then spoke will be invite-only and you'll need to generate an invite. Make sure that your invite hash is UUID encoded to pass the validation. Run:
```

README.md
```diff
@@ -71,10 +71,10 @@ At this point, you should be ready to start your app in development mode.
...
      ```
-     echo "INSERT INTO invite (hash,is_valid) VALUES ('abc', 1);" |sqlite3 mydb.sqlite
+     echo "INSERT INTO invite (hash,is_valid) VALUES ('E4502B28-301E-4E63-9A97-ACA14E8160C8', 1);" |sqlite3 mydb.sqlite
      # Note: When doing this with PostgreSQL, you would replace the `1` with `true`
```
ibrand, line 75:
```suggestion
     # Notes: When doing this with PostgreSQL, you would replace the `1` with `true`, in this example the invite is a hashed version of the string 'abc'
```

README.md
```diff
@@ -71,10 +71,10 @@ At this point, you should be ready to start your app in development mode.
...
      ```
-   - Then use the generated key to visit an invite link, e.g.: http://localhost:3000/invite/abc. This should redirect you to the login screen. Use the "Sign Up" option to create your account.
+   - Then use the generated key to visit an invite link, e.g.: http://localhost:3000/invite/E4502B28-301E-4E63-9A97-ACA14E8160C8. This should redirect you to the login screen. Use the "Sign Up" option to create your account.
```
ibrand, line 77: Does this actually name the organization "abc" or does it end up naming the org `E4502B28-301E-4E63-9A97-ACA14E8160C8`?

I'm a little concerned to see the hash in the invite url. I'm pretty sure when I ran `echo "INSERT INTO invite (hash,is_valid) VALUES ('abc', 1);"` on postgres it *did* actually generate an org called "abc"

ibrand (commented):
Left some questions here. Thanks for being proactive about this!

README.md
schuyler1d: it shouldn't create an org called 'abc' (or the hash) -- `hash` is a column in the `invite` table -- so it would be very odd if it populated the org name.

@ibrand Can you confirm that?

README.md
schuyler1d: Why does it have to be UUID encoded?  Does it really not work without that?

Honestly, looking at the createOrganization and inviteByHash code in server/api/schema.js, I have no idea how this works at all -- e.g. createOrganization does a load by `inviteId` but that defeats the whole purpose of testing based on a hash, so this code needs some work.
If we have a volunteer interested in this, I can leave some notes -- we could fold fixing this into the warren improvement with a uuid per-campaign.

kylehotchkiss: Hi @schuyler1d - when i tried to setup an initial account (before the initial org was created), i found the initial signup failing, with a header saying `abc` was not a valid invite code. When i used a UUID in the SQL insert instead of `abc`, the initial invite code then did work.

I dug around the code and found that invite codes must be UUIDs, according to this validation code: https://github.com/MoveOnOrg/Spoke/blob/8642cd3413101c2df5e5c4be5f550ebe63d10ac7/src/server/local-auth-helpers.js#L49-L59

Might be worth adding some notes in ticket #1479.

README.md
kylehotchkiss: I can't claim to be super experienced with Spoke's codebase but i found that the hash here was not associated with the company name, as once I signed up using this hash, I was then given the prompt for company name and proceeded to create the first login from there.

The hash I provided was randomly generate just from the `uuidgen` cli command and is not the hash of `abc`.

ibrand: I reran this and was mistaken! My bad here. I was misremembering the last time we used this command. This *does* only impact the invite table and, like @kylehotchkiss said, it works when it's encoded

README.md
ibrand: I am also seeing that when I experiment with this locally the UUID encoding works and is how the other invites in the DB are stored, while the non-encoded version does not work.

schuyler1d: ok, @kylehotchkiss -- thanks!  I missed where that was validated!
Let's definitely plan to fix this further, but I'm :+1: on the documentation change here until we do.

README.md
schuyler1d: This isn't a hash of 'abc' -- it's just a random uuid.

schuyler1d (approved):
I'm pretty sure this is ready now, so I'm marking approved.

Thanks @kylehotchkiss for your patience, while we figured out what our own system does -- clearly it's important to document!

README.md
ibrand: Ah yep, after the conversation we had in the other comment I realized I was mixed up. I'll resolve!

PR #1491, opened 2020-04-19
dev-tools: Increase consistency of development entrypoints

tmc:
This changes the development entry points to consistently use dotenv to load values from `.env`.

### DIFF
--- a/dev-tools/babel-run
+++ b/dev-tools/babel-run
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
-require('babel-register')
-require('babel-polyfill')
-require('../' + process.argv[2])
+require("dotenv").config();
+require("babel-register");
+require("babel-polyfill");
+require("../" + process.argv[2]);
--- a/dev-tools/babel-run-debug
+++ b/dev-tools/babel-run-debug
@@ -1,5 +1,6 @@
 #!/bin/sh
 ":" //# comment; exec /usr/bin/env node --inspect "$0" "$@"
+require("dotenv").config();
 require('babel-register')
 require('babel-polyfill')
 require('../' + process.argv[2])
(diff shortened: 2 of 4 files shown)


schuyler1d:
Which commands does this enable .env with?  All babel commands now?
`.env` is really meant for dev/debug environments.  Many places it's avoided are to ensure that a dev `.env` file doesn't override a production deploy environment.

It also needs to avoid *breaking* things (e.g. in production/deploy envs) if a .env file doesn't exist -- I know at one point doing `require("dotenv").config();` would break if it wasn't present, but not sure if that's still the case.

tmc:
> Which commands does this enable .env with? All babel commands now?
> `.env` is really meant for dev/debug environments. Many places it's avoided are to ensure that a dev `.env` file doesn't override a production deploy environment.
> 
> It also needs to avoid _breaking_ things (e.g. in production/deploy envs) if a .env file doesn't exist -- I know at one point doing `require("dotenv").config();` would break if it wasn't present, but not sure if that's still the case.
That is still the case. The entrypoints I modified are in a `dev-tools` directory but that directory is actually used for these `package.json` entry points:

* process-message-csv
* dev-message-sender-01
* prod-message-sender-01
* dev-message-sender-234
* prod-message-sender-234
* dev-message-sender-56
* prod-message-sender-56
* dev-message-sender-789
* prod-message-sender-789
* dev-job-handler
* prod-job-handler
* dev-incoming-message-handler
* prod-incoming-message-handler
* dev-start
* dev-start-debug
* generate-contacts

some of these could be targeted in production environments with out a `.env` so this change may probably shouldn't go in.

I do think it's a conceptual mismatch to have a `dev-tool` on a prod path. Closing.

PR #1497, opened 2020-04-22
Add schema changes to the RFC template

matteosb:


ibrand (dismissed):
This is a great addition :+1:

ibrand (approved):

PR #1500, opened 2020-04-22
Add title to campaign edit

### LINKED ISSUE #840: Add the Title of the campaign above "This campaign is running!"
It would be nice to have the name of the campaign above "This campaign is running."

image

higgyCodes:
# Fixes #840 

## Description

As above, a simple one liner to add the campaign title to the campaign edit view. Title updates if a valid title is submitted successfully in the "Basics" section.

image

### DIFF
--- a/src/containers/AdminCampaignEdit.jsx
+++ b/src/containers/AdminCampaignEdit.jsx
@@ -508,6 +508,7 @@ class AdminCampaignEdit extends React.Component {
           fontSize: 16
         }}
       >
+        <h2>{this.props.campaignData.campaign.title}</h2>
         {this.state.startingCampaign ? (
           <div
             style={{
--- a/src/containers/AdminCampaignEdit.jsx
+++ b/src/containers/AdminCampaignEdit.jsx
@@ -503,7 +503,9 @@ class AdminCampaignEdit extends React.Component {
 
     return (
       <div>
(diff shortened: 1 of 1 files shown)


src/containers/AdminCampaignEdit.jsx
```diff
@@ -502,31 +502,34 @@ class AdminCampaignEdit extends React.Component {
...
-          <div
-            style={{
-              color: theme.colors.gray,
-              fontWeight: 800
-            }}
-          >
-            <CircularProgress
-              size={0.5}
+      <div>
+        <h2>{this.props.campaignData.campaign.title}</h2>
```
schuyler1d: what happens with a new campaign?  Maybe we should condition the `h2` on a title exisitng?

schuyler1d (dismissed):
I think this is fine as-is, but maybe better to just condition the h2 on a title being present

src/containers/AdminCampaignEdit.jsx
higgyCodes: A new campaign appears to always contain the title property value 'New Campaign' as seen below:

image

 Given that the value appeared to be guaranteed by the validation, I omitted any additional conditionality. If this feels presumptuous, I'd be happy to add it in.

schuyler1d (approved):
Cool, thanks!

Issue #1473, opened 2020-04-03
Feature Request: Admin Side Tagging


ibrand:
**Problem**
Admin want to be able to create tags for their organization that are usable for contacts across campaigns. This will allow them to collect more metadata on responses that aren't strictly survey responses.

**Solution**
- [x] Tags schema - #1462 
- [x] UI for creating, editing, and deleting tags
- [ ] Add a field to canned responses to allow adding tags to them so that those canned responses auto-tag when they are used

**Screenshots**
*Admin UI screenshots from the Politics Rewired backend*
<img width="1440" alt="tags1" src="https://user-images.githubusercontent.com/3953117/78413764-e6016380-75e6-11ea-972f-f36e5480bf49.png">
<img width="1440" alt="tags2" src="https://user-images.githubusercontent.com/3953117/78413768-e8fc5400-75e6-11ea-9ae4-f04a1e1273c3.png">

**Context**
[see the project board description for further context](https://github.com/MoveOnOrg/Spoke/projects/15)
This is a more in-depth re-visitation of #16 and we will close that ticket when this work is done.

schuyler1d:
This was done! It's currently behind EXPERIMENTAL_TAGS=1 environment variable, but we will show it by-default when tags can be wired up for interactions/canned responses/etc.

PR #1505, opened 2020-04-23
#1473: Feature Request: Admin Side Tagging

aschneit:
# Fixes # (issue)
#1473 
## Description

Implemented UI for Tags admin screen and related CRUD functionality (creating, editing, deleting tags). 

# Checklist:

Screen Shot 2020-04-23 at 11 54 27 AM


azuzunaga:
Taking a look now!

src/containers/Tags.jsx
```diff
@@ -0,0 +1,254 @@
...
+            defaultValue={dialogTag}
+          >
+            <div className={css(styles.fields)}>
+              <Form.Field label="Name" name="name" />
+              <Form.Field label="Description" name="description" />
+            </div>
+            <div className={css(styles.submit)}>
+              <Form.Button
+                type="submit"
+                label={dialogMode === "create" ? "Create" : "Save"}
```
azuzunaga: Would it make more sense to have `handleOpen` and `handleOpenEdit` set state values for `title`, `submitHandler`, and `buttonLabel`? It seems to me that would be cleaner (and more extensible) than the ternary expressions.

src/server/api/schema.js
```diff
@@ -1197,6 +1199,43 @@ const rootMutations = {
...
+        .update(tagUpdates);
+
+      return { id, ...tagUpdates };
+    },
+    deleteTag: async (_, { organizationId, id }, { user }) => {
+      await accessRequired(user, organizationId, "SUPERVOLUNTEER");
+      await r
+        .knex("tag")
+        .where("id", id)
+        .update({ is_deleted: true });
```
azuzunaga, line 1254: Maybe I missed something in the issue conversation, but why don't we actually delete the tags? Is it to avoid issues with cascading deletes in the `tag_campaign_contact` table?

src/server/models/index.js
```diff
@@ -57,9 +59,11 @@ const tableList = [
   "opt_out", // good candidate
   "pending_message_part",
   "question_response",
+  "tag",
+  "tag_campaign_contact",
   "user_cell",
   "user_organization",
-  "zip_code" // good candidate (or by contact)?
+  "zip_code" // good candidate (or by contact)?,
```
azuzunaga: Nit:
```suggestion
  "zip_code" // good candidate (or by contact)?
```

azuzunaga (commented):
Great work on this! This feature has been requested a bunch, glad it is getting implemented!

I have a few questions before I approve, let me know what you think.

Thanks!

src/server/middleware/render-index.js
```diff
@@ -86,6 +86,7 @@ export default function renderIndex(html, css, assetMap, store) {
       window.PASSPORT_STRATEGY="${process.env.PASSPORT_STRATEGY || ""}"
       window.CORE_BACKGROUND_COLOR="${process.env.CORE_BACKGROUND_COLOR || ""}"
       window.CAN_GOOGLE_IMPORT=${canGoogleImport}
+      window.SHOW_TAGS=${process.env.SHOW_TAGS || 0}
```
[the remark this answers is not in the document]
aschneit: Seems like the other boolean vars above (NOT_IN_USA and ALLOW_SEND_ALL) have 0 as their default if the variable isn't set. But I can change it if that doesn't seem right. (also, this environment variable is also just temporary, so that Tags option won't show up in the admin sidebar until we get the texter tag ui in place)

src/components/AdminDashboard.jsx
```diff
@@ -88,13 +88,22 @@ class AdminDashboard extends React.Component {
...
       {
         name: "Settings",
         path: "settings",
         role: "SUPERVOLUNTEER"
       }
     ];
 
+    if (window.SHOW_TAGS === false) {
+      sections = sections.filter(section => section.name !== "Tags");
+    }
```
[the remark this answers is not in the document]
aschneit: See above, since this is probably just a temporary variable, I thought just being able to remove the false conditional when we no longer need it would be easier. But I'm open to changing it if it seems messy. @ibrand, we'll probably just remove this env. variable when we get the texter tag ui set to go,  right?

src/server/api/schema.js
aschneit: @ibrand  gave me part of the explanation for this, but I wasn't totally clear on it and there may be more that I'm not aware of. Ilona?

src/api/schema.js
```diff
@@ -50,6 +50,7 @@ import {
   schema as cannedResponseSchema,
   resolvers as cannedResponseResolvers
 } from "./canned-response";
+import { schema as tagSchema, resolvers as tagResolvers } from "./tag";
```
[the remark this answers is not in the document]
aschneit: removed that

src/containers/Tags.jsx
aschneit: added those, thanks

src/server/models/index.js
aschneit: removed

src/server/middleware/render-index.js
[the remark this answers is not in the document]
azuzunaga: Gotcha. Let's set it to `false` then, since we check explicitly for a `false` value in the admin dashboard.

src/components/AdminDashboard.jsx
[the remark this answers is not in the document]
azuzunaga: If we are removing that if block anyway then we can leave it as is. Thanks!

azuzunaga (dismissed):
Looks great @aschneit! Thanks for pushing up the changes so quickly. I just left a final comment on the `window.SHOW_TAGS=${process.env.SHOW_TAGS || 0}` piece of code, but it is not a blocker.

I'm interested to hear about the `is_deleted` field, but again, not blocking.

🥳

azuzunaga (dismissed):
Looks like the last commit dismissed the previous approval 🙂

src/server/api/schema.js
ibrand: Hi @aschneit and @azuzunaga thanks for the ping :)

I encouraged a soft delete because if tags are soft deleted but still exist attached to other things (like canned responses) then no other data will be messed up. Secondarily, I figure an admin could accidentally delete one and want it back.

I think that first situation is a lot more important because it could lead to some bad data situations for a tag to no longer exist but a lot of the data export has tags already in it.

azuzunaga: Thanks for the explanation @ibrand! Makes sense to me.

src/server/models/tag-campaign-contact.js
```diff
@@ -0,0 +1,28 @@
...
+      updated_at: timestamp()
+    })
+    .allowExtra(false),
+  {
+    noAutoCreation: true,
+    depenencies: [Tag, CampaignContact]
+  }
+);
+
+TagCampaignContact.ensureIndex("campaign_contact_id");
```
matteosb, line 26: These don't do anything right now and knex migrations are the source of truth, so I would remove them 

cc/ @schuyler1d. If we want to continue adding these for some reason, then let's merge this PR: https://github.com/MoveOnOrg/rethink-knex-adapter/pull/28

schuyler1d (dismissed):
:+1:

aschneit:
@schuyler1d, @azuzunaga, @matteosb, I believe I resolved all of your comments (but the approvals seem to have disappeared after I pushed new commits). The only one I didn't address is the one about ensureIndex in the TagCampaignContact model, since that still seemed like an open question contingent on merging that other pr...and the models/migrations weren't commits I actually worked on. But I can make the change if people want, just let me know, thanks!

src/containers/Tags.jsx
```diff
@@ -0,0 +1,280 @@
...
+    this.handleClose();
+  }
+
+  async handleEdit(formData) {
+    await this.props.mutations.editTag(
+      formData,
+      this.props.params.organizationId,
+      this.state.tagId
+    );
+    await this.props.data.refetch();
```
aschneit, line 106: @matteosb I added this line in all of the mutation event handlers to refetch data after the mutation. Without it, dom updates don't seem to work. But if I made the changes you suggested above for updating the cache correctly after a mutation, should a refetch be necessary? My experience with GraphQL is limited, so I may be missing something. Thanks

matteosb: Interesting, this shouldn't be necessary if you are returning the fields you are updating from the mutation but we are currently on an ancient version of apollo-client, so it might be broken. It's fine to leave this in here for now. I'm going to do a pass on this once I upgrade apollo-client (see the discussion here: #1461)

aschneit: @matteosb Yeah, I noticed there isn't an apollo-cache package installed, and I wasn't totally clear on the caching strategies being used overall. I guess we'll see what can be changed with the upgrade. Thanks!

schuyler1d:
I believe everything in this PR was merged into what will be released as Spoke 5.5.  I think commit order meant that it isn't showing as merged, but it is.

PR #1509, opened 2020-04-24
track campaign.errorCount in cache

schuyler1d:
# Just like messagedCount and assignedCount, let's track errorCount in cache for fast reporting

## Description

In order to be able to quickly see how many errors have occurred per-campaign, we should cache this value.  This PR moves deliveryReport into the cacheableData to simplify twilio.js further and then increments errorCount just like in the past.

# Checklist:


schuyler1d:
cc @hiemanshu -- don't think this should conflict too much, but FYI.

Issue #1518, opened 2020-04-29
RFC: Phone Number Management


matteosb:
## Proposal

To phone number management to the Spoke admin interface similar to phone number provisioning in the [Warren fork](https://github.com/Elizabeth-Warren/Spoke#phone-number-provisioning). We would initially provide support for `twilio` and `fakeservice`, but leave the design open to integration with more SMS providers. Unlike the Warren version, the phone numbers will belong to individual organizations, though we may want to eventually add an option to allow transfer between organizations for single-tenant deployments.

Phone number management will be decoupled from how numbers are used, but we will start by integrating it with per-campaign messaging services introduced in #1495 . In the future, we may want to integrate it with per-organization messaging services and other SMS providers.

Twilio numbers provisioned through Spoke will have a configurable twilio voice url, resolving issue #1464 

## Motivation

Managing numbers in Twilio is a tedious manual process and a barrier to adoption for newcomers and non-technical folks who may be discouraged by a CLI like switchboard-twilio. This gets even more complicated for large organizations that need to scale beyond a single Messaging Service, which Spoke doesn't currently allow. #1495 allows organizations to scale but would require a lot of manual work without the additions described here. Some organizations may also want the added control over what area codes they use for individual campaigns rather than relying on Twilio's geomatch.
 
## Detailed Design

The Warren fork README has a fairly detailed explanation of phone number management combined with per-campaign messaging services: https://github.com/Elizabeth-Warren/Spoke#phone-number-provisioning.

We will depart from the Warren design in a few key ways:
- Phone numbers will belong to an organization, not the entire Spoke instance
- The phone number inventory table will support multiple SMS providers
- Phone numbers will only have two states: "ALLOCATED" or "AVAILABLE". We may want to add a phone number reservation system like the Warren fork, but it isn't necessary and can be added later.

The work will be divided into three tasks:
- Merge the core inventory management feature (admin screen to buy numbers and resolver to see what phone numbers the organization has available) #1525 @matteosb 
- Merge per-campaign Twilio messaging services #1539 @hiemanshu 
- Integrate per-campaign Twilio messaging service with the phone number inventory by allocating numbers to a campaign when it starts. @matteosb 

## Alternatives

See the discussion [here](https://github.com/MoveOnOrg/Spoke/issues/1495#issuecomment-617222603). The two other options described are:

[...]

PR #1525, opened 2020-05-01
Phone number inventory management

matteosb:
First part of #1518.

This adds the phone number inventory table and basic admin functionality to list and buy numbers. It also allows numbers to be added directly to organization-scoped Twilio Messaging Services.

Note: allowing individual campaigns to create messaging services with numbers from the inventory will be added in a later release.


src/server/api/lib/twilio.js
```diff
@@ -10,6 +10,8 @@ import {
 import { log } from "../../../lib";
 import { saveNewIncomingMessage } from "./message-sending";
 import { getConfig } from "./config";
+import _ from "lodash";
+import urlJoin from "url-join";
```
schuyler1d: doesn't look like you're using this method or library -- can we delete it after all?

src/server/api/lib/fakeservice.js
```diff
@@ -90,8 +91,24 @@ async function handleIncomingMessage(message) {
   return part.id;
 }
 
+async function buyNumbersInAreaCode(organization, areaCode, limit) {
+  const rows = _.times(limit, () => ({
+    organization_id: organization.id,
+    area_code: areaCode,
+    phone_number: `+1${areaCode}555FAKE`,
```
schuyler1d: doesn't this need to be different for each row? -- I do like the TEXT in the number to make it 'really broken' -- maybe 555FK{random two digits}?
Similar to this code:
https://github.com/MoveOnOrg/Spoke/blob/main/src/integrations/contact-loaders/test-fakedata/index.js#L128-L130

Nit: Aesthetically, we try to avoid using lodash, when a native JS pattern is ok.  _.chunk seems like a decent exception.  I'm on the edge here -- I think I'd favor a boring for loop, so people don't need to lookup what `_.times` does.

src/server/middleware/render-index.js
```diff
@@ -91,6 +91,13 @@ export default function renderIndex(html, css, assetMap, store) {
       window.EXPERIMENTAL_TAGS=${process.env.EXPERIMENTAL_TAGS || false}
       window.EXPERIMENTAL_TEXTERUI="${process.env.EXPERIMENTAL_TEXTERUI || ""}"
       window.TWILIO_MULTI_ORG=${process.env.TWILIO_MULTI_ORG || false}
+      window.EXPERIMENTAL_PHONE_INVENTORY=${getConfig(
```
schuyler1d: one downside of using global window variables for these things is that it can't be per-org, then, because this globally loaded -- it might be better to have an organization graphql property that can return `getConfig('EXPERIMENTAL_PHONE_INVENTORY', organization)` and then condition on that (after adding to a (apparently not yet present) organization query in the components/AdminDashboard.jsx component

schuyler1d (commented):
Looking good so far!

matteosb:
@schuyler1d This is ready for a full review now. Screenshots:

PhoneNumbersForm
Screenshot from 2020-05-11 19-30-06
Screenshot from 2020-05-11 19-34-36

matteosb:
Added functionality to add purchased numbers to organization messaging services from the UI

Screenshot from 2020-05-15 16-53-32

cc @jeffm2001, this should work well with TWILIO_MULTI_ORG

src/server/api/lib/twilio.js
```diff
@@ -403,12 +410,126 @@ async function handleIncomingMessage(message) {
   }
 }
 
+/**
+ * Search for phone numbers available for purchase
+ */
+async function searchForAvailableNumbers(twilioInstance, areaCode, limit) {
+  const count = Math.min(limit, 30); // Twilio limit
+  return twilioInstance.availablePhoneNumbers("US").local.list({
```
schuyler1d: let's make this an environment variable `getConfig(COUNTRY_CODE, organization) || "US"`

src/server/api/schema.js
```diff
@@ -1298,6 +1299,52 @@ const rootMutations = {
         .where("id", id)
         .update({ is_deleted: true });
       return { id };
+    },
+    buyPhoneNumbers: async (
```
schuyler1d: @lperson has started what might be a good trend in creating these as mutations/ files rather than continue making the large schema.js file even bigger.  

Optional, but if you move it to one, I think that would be good.  If you do so, please move the whole function, so you can import and then just include `buyPhoneNumbers` as a unitary key-value in the list.

src/server/models/index.js
```diff
@@ -61,6 +61,7 @@ const tableList = [
   "question_response",
   "tag",
   "tag_campaign_contact",
+  "owned_phone_number",
```
schuyler1d, line 64: I know it's silly, but can you create a models/owned-phone-number.js file in the style of the old ones.  even if it's dead code, I'd like us to keep that directory documenting the current schema for all the models (rather than needing to search through changes in the migrations).

src/workers/jobs.js
```diff
@@ -1118,3 +1118,39 @@ export async function clearOldJobs(delay) {
...
+  try {
+    if (!job.organization_id) {
+      throw Error("organization_id is required");
+    }
+    const payload = JSON.parse(job.payload);
+    const { areaCode, limit, messagingServiceSid } = payload;
+    if (!areaCode || !limit) {
+      throw new Error("areaCode and limit are required");
+    }
+    const organization = await Organization.get(job.organization_id);
```
schuyler1d: please use `loaders` here, so cache can be used.

__test__/cypress/plugins/index.js
```diff
@@ -5,6 +5,20 @@ require("dotenv").load();
 require("babel-register");
 require("babel-polyfill");
 
+if (process.env.DEFAULT_SERVICE !== "fakeservice") {
+  throw Error(
```
schuyler1d: Even so, can we do conditionals on DEFAULT_SERVICE == 'fakeservice' for the number buying parts within the cypress tasks themselves?

schuyler1d (commented):
Looks good -- just a few nits.

src/workers/jobs.js
matteosb: `cacheableData`, not `loaders`, since we are outside of the request scope right?

src/server/api/lib/twilio.js
matteosb: Can I use the PHONE_NUMBER_COUNTRY env var, which appears to be used for formatting, or is there a reason to introduce a new one?

src/server/api/schema.js
matteosb: nice, i definitely prefer this pattern

matteosb:
@schuyler1  review comments addressed, mind taking another look?

Issue #1533, opened 2020-05-05
RFC: Request/Release-style dynamic assignment + Texter Sidebox extensibility


schuyler1d:
## Proposal
* Change dynamic assignment from incremental 'pull' as the texter is texting to a Batch Request/Release model
* Mostly do this with an extensible Texter Sidebox 

![spoke-texter-rightpanel](https://user-images.githubusercontent.com/52117/81089441-5fcc8b80-8eca-11ea-9fad-f926301fc061.png)

## Motivation
The [Warren Campaign Spoke Project](https://github.com/Elizabeth-Warren/Spoke) successfully deployed a great model to manage texters.  We'd like to make the following workflows possible in Spoke:
* New texters are invited through a (campaign-specific secure link) and can "Request a batch" (e.g. of 300)
* During texting, they are assigned the full 300 right away.
* It is easy for them to 'give up' and release their 'batch' for other people to text.

Furthermore, a lot of use-cases have come up both in the Warren campaign and other contexts for a Texter 'sidebox' -- where other widgets, etc can be contextually added. These 'request/release' buttons are an opportunity to structure this framework.  Some Texter Sidebox use cases include:
request a new batch button
* Release batch (texter can give up for other texters to take it on)
* Release replies (an optional button for texters to 'give up' on replies and pass the conversations on)
* "Great Job! you finished" **confetti** widget
* Survey iframe (The Warren campaign used this for shift signups)
* 'Add feedback' widget for Admins viewing the contact thread. (Something I believe WFP has discussed)

Because of recent [Texter UI work](https://github.com/MoveOnOrg/Spoke/issues/1493) there is also quite a bit of simplification to do to remove the older interfaces, so doing that simultaneously seems like a good opportune time.

## Detailed Design

* Per-campaign invites will move to the Warren model where there's a different uuid per campaign -- however I plan to just use the same invite link with the campaign uuid NOT the organization uuid, but instead the campaign uuid.  Thus the same request can match both and then join the campaign's organization at the same time.
* I plan to have a `sidebox` directory where new sideboxes are added.  The api will need to be worked out further in implementation but roughtly for each sidebox there will be:
  * an admin UI react component which returns config data
  * a texter UI react component for the sidebox widget -- in some cases, including the release/request buttons, this will be a container with a mutation graphql hook
[...]

PR #1534, opened 2020-05-05
AssignmentTexter refactor

schuyler1d:
# Part of https://github.com/MoveOnOrg/Spoke/issues/1533

## Description

This moves a BUNCH of files around -- every component that is only used for the texter's contact screen is moved into components/AssignmentTexter/

I think the main question in this PR is how do we want to organize components/containers further?
I think @lperson started doing work that grouped these things.  While this still groups things in containers/components division is it better to group even further and lose the distinction between container and component (which I don't think we're perfect on) or group them all together (in this case it would move the directory into containers/ and then everything would be under, e.g. TexterTodo/

Do folks have bikeshed thoughts on organization here?

# Checklist:


matteosb (approved):
Looks great, I'm all for this grouping. IMO the `containers`/ `components` distinction isn't particularly useful right now. I'd prefer to see `components` used only for those components that are designed to be re-usable and ideally stateless, while keeping all "private" components grouped with containers.

PR #1543, opened 2020-05-09
Polishing action handler framework

### LINKED ISSUE #1512: RFC: simple framework for providing client-choice data for action handlers
## Proposal
Action handlers are an existing lightweight framework in Spoke. They are modular and well defined.  Currently in Spoke, an admin can create a script that links an `action` to the answer to a question. When a tester selects that answer, the linked action executes on the server.  The use case for action handlers (to date) is to update something in an external system.  We currently have action handlers for Actionkit, Mobilecomm and Revere.

I propose to add a way to associate an optional, additional blob of data with question answers in Spoke so that the same action handler can do a generic something on a remote system, with specifics provided per question answer. 

## Motivation
The first use case is an integration with NGP/VAN. 

The generic thing we want to do is update a user in VAN with some information acquired when we interact with them. For VAN, the specific things we want to do are 
- apply an activist code to the contact
- designate the contact as having answered a specific survey question with a specific response
- memorialize an interaction with the contact by recording a canvass result

All of these things can be done with the same code -- i.e., the same action handler -- given specific data indicating the type of thing we want to do (one of the bullet points above) and details about the thing (such as survey question ID and survey response ID).

## Detailed Design
- Add column `answer_actions_data` to table `interaction_step`.
- Pass the contents of that column down to the interaction step editor in the UI.
- If a handler supports this client choice data (so named because it gives the admin user a choice in the script editor client), it will be provided with the list of available handlers.
  - We will require handlers that require client choice data to implement a method with a well-known name to provide the data on request
  - We will build a lightweight framework to wrap this data with a cache layer
  - We will add a single autocomplete field to the script editor that will appear and be enabled when a handler provides client choice data
  - We will save the admin user's selection of client choice data in the `answer_actions_data` column.
- Each handler already exposes a `processAction` member. That member already has access to the `interaction_step`.  Action handlers can implement `processAction` as necessary to use `answer_actions_data`.
- Wrap initialization, availability and client choice functions with a cache layer to avoid expensive calls to the database and especially external services for data that changes infrequently.

## Project Plan Information
Implement this as a series of PRs linked to this issue.

#### Sponsor Organization (will test alphas/betas)
NYCET

#### Developer allocated
@lperson 

#### Planned timeline
Next Spoke release (5.5 as of April 25, 2020)

#### Files/File-sections touched
- Add the new column and ship the data around
  - migrations/20200328004744_add_answer_actions_data.js
  - src/server/models/interaction-step.js
  - src/api/campaign.js
  - src/server/api/campaign.js
  - src/server/api/interaction-step.js
  - src/server/api/question.js
  - src/api/interaction-step.js
  - src/server/api/lib/import-script.js (Google doc script imports create interaction steps. I needed to update this code to create initialize the new column.)
 - src/api/schema.js

- Update interaction step editing
  - src/components/CampaignInteractionStepsForm.jsx
  - src/containers/AdminCampaignEdit.jsx
  - src/lib/interaction-step-helpers.js

- src/integrations/action-handlers/index.js (The framework.)
- src/server/api/schema.js (Updated the code around the call to processAction.)

- Bring examples up to date with the interface required by the new framework
  - src/integrations/action-handlers/mobilecommons-signup.js
  - src/integrations/action-handlers/revere-signup.js
  - src/integrations/action-handlers/test-action.js
  - src/integrations/action-handlers/actionkit-rsvp.js
  - src/integrations/action-handlers/complex-test-action.js (This one is a new sample/test handler to serve as demonstration of how to build a handler that supports client choice data.  It also serves as a fixture for writing tests that require an action handler that supports client choice data.)

#### Schema changes and migrations
Add a text column called `answer_actions_data` to `interaction_step` as described above.

#### Any other links to details about the plan (maybe a Google Doc link, or other place where folks have commented)
Not at this time.

#### Existing github issues that this RFC will address
There is a rough implementation of this in #1499. I intend to split that into several small PRs that should be reviewed and merged in sequence.  There is an rough implementation of this for NGPVAN in [this branch](https://github.com/lperson/Spoke/tree/lp_action_handler_client_choice). The VAN stuff in there is in good shape but I'm iterating on it so it lives in the proposed framework.

lperson:
# Fixes #1512

## Description

- Improve aesthetics and error handling for actionHandler selection in campaignInteractionStepsForm
- Make updateQuestionAnswers fire and forget, with better error logging

# Checklist:


lperson:
@schuyler1d Commit 7072e5d9509823119164c384b64d16bd6b1dde08 shows a test that mounts a react component and uses its methods to call a server mutation.   Calling out some specific changes:

https://github.com/lperson/Spoke/blob/09d4f162563d34a001714b05b8598020daec3396/src/containers/AssignmentTexterContact.jsx#L434 -- export `mapMutationsToProps`.

https://github.com/lperson/Spoke/blob/09d4f162563d34a001714b05b8598020daec3396/__test__/server/api/campaign/updateQuestionResponses.test.js#L21-L25 -- import `mapMutationsToProps` and import the component with a destructured import so we don't get it wrapped in `wrapMutations` and `loadData`

https://github.com/lperson/Spoke/blob/09d4f162563d34a001714b05b8598020daec3396/__test__/server/api/campaign/updateQuestionResponses.test.js#L395-L412 -- this is a utility that wraps the mutations in the react component with something that delegates to `runGql`. This is how we run the mutation against the server implementation.  This piece can be put into `test_helpers` or something like it and it should be totally reusable. Two things I'd like to do with it are 
1. validate that the functions passed to the mutation align with the schema (graphQL itself will validate the parameters in the call, but it won't validate that there are no extra parameters)
2. enhance it to handle queries

The purpose for all of this is to have tests to ensure that if a query or mutation changes, a test will fail if the react component is not also updated.

I believe this can replace `getGql` which is used in 6 places.  It's similar to `runGql` with the added benefit of extracting the query directly from mapMutationsToProps (and potentially mapQueriesToProps). Further it allows us to test calls to the graphql server that result from events in react components.  

In this example, I instantiated `AssignmentTexterContact` and called its `handleQuestionResponseChange`, which added a QuestionResponse to the object state. Then the code calls `handleSubmitSurveys`, which resulted in the graphql mutation being invoked. After the graphql mutation was executed, we verified that the QuestionResponse was added to the database.

src/server/api/mutations/updateQuestionResponses.js
```diff
@@ -0,0 +1,115 @@
...
+    .save(campaignContactId, questionResponses)
+    .catch(err => {
+      log.error(
+        `Error saving updated QuestionResponse for campaignContactID ${campaignContactId} questionResponses ${JSON.stringify(
+          questionResponses
+        )} error ${err}`
+      );
+    });
+
+  contact = await loaders.campaignContact.load(campaignContactId);
```
schuyler1d: Why is this needed vs. the one on line 12?  contact data doesn't change with changes to question responses, and if it did, you'd need to 'hard refresh' with e.g. cacheableData.campaignContact.load -- because loaders won't clear on its own.

src/server/api/mutations/updateQuestionResponses.js
```diff
@@ -0,0 +1,115 @@
...
+
+  // The rest is for ACTION_HANDLERS
+  const interactionSteps = await cacheableData.campaign.dbInteractionSteps(
```
schuyler1d: In the move from the schema.js version the high conditional here:
https://github.com/MoveOnOrg/Spoke/pull/1543/files#diff-e3e0c0fe13e813c7748359d6a931a6faL1128
to test for any action handlers is meant to short-circuit any additional db calls if there aren't any action handlers configured.  Then we don't need any of this info.

Also, you're removing a cache optimization from the schema.js version:
`        const interactionSteps =  campaign.interactionSteps ||   (await cacheableData.campaign.dbInteractionSteps(campaign.id));`
skips loading them if the campaign was from a redis cache (which includes the interactionSteps in the cache object)

src/server/api/mutations/updateQuestionResponses.js
```diff
@@ -0,0 +1,115 @@
...
+
+  if (stepsWithActions.length && !actionHandlersConfigured) {
+    log.error(
+      "Encountered one or more interaction steps with an action handler but no action handlers are configured"
+    );
+    return contact;
+  }
+
+  if (stepsWithActions.length && actionHandlersConfigured) {
+    const getAndProcessAction = async questionResponse => {
```
schuyler1d: I think I'd rather an anonymous function below, so this is inside the .map() call -- purpose and indentation is clearer then.

src/server/api/mutations/updateQuestionResponses.js
```diff
@@ -0,0 +1,115 @@
...
+        );
+      } catch (err) {
+        log.error(
+          `Error executing handler for InteractionStep ${interactionStepId} InteractionStepAction ${interactionStepAction} error ${err}`
+        );
+      }
+    };
+
+    const promises = questionResponses.map(getAndProcessAction);
+    await Promise.all(promises);
```
schuyler1d: Please no awaiting action processing before returning the request.  you can do `Promise.all(promises).catch((err) => { console.log(' action processing error', err})` (and not catch within the promise perhaps?)

src/server/api/mutations/updateQuestionResponses.js
```diff
@@ -0,0 +1,115 @@
...
+
+      const interactionStepAction =
+        questionResponseInteractionStep &&
+        questionResponseInteractionStep.answer_actions;
+
+      if (!interactionStepAction) {
+        return;
+      }
+
+      const organization = await loaders.organization.load(
```
schuyler1d: This doesn't need to be in the loop, and can be above it.

src/server/api/mutations/updateQuestionResponses.js
```diff
@@ -0,0 +1,115 @@
...
+      "Encountered one or more interaction steps with an action handler but no action handlers are configured"
+    );
+    return contact;
+  }
+
+  if (stepsWithActions.length && actionHandlersConfigured) {
+    const getAndProcessAction = async questionResponse => {
+      const { interactionStepId, value } = questionResponse;
+
+      const questionResponseInteractionStep = interactionSteps.find(
```
schuyler1d, line 54: nit: seems it would be clearer if this was done in a filter on the questionResponses to process rather than inside the promises -- none of this is async, so it seems better to have fewer promises in the first place.

src/server/api/schema.js
```diff
@@ -61,6 +60,8 @@ import { change } from "../local-auth-helpers";
 import { symmetricEncrypt } from "./lib/crypto";
 import Twilio from "twilio";
 
+const UpdateQuestionResponses = require("./mutations/updateQuestionResponses");
```
schuyler1d: please move this into the more standard import inside the "./mutations" two lines down, and just importing the function rather than the file as a whole.

src/server/api/schema.js
```diff
@@ -1106,74 +1107,12 @@ const rootMutations = {
...
-      const campaign = await loaders.campaign.load(contact.campaign_id);
-      await assignmentRequiredOrAdminRole(
-        user,
-        campaign.organization_id,
-        contact.assignment_id,
-        contact
-      );
-
-      await cacheableData.questionResponse.save(
+      return UpdateQuestionResponses.updateQuestionResponses(
```
schuyler1d: I think my preference is that rather than change the function signature across the mutation file, line 1105 here should just be the imported function `updateQusetionResponses` without any ": async...." at all -- i.e. just the matched value exactly.

And then in the mutation file `updateQusetionResponses` should take the exact signature above.

schuyler1d (changes requested):

src/server/api/mutations/updateQuestionResponses.js
lperson: I intended this to ensure we're returning the updated contact.   **is that desirable?**

If so, yes, we'll need to add the hard refresh.  I do think this is desirable because if contact is cached we'd want it refreshed so these respohses are retrieved next time someone retrieves the contact.

src/server/api/mutations/updateQuestionResponses.js
lperson: I'll revise.

src/server/api/mutations/updateQuestionResponses.js
lperson: Sure, although I think this is more readable.  I'll make the change.

src/server/api/schema.js
lperson: This makes it testable.  We can discuss.

lperson:
@schuyler1d, you asked:

> Why is this needed vs. the one on line 12?  contact data doesn't change with changes to question responses, and if it did, you'd need to 'hard refresh' with e.g. cacheableData.campaignContact.load -- because loaders won't clear on its own. 

I said that I was reloading the contact to ensure we return a contact with updated QuestionResponses. 

Instead of returning possibly stale information, I'm changing the mutation to return only contactId. No code uses the return value anyway, and I think it's better to return information that is not possibly stale.

schuyler1d (approved):
:+1:

PR #1547, opened 2020-05-09
github: Add 10m timeouts to each jest job

### LINKED ISSUE #1546: Bug: coverage runs taking 6h
**Describe the bug**
Recent coverage jobs are stalling and taking 6h: https://github.com/MoveOnOrg/Spoke/actions

### LINKED ISSUE #1564: Feature Request: Allow reordering of survey questions
Problem
Once survey questions are created, they can't be reordered and so they need to be deleted and then recreated.

Solution
Allow for drag and drop reordering of survey questions once they're created.

Context
There is currently a bug (#1144) where the survey questions change order on the texter side. This would fix that by creating a set order (that can also be reordered)

tmc:
Fixes #1546 

## Description
This adds a 10m timeout to get much faster feedback and prevents wasting resources.

This does not address why the coverage job is failing to end cleanly.

### DIFF
--- a/.github/workflows/jest-tests.yaml
+++ b/.github/workflows/jest-tests.yaml
@@ -5,6 +5,7 @@ on: push
 jobs:
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     strategy:
       matrix:
         node-version: [10.x, 11.x, 12.x]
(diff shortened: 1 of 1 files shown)


schuyler1d (approved):
:+1:

PR #1550, opened 2020-05-09
Update REFERENCE-environment_variables.md

Frydafly:
added note that the API key request in VAN is called "Hustle", not "Spoke", which is counter-intuitive and worth mentioning


schuyler1d:
Defer to @lperson in this, but I preliminarily approve.  Thanks! :+1:

lperson (approved):
🎉

ibrand (approved):
Thanks!

PR #1555, opened 2020-05-17
Fix Buggy Environment variables reference claims

### LINKED ISSUE #1476: Bug: Environment variables reference claims Spoke supports MySQL when PG is hardcoded for any DB_TYPE value
Hi Spoke Team,

I found that the [Environment Variables Reference Guide](https://github.com/MoveOnOrg/Spoke/blob/main/docs/REFERENCE-environment_variables.md) claims that Spoke supports MySQL

However, if any value is set in the `DB_TYPE` environment variable, `pg` is autoselected according to the code here:

https://github.com/MoveOnOrg/Spoke/blob/46e54ed1d33a35ee5263b4a153b7dad599ca5ee4/src/server/knex-connect.js#L33-L35

I understand that Postgres is the preferred database and think the doc and the environment variable should be updated to reflect this.

alexanderankin:
# Fixes #1476 

## Description

The issue describes this well

# Checklist:


ibrand:
@schuyler1d is this reasonable? Can Spoke actually support dbs that aren't pg or should we update docs to reflect that this should always be pg

alexanderankin:
I somehow think this will break tests and stuff which depend on say `postgres` value to -> `pg`, so it will need modifications elsewhere in the codebase, but it seemed like a necessary first step

schuyler1d (approved):
Yeah.  We don't support other DBs in production, but this shouldn't block it.

PR #1560, opened 2020-05-21
Update docker-compose for local dev and add some helpful dev tools

matteosb:
The simplified docker-compose file just runs redis and postgres and should make it easier to run them for local development while continuing to run the development server with `yarn run dev`. 

I also added dev tools to wrap the redis and postgres CLIs so folks don't need to install them locally and a script to set up the test database.

To use this set up:
1. Install docker and docker compose
2. `docker-compose up -d` to run redis and postgres in the background 
3. `./dev-tools/create-test-database`

`psql` and `redis-cli` can be installed locally or run using dev-tools, for example:

```bash
$ ./dev-tools/psql -c "SELECT COUNT(*) FROM organization"
 count 
-------
     3
(1 row)

$ ./dev-tools/psql
psql (12.2)
Type "help" for help.

spokedev=# 

$ ./dev-tools/redis-cli --scan 0
```


ibrand (approved):
This is awesome!

ibrand:
Is there any dev documentation on devtools/ btw?

docker-compose.yml
```diff
@@ -9,28 +9,15 @@ services:
...
     volumes:
       - postgres:/var/lib/postgresql/data
+    ports:
+      - 5432:5432
   redis:
     image: redis:alpine
     restart: always
     volumes:
       - redis:/data
-  spoke:
```
schuyler1d, line 17: Will this work with the heroku deploy? Maybe Heroku just uses the `Dockerfile`?

Also, I am super-naive about docker, but doesn't this delete the spoke app itself here?  How does spoke get run within Docker?  Or the idea is that Dockerfile is run 'automatically' and the services here are supplemental?

I would also ++ @ibrand 's request for maybe a "HOWTO-run-docker.md" file -- then we could link to it -- maybe even as the default instructions?

ibrand: @schuyler1d this will not mess with the Heroku deploy cause [it does use the Dockerfile](https://github.com/MoveOnOrg/Spoke/blob/main/heroku.yml#L3)

I'm also not so Docker knowledgeable, so not sure about your question there

matteosb: Yeah, it doesn't impact heroku at all. Heroku uses the Dockerfile to build an image for Spoke, whereas the docker-compose file is just used to run the standard postgres and redis images for local development.

matteosb:
I'll add some documentation to this PR in the next few days.

PR #1581, opened 2020-06-01
Upgrade to apollo-client and react-apollo 2.x

matteosb:
# Fixes # (issue)

This fixes a number of bugs caused cache collisions in the Apollo store (see #1461) and fixes polling on the AdminCampaignEdit screen.

## Description

Upgrades apollo-client and react-apollo to recent versions and removes redux along the way. This has numerous benefits beyond fixing the apollo store, including a better, less buggy api and much better documentation.

I borrowed heavily from the upgrade in politics-rewired's [fork](https://github.com/politics-rewired/Spoke/pull/517), which does a great job of keeping the api of the load-data HOC close to what is currently in Spoke.

Other notes:
* I changed how polling works in AdminCampaignEdit and rewrote the AdminScriptImport component to poll from the parent. If we want to avoid querying for all the campaign data when polling, we can add the pendingJobs query back in, but it complicated the code a bit and requires a refetch, which re-renders the screen.
* fixed a bad permissions check in AdminScriptImport
* apollo-client 2 removes the dependency on redux, so I removed it entirely because I didn't see the benefit

# Checklist:


schuyler1d (approved):
:+1:

PR #1599, opened 2020-06-09
Stage-main 7.0 prep

schuyler1d:
* texter sideboxes (behind TEXTER_SIDEBOXES) (schuyler1d)
* Tag improvements (behind EXPERIMENTAL_TAGS) (schuyler1d)
  * texter sidebox for tagging
  * Tag filtering in Message Review
* VAN action handler (lperson)
* Conversation links (as a sidebox and in Message Review) (lperson)
* https://github.com/MoveOnOrg/Spoke/pull/1602 Make dataloaders per-request (matteosb)
  * New dev rule: no `loaders` in mutations -- only use cacheableData
* Some accessibility tweaks for the Texter Experience; Different keyboard shortcuts to send:
  * Enter key is removed to send for accessibility reasons (needs to be used for keyboard 'click'ing)
  * Ctrl-x to Send Ctrl-y to Skip, When you are sending initial texts, pressing any key will send.
* https://github.com/MoveOnOrg/Spoke/pull/1604 remove buggy notification on replies (matteosb)

## Awaiting completion
* https://github.com/MoveOnOrg/Spoke/pull/1562 - Suspending texters

## Draft Release Notes
* Spoke 7.0 has several valuable infrastructure updates
  * Bug fixes:
    * Twilio incoming message handlers would sometimes fail due to a defunct database (knex) listener
    * Local memory cache would sometimes be used after updates, especially in persisting multi-server environments (tech: 'loaders' are now appropriately per-request and only used in queries, not mutations)
   * Improvements
     * There is now further support for VAN -- a robust action handler that can be enabled by adding `ACTION_HANDLERS=ngpvan-action` to your environment variables will enable this.
     * We have a new framework for extending the Texter UI interface called ["sideboxes"](https://github.com/MoveOnOrg/Spoke/issues/1533).  Enabled with environment variable TEXTER_SIDEBOXES which works similarly to action handlers, where it should be a comma separated list of enabled sideboxes. Developers can add sidebox functionality, and admins can set defaults in the organization Settings section and changes per-campaign -- enabling/disabling sideboxes.  Two so-far are:
         * `contact-reference` which is a link the texter can click and/or share with an admin for direct access to that conversation later
         * `tag-contact` which if you have EXPERIMENTAL_TAGS=1 enabled and create tags, a texter will have an interface to mark them.  They can then be filtered for in Message Review.

       There will be more work here going forward -- feedback is welcome as this is still a feature in active development.
[...]

PR #1602, opened 2020-06-09
Return to request-scoped data loaders

matteosb:
Data loaders cache objects in memory indefinitely, which could cause
any number of issues due to stale reads. The code is full of partial
work-arounds, but none of them can guarantee data freshness without
returning to request-scoped loaders. This change does that
and removes some of of the work-arounds since they are no longer necessary.
I also cleaned up some unnecessary usage of loaders in mutations but left most
intact, even though they don't really offer any advantage over calling
cacheableData.

Note: this PR will probably noticeably increase Redis / PG usage
for some deployments, though probably not that much on Lambda.

# Checklist:

### DIFF
--- a/__test__/integrations/action-handlers/index.test.js
+++ b/__test__/integrations/action-handlers/index.test.js
@@ -1,4 +1,4 @@
-import { r, createLoaders } from "../../../src/server/models";
+import { r } from "../../../src/server/models";
 import each from "jest-each";
 import {
   setupTest,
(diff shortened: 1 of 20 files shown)


schuyler1d (approved):
:+1:

schuyler1d:
Also removed the rest in api/mutations/ and a test https://github.com/MoveOnOrg/Spoke/pull/1599/commits/ae01a01310b1b8098a3fb4132cd702f073d245d1

PR #1610, opened 2020-06-11
Stage main with wfp hot fixes

lperson:
# Fixes # (issue)

## Description

# Checklist:


schuyler1d (changes requested):
Rather than change the filter (and add a new one), I'd rather just not link to the conversation when it's in `needsMessage` or `messaged` status -- the conversation has no usefulness until there's a contact response.

Maybe that can be a separate PR and we can revert this in WFP, if that's ok.

If WFP use-cases require linking into those conversations, then I'd like to understand more what that use-case is.

PR #1611, opened 2020-06-11
Disable next/prev when in review mode

lperson:
When we've used a conversation permalink to see a conversation in the texter view, we should disable the next and previous buttons so we don't go stepping through conversations. The purpose of arriving in the texter view is to look at a single conversation, not to go from conversation to conversation. That is best done from the texter home screen.

# Checklist:


schuyler1d:
As we discussed, I think it's better to keep the navigation options, as a feature to step through other conversations from the same texter if the admin wants to (not to mention I also think of this link as being useful to texters)

PR #1616, opened 2020-06-11
User menu: route users to admin or texter based on role

matteosb:
Fixes issue where texters get a blank screen or an error page (as of 6.1) when they switch organizations because we attempt to route to /admin.

# Checklist:


schuyler1d (approved):
I think what we should do long-term is just have a `role` result for currentUser { organizations } and then calculate what that means per place to test, but this is fine as a client solution

PR #1619, opened 2020-06-13
Minor changes to fix script import ui

lperson:
## Description

WFP reported that the script import button is forever disabled for a campaign after a failed import attempt. I noticed that we didn't give feedback in the UI after a failed attempt, which important for folks to diagnose why an import failed.

Changes affect only script import.

 * Display error from last import attempt, if any
 * Fix bug that caused import button to forever be disabled for a
   campaign after a failed import
 * Delete old jobs of type `import_script` for the campaign before
   starting a new script import

# Checklist:

### DIFF
--- a/src/containers/AdminScriptImport.jsx
+++ b/src/containers/AdminScriptImport.jsx
@@ -20,12 +20,18 @@ const styles = StyleSheet.create({
 export default class AdminScriptImport extends Component {
   static propTypes = {
     startImport: PropTypes.func,
-    hasPendingJob: PropTypes.bool
+    hasPendingJob: PropTypes.bool,
+    jobError: PropTypes.bool,
+    onSubmit: PropTypes.bool
   };
 
   constructor(props) {
     super(props);
-    this.state = {};
+    this.state = {
+      ...(!!props.jobError && {
+        error: `Error from last attempt: ${props.jobError}`
+      })
+    };
   }
 
   startImport = async () => {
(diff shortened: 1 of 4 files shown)


src/workers/jobs.js
```diff
@@ -826,14 +848,15 @@ export async function exportCampaign(job) {
 export async function importScript(job) {
   const payload = await unzipPayload(job);
   try {
+    await defensivelyDeleteOldJobsForCampaignJobType(job);
     await importScriptFromDocument(payload.campaignId, payload.url); // TODO try/catch
-    console.log(`Script import complete ${payload.campaignId} ${payload.url}`);
+    log.info(`Script import complete ${payload.campaignId} ${payload.url}`);
```
matteosb: for now please prefer console.log. `minilog` output gets messed up in Lambda

src/containers/AdminScriptImport.jsx
```diff
@@ -20,19 +20,26 @@ const styles = StyleSheet.create({
...
+      })
+    };
   }
 
   startImport = async () => {
     const res = await this.props.startImport(this.state.url);
     if (res.errors) {
       this.setState({ error: res.errors.message });
     }
+    this.props.onSubmit();
```
matteosb, line 42: is this on submit necessary?

lperson: I think so. It causes the section to close.

matteosb: Ok, this component is kind of weird. It seems like startImport is the "real" onSubmit and onSubmit here is just being called to advance sections. If you call onSubmit you can get rid of the redundant refetchQueries on the import script mutation.

matteosb (approved):

joemcl:
Thanks @lperson tested on spoke-wf-mo and working. Just to note that the behavior was not confined to failed imports, but also occurred after a first successful import for subsequent successful imports.

PR #1620, opened 2020-06-13
fix unwanted pop ups when loading script editor ON SAFARI

lperson:
This is described by users as "I need to step through every interaction step before I can do anything."

This was fixed in the WFP fork and never merged. It resurfaced when WFP started using Spoke 6 recently.

This happens only on safari on desktop and iOS.

Safari is causing onFocus to fire for the script in every interaction step after the first render. It's not clear to me why that was happening.

We were handling onFocus to cause the dialog to open and present the script editor, rather than handling it on the click. That seems unnecessary, because is anybody really going to tab from script to script and expect the dialog to open? Once it opens, they can't tab to the next step anyway.

## Description

# Checklist:

### DIFF
--- a/src/components/forms/GSScriptField.jsx
+++ b/src/components/forms/GSScriptField.jsx
@@ -89,9 +89,6 @@ export default class GSScriptField extends GSFormField {
       <div>
         <TextField
           multiLine
-          onFocus={event => {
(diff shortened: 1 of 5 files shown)


schuyler1d (approved):
Thanks!

PR #1624, opened 2020-06-13
If a contact has no timezone, format local time using campaign's timezone.

lperson:
The texter UI is meant to show the local time for the contact. Unfortunately, it has been displaying UTC if the contact has no timezone on their record in campaign_contact.

This PR will cause Spoke to use the campaign's timezone to determine the contact's local time to display on the texter UI.  The campaign's timezone defaults to `US/Eastern` because it had to default to something.

It also allows the campaign's timezone to be set regardless of whether "override organization texting hours` and `texting hours enforced` are set.

# Checklist:


schuyler1d (approved):
:+1:

schuyler1d:
It would be good to have an env var (also checked by the org) for a default timezone for the campaign

PR #1627, opened 2020-06-14
Use a default timezone moment understands when running on Lambda

matteosb:
This should fix the following error in the logs when running on lambda:
`ERROR	Moment Timezone has no data for :UTC. See http://momentjs.com/timezone/docs/#/data-loading/.`

Lambda passes the value `:UTC` for the `TZ` env var. Because this is a reserved env var, we should probably switch to using a different one altogether.


hiemanshu (approved):

schuyler1d (approved):
Argh!
Thanks

PR #1634, opened 2020-06-18
Fix data warehouse contact loader

jeffm2001:
## Description

The data warehouse contact loader is not currently working. This PR fixes the most glaring issues — although in a relatively basic way. I suspect that the contact loader code could be refactored and greatly simplified, but not sure I'm the person to do it. (While I've mostly been able to figure out *what* the code is doing, I lack context for the *why*).

The first issue is on the frontend — the onChange event is not firing, so the SQL is not validated and you can't proceed. I've fixed that by adding a "Validate SQL" button to manually trigger the validation.

The second issue is a ReferenceError for undefined job and jobMesseges variables when `completeContactLoad` is called. Fixed this by passing job into `loadContactsFromDataWarehouseFragment`.

Note: I don't have an AWS instance of Spoke running, so I'm only able to verify this is working locally and on Heroku.

# Checklist:


schuyler1d (dismissed):
Thank you!

schuyler1d (dismissed):
:+1:

schuyler1d:
I missed a commit in 7.1, but merging into main so this is resolved.

PR #1637, opened 2020-06-19
Correct A Line About How To Test

JeremyParker:
## Description

This change removes some outdated info and makes the docs more D.R.Y. by pointing to the authoritative source on how to test, rather than duplicating the information in summarized form. This should make the docs easier to maintain going forward.


schuyler1d:
changed base in preparation for 7.0 merge/release

PR #1642, opened 2020-06-23
7.0 release notes

schuyler1d:
@ibrand please review


ibrand (approved):
One nit: can we change the `###` headers to be just bold with `**` ? None of our other sections use extra headers

PR #1646, opened 2020-06-25
Add tags to campaign exports

schuyler1d:
# Fixes Tags missing from export data

## Description

Includes tags in campaign data exports

passed on by: @hiemanshu (and sponsored by Scale-to-Win) -- thanks!


ibrand (approved):
🥳

PR #1647, opened 2020-06-26
Extra Profile Fields

jeffm2001:
## Description

This allows the configuration of additional profile fields that texters are required to fill out. Fields are set per-organization, but since there is no organization context at user signup, the extra profile fields must be completed after logging in. If they're not, texters are redirected from the ToDo dashboard to the account edit page to complete them. Responses are stored in a new `extra` column on the users table.

2020-06-26 at 11 11 AM

We plan to use this feature for better texter attribution.

To test this out, you can add this to the organization.features column in the Spoke db:
```
{
 "profile_fields" : [
    {
      "label" : "Affiliate",
      "name" : "affiliate"
    },
    {
      "label" : "Local Number",
      "name" : "local"
    },
    {
      "label" : "Foo",
      "name" : "foo"
    }
  ]
}
```

There are a few obvious next steps, which I would suggest could be follow-up PRs:

- An interface in the organization settings page to configure the profile fields (I think a [FieldArray](http://jquense.github.io/react-formal/api/FieldArray) would be good for that, but we would need to upgrade react-formal
- Showing the profile values in the UI and/or exports
- Allowing other field types and/or optional fields

Possibly, the settings interface should be included in this PR, but I ran out of time for working on it. If there is interest, I would welcome support on the frontend part of that.

# Checklist:

### DIFF
--- a/__test__/backend.test.js
+++ b/__test__/backend.test.js
@@ -743,6 +743,7 @@ describe("editUser mutation", () => {
       data: {
         editUser: {
           ...userData,
+          profileComplete: true,
           id: testTexter.id.toString()
         }
       }
--- a/__test__/backend.test.js
+++ b/__test__/backend.test.js
@@ -731,7 +731,8 @@ describe("editUser mutation", () => {
       lastName: "Garcia",
       alias: "JG",
       email: "jerry@heaven.org",
-      cell: "4151111111"
+      cell: "4151111111",
+      extra: null
     };
 
     variables.userData = userData;
--- a/__test__/containers/UserEdit.test.js
+++ b/__test__/containers/UserEdit.test.js
@@ -1,0 +1,40 @@
+/**
+ * @jest-environment jsdom
+ */
+import React from "react";
+import { mount } from "enzyme";
+import MuiThemeProvider from "material-ui/styles/MuiThemeProvider";
+import { UserEdit } from "../../src/containers/UserEdit";
+import { StyleSheetTestUtils } from "aphrodite";
+
+describe("User edit page", () => {
+  StyleSheetTestUtils.suppressStyleInjection();
+  const wrapper = mount(
+    <MuiThemeProvider>
+      <UserEdit />
+    </MuiThemeProvider>
+  );
+
+  test("Render edit page with standard fields", () => {
+    expect(wrapper.exists({name: "email"})).toBeTruthy();
+    expect(wrapper.exists({name: "firstName"})).toBeTruthy();
+    expect(wrapper.exists({name: "extra.foo"})).toBeFalsy();
+  });
+
+  test("Render edit page with extra profile fields", () => {
+    const org = {
+      organization: {
+        profileFields: [{
+          name: 'foo',
+          label: 'Foo'
+        }]
+      }
+    };
+    wrapper.find(UserEdit).setState({currentOrg: org});
+
+    expect(wrapper.exists({name: "email"})).toBeTruthy();
+    expect(wrapper.exists({name: "firstName"})).toBeTruthy();
+    expect(wrapper.exists({name: "extra.foo"})).toBeTruthy();
+  });
+
+});
--- a/migrations/20200623101608_profile_fields.js
+++ b/migrations/20200623101608_profile_fields.js
@@ -1,0 +1,11 @@
+exports.up = knex => {
+  return knex.schema.table("user", table => {
+    table.jsonb("extra").defaultTo(null);
+  });
+};
+
+exports.down = knex => {
+  return knex.schema.table("user", table => {
+    table.dropColumn("extra");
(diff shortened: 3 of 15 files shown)


schuyler1d (dismissed):
:+1: 
would be nice if you could add a test for an extra field being present, as well.

schuyler1d:
a small bugfix was pushed to stage-main to fix some edge cases:
https://github.com/MoveOnOrg/Spoke/pull/1652/commits/ce02872b8d00be38c3e0a23828f67944c483ace5

schuyler1d:
and https://github.com/MoveOnOrg/Spoke/pull/1652/commits/3d664fa79dc3d8a0f6d4b26d85144d13e1bb789f

schuyler1d:
Some QA feedback: user was missing the need to fill out the (new) required fields.  Maybe, when relevant, we can have a message at the top "Please fill in the missing fields".  Thoughts?

Also, apparently changing the password was (also) blocked without the field being saved -- I can imagine it being ok to block password changing if that's too annoying, but in that case, I guess we should at least hide the "change password" link until they save a complete profile.

jeffm2001:
That's good feedback. My original idea was to have a box on the ToDo page with a link to UserEdit if the profile wasn't complete, but that proved complicated so I changed to the redirect. But agree that with no instructions/prompt it's confusing. A message at the top seems doable.

Not sure about the password change. I'll take a look.

jeffm2001:
After noticing how `UserEditDIalog` is used in `PeopleList`, I have some thoughts about making it a dialog that automatically pops up from the texter dashboard if you have profile fields that need to be filled out. But that looks like more work than I have time for at the moment, so just added a message to the UserEdit page for now.

I did not experience the password change being blocked, but I think it's better to hide it anyway, so I went ahead and did that too.

schuyler1d (approved):
:+1:

joemcl:
This seems like a way to add texters to 'teams' as well and to then have leaderboards for texter teams, not just individual texters. We can use whatever field names we want to substitute for Affiliate, Local Number, 'Foo' etc right? cc: @deasterdaywfp @lperson . 

{
 "profile_fields" : [
    {
      "label" : "Affiliate",
      "name" : "affiliate"
    },
    {
      "label" : "Local Number",
      "name" : "local"
    },
    {
      "label" : "Foo",
      "name" : "foo"
    }
  ]
}

jeffm2001:
@joemcl Yep, you can use whatever field names you want. And I don't think it would be difficult in a future PR to make it so you can use different input types, like a dropdown of team names.

joemcl:
Thanks @jeffm2001 and so the only place that the fields would show up in the texter UI is on new account creation? 
And not at all yet in the admin UI, like in People? 
So finding, editing, querying etc all has to be done in the backend Postgres database, or via dashboards connected to that, etc.

jeffm2001:
> Thanks @jeffm2001 and so the only place that the fields would show up in the texter UI is on new account creation?
> And not at all yet in the admin UI, like in People?
> So finding, editing, querying etc all has to be done in the backend Postgres database, or via dashboards connected to that, etc.

The fields actually don't appear on account creation. That's because the extra fields can be configured per-organization, and at account creation there is no organization context (and the different auth options complicate things further). Instead, it shows the fields on the UserEdit page after login. And if there are missing fields you get redirected from the texter dashboard to UserEdit, so that you have to complete them before you start texting.

And correct that, besides the User Edit page, the fields don't appear anywhere in the UI or exports. Just in the database for now, so some good followup work there too.

joemcl:
> > Thanks @jeffm2001 and so the only place that the fields would show up in the texter UI is on new account creation?
> > And not at all yet in the admin UI, like in People?
> > So finding, editing, querying etc all has to be done in the backend Postgres database, or via dashboards connected to that, etc.
> 
> The fields actually don't appear on account creation. That's because the extra fields can be configured per-organization, and at account creation there is no organization context (and the different auth options complicate things further). Instead, it shows the fields on the UserEdit page after login. And if there are missing fields you get redirected from the texter dashboard to UserEdit, so that you have to complete them before you start texting.

OK so not strictly on account creation but before the user can text. Got it. 

And this will have no effect on existing users, just applies to new users on creation right? 

For *existing* users, adding the info is optional, if not provided, won't effect sending, correct? But existing users will also see the fields in their account edit page and can add the info  if they want to. 
 
> And correct that, besides the User Edit page, the fields don't appear anywhere in the UI or exports. Just in the database for now, so some good followup work there too.

jeffm2001:
It does effect existing users. When a new extra field is configured, all users without a value for that field will be required to fill it out before texting.

joemcl:
@jeffm2001 OK got it, thanks. So we just have to think through what we want folks to add as a 'team name' before implementing it for our instance.

PR #1648, opened 2020-06-27
refactor message handlers a little

lperson:
## Description

This PR is a minor refactoring of the message handling code. I'm creating this PR so it's clear that I did here, although the NGPVAN message handler will be based on this branch, and these changes will be included therein.

* made `available` async for the ngpvan case (future PR)
* moved filtering of message handlers to `message-handlers/index` for
  separation of conerns
* fixed profanity-tagger test

Going forward we may want to add caching around the `available` method of the message handlers the way we do for action handlers and contact loaders.

# Checklist:

### DIFF
--- a/__test__/integrations/message-handlers/profanity-tagger.test.js
+++ b/__test__/integrations/message-handlers/profanity-tagger.test.js
@@ -39,11 +39,13 @@ describe("Message Hanlder: profanity-tagger", () => {
     const c = await createStartedCampaign();
     await r.knex("tag").insert([
       {
+        id: 1,
         name: "Contact Profanity",
         description: "mean contact",
         organization_id: c.organizationId
       },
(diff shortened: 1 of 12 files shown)


schuyler1d (changes requested):
I specifically made `available` non-async -- more message-passing, availability is during a current action (rather than whether to show admin a choice -- this is IN the message loop cycle) -- if you want to bail on availability because of downtime, etc, then put that logic in the message handler preSave or postSave call itself.

If you want to move the logic into an index, that's fine (and I appreciate the tests!)

schuyler1d:
this also conflicts with upcoming async-dispatch work, so I'm going to close this.  Stop gaps can be locally setting things to `await` but let's ....await these upcoming changes.

PR #1649, opened 2020-06-29
Slack auth

matteosb:
## Description
Adds slack authentication, see the docs for more details.

# Checklist:

### DIFF
--- a/src/client/auth-service.js
+++ b/src/client/auth-service.js
@@ -3,27 +3,63 @@ import auth0 from "auth0-js";
 const baseURL =
   window.BASE_URL || `${window.location.protocol}//${window.location.host}`;
 
-export function logout() {
-  const webAuth = new auth0.WebAuth({
-    domain: window.AUTH0_DOMAIN,
-    clientID: window.AUTH0_CLIENT_ID
-  });
+const loginStrategies = {
+  auth0: {
+    login(nextUrl) {
+      const webAuth = new auth0.WebAuth({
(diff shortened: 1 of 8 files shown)


src/client/auth-service.js
```diff
@@ -3,27 +3,63 @@ import auth0 from "auth0-js";
 const baseURL =
   window.BASE_URL || `${window.location.protocol}//${window.location.host}`;
 
+const loginStrategies = {
```
schuyler1d, line 6: A good first step -- I think the next one, we'd want to move all this into src/integrations/auth/ files/directories and this could be in the react-component.js file.

schuyler1d (approved):
:+1: 
This is good -- I think we're really close to being able to move this into src/integrations/auth/

schuyler1d:
Added https://github.com/MoveOnOrg/Spoke/commit/f3b271a9e2adb6ccddf782258bf4de02f588b432
to stage-main to maintain backwards compatibility where PASSPORT_STRATEGY may be empty

joemcl:
Going through the howto doc to add permissions on a non-paid Slack and I see this - 
```
Your workspace is not on a paid tier and therefore cannot switch authentication modes. Upgrade today
```

joemcl:
OK never mind, maybe, I'm now actually looking at this - https://api.slack.com/apps . I was just looking at the settings in our existing Slack workspace, I have not created an app yet.

PR #1650, opened 2020-06-29
ctrl-Enter to send from Ctrl-x conflicting as a Cut command

### LINKED ISSUE #1643: Bug: Ctrl-X to send overrides 'Cut'
**Describe the bug**
The new keyboard shortcut 'Ctrl-X' sends a message. Typical behavior is the 'cut' command, and this overrides that.

**To Reproduce**
1. Go to send a reply.
2. Select a response.
3. Highlight the part you want to combine with another response.
4. Ctrl-X to cut that part.
5. The message inadvertently sends before desired.

**Expected behavior**
Ctrl-X should 'cut' the selected text. This deletes it and adds it to the clipboard.

**Screenshots**
N/A

 - OS: Windows 10, version 2004
 - Browser: Chrome
 - Desktop
 - Version: 83.0.4103.106

**Additional context**
Can a different shortcut be made for send? (Yes, I know that copying with Ctrl-C is an option, but I don't see how it's a good idea to override an default OS-level shortcut, to cause accidental sends.)

schuyler1d:
# Fixes #1643

## Description

Switches to Ctrl-Enter for sending texts

### DIFF
--- a/src/components/AssignmentTexter/Controls.jsx
+++ b/src/components/AssignmentTexter/Controls.jsx
@@ -157,8 +157,8 @@ export class AssignmentTexterContactControls extends React.Component {
     // if document.activeElement then ignore a naked keypress to be safe
     // console.log('KEYBOARD', evt.key, document.activeElement);
     if (
-      // SEND: Ctrl-x
-      evt.key === "x" &&
+      // SEND: Ctrl-Enter
+      evt.key === "Enter" &&
       // need to use ctrlKey in non-first texting context for accessibility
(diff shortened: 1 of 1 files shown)


ibrand (approved):
🎉

ibrand:
quick q, should this be modifier + x so that it's `cmd` on mac?

joemcl:
Tested on WFP staging, works as described - Ctrl + X does cut of text, does not send the message. Ctrl + Enter does send the message.

PR #1654, opened 2020-06-30
Configurable background processing backends with bull and lambda implementations

matteosb:
Motivation: the optimal strategy for executing background work is specific to how Spoke is deployed and how much it needs to scale. By making job running an integration point, it will be easier for organizations to customize their Spoke deployment while keeping up to date with MoveOn/main.

`job-runners` expose a simple api for executing "Jobs", i.e. heavy-weight jobs tracked in the DB, and "Tasks", light-weight fire-and-forget functions. I tried to add tasks for all the places where we intentionally don't await promises, since that does not work well on Lambda.

In addition to migrating the existing job execution strategies in a backwards-compatible way, I added job runners implementations using Lambda async invocation and `bull`, a Redis-backed job queuing library. `lambda-async` has been used in production while `bull` is really more of an example at this point.

Possible follow ups to this PR:
1. Make sending notifications a task
2. Delegate job tracking to the job runner since some backends can do it without the JobRequest table
3. Expose an api for optional features like retries, locking, and rate limiting per job/task type
4. Improve the `bull` job runner and make it the default for non-lambda deployments
5. Add an SQS-backed job runner for Lambda with support for retries and rate limiting
6. Split tag handling out of action handlers into its own integration

### DIFF
--- a/src/integrations/job-runners/README.md
+++ b/src/integrations/job-runners/README.md
@@ -14,12 +14,16 @@ introducing conflicts with MoveOn/main.
 "Tasks" are lightweight fire-and-forget operations, while "Jobs" are
 long-running operations that are tracked in the JobRequest table.
 
+Task payloads should always be JSON-serializable and should be relatively small because
+they are passed to the job runner directly without being written to the database.
+If the maximum payload for a background operation exceeds 100Kb,
+consider using a Job rather than a Task.
+
 While most queueing systems should be able to handle both types of background
 processing, we are keeping them separate for now to allow for different execution
 strategies. For example, it might be reasonable to always run tasks in-process
 outside of AWS Lambda while still pushing Jobs to a separate worker process.
 
-Note: This api may change in the future.
 
 ## Configuration
 
--- a/src/integrations/job-runners/README.md
+++ b/src/integrations/job-runners/README.md
@@ -1,0 +1,52 @@
+# Job Runners
+
+## Intro
+
+Spoke has several functions that need to be run in the background after
+an API request has returned. Strategies for executing this background
+work depend on the deployment environment. This job runner API allows
+organizations to customize background jobs to their deployment without
+introducing conflicts with MoveOn/main.
+
+
+## Tasks vs Jobs
+
+"Tasks" are lightweight fire-and-forget operations, while "Jobs" are
+long-running operations that are tracked in the JobRequest table.
+
+While most queueing systems should be able to handle both types of background
+processing, we are keeping them separate for now to allow for different execution
+strategies. For example, it might be reasonable to always run tasks in-process
+outside of AWS Lambda while still pushing Jobs to a separate worker process.
+
+Note: This api may change in the future.
+
+## Configuration
+
+A Spoke instance must have exactly one job runner configured via the `JOB_RUNNER`
+env var. For backwards compatibility, JOB_RUNNER will default to 'legacy', which
+uses the `JOBS_SAME_PROCESS` and `JOBS_SYNC` env vars.
+
+Individual job runners may require additional configuration, for example
+an SQS queue URL or Lambda function ARN.
+
+## Implementing a Job Runner
+
+Job Runner modules must export the following functions:
+  * `fullyConfigured() -> boolean`
+  * `dispatchJob(jobData) -> JobRequest`
+  * `dispatchTask(taskName, payload) -> void`
+
+Use fullyConfigured to fail-fast if your job running strategy isn't properly configured.
+`dispatchTask` and `disptachJob` both enqueue units of work, but `dispatchJob` has to also
+save and return a JobRequest object.
+
+## Adding new Jobs and Tasks
+
+Tasks are defined in `src/workers/tasks.js` and Jobs are defined
+in `src/workers/job-processes.js`. To add a new task/job add its
+ name to the Tasks/Jobs enum, and the function to run in the
+ taskMap/jobsMap object.
+
+_Note: when writing new jobs/tasks keep in mind that they
+may execute in a separate process and design accordingly._
--- a/src/integrations/job-runners/README.md
+++ b/src/integrations/job-runners/README.md
@@ -50,3 +50,21 @@ in `src/workers/job-processes.js`. To add a new task/job add its
 
 _Note: when writing new jobs/tasks keep in mind that they
 may execute in a separate process and design accordingly._
+
+## Notes
+
+#### Retries
+
+As currently implemented, neither the 'legacy' nor the 'lambda-async' job runners
+support retries, but this could be a good feature to add to the API. Individual jobs
+would have to be tagged as retry-able/idempotent.
+
+#### Recommendations for running on AWS Lambda
+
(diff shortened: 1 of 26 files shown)


src/integrations/job-runners/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+"Tasks" are lightweight fire-and-forget operations, while "Jobs" are
+long-running operations that are tracked in the JobRequest table.
+
+While most queueing systems should be able to handle both types of background
+processing, we are keeping them separate for now to allow for different execution
+strategies. For example, it might be reasonable to always run tasks in-process
+outside of AWS Lambda while still pushing Jobs to a separate worker process.
+
+Note: This api may change in the future.
```
schuyler1d: Maybe worth adding something about a task's maximum payload should be. Lambda's is, I believe, 256K so maybe 255K for wrapping space.

src/server/api/mutations/sendMessage.js
```diff
@@ -139,17 +140,14 @@ export const sendMessage = async (
...
-  //   `Sending (${serviceName}): ${messageInstance.user_number} -> ${messageInstance.contact_number}\nMessage: ${messageInstance.text}`
-  // );
-  //NO AWAIT: pro=return before api completes, con=context needs to stay alive
-  service.sendMessage(
-    saveResult.message,
+  await jobRunner.dispatchTask(Tasks.SEND_MESSAGE, {
+    message: saveResult.message,
     contact,
-    null,
+    // TODO: trx can't be supported with remote dispatch, see if it's necessary
```
schuyler1d: I don't think trx is necessary when called from here, anyway.  I wonder if instead of all our 'if trx....` from twilio.sendMessage, we should just start the transaction *inside* sendMessage (not in this PR, but later)

src/server/api/mutations/updateContactTags.js
```diff
@@ -25,29 +26,19 @@ export const updateContactTags = async (
...
-          .catch(err => {
-            // eslint-disable-next-line no-console
-            console.error(
-              `Error executing handler.onTagUpdate for ${handler.name}, campaignContactId ${campaignContactId} error ${err}`
-            );
-          });
-        promises.push(tagUpdatePromise);
+    );
+    if (handlers && handlers.length > 0) {
+      // TODO: see if there is a better way to avoid unnecessary dispatches
```
schuyler1d: this looks good, though I wonder if we could pass the handlers list by-key maybe -- i.e. currently getActionHandlersAvailableForTagUpdate gets the loaded objects, but maybe it should just return the keys, which you can pass to the dispatchTask.

src/server/api/schema.js
```diff
@@ -7,20 +7,13 @@ import { gzip, makeTree, getHighestRole } from "../../lib";
...
 import twilio from "./lib/twilio";
 
-import {
-  assignTexters,
-  dispatchContactIngestLoad,
-  exportCampaign,
-  importScript,
-  loadCampaignCache
-} from "../../workers/jobs";
+import { loadCampaignCache } from "../../workers/jobs";
```
schuyler1d, line 11: This should be a task, as well -- doesn't have to block this PR, but can you add a TODO

schuyler1d (commented):
looks really solid -- agree with all your choices.

hiemanshu (commented):
Overall looks good to me! Tested it on lambda, works relatively well. 

Couple of changes I'd love to see:
- A `job_result` table that a job can post results to once done.
- `payload` should be optional, for example, if I'm working on a job that works on a campaign (provisioning messaging services, etc), and need to convert it to a job, I don't need to actually send the job any info other than `campaign_id` which is a part of the payload, so allowing `payload` to be nil would allow this to happen without sending dummy payload.

hiemanshu:
Also here's the code for serverless to give permissions to the app to invoke the worker correctly:

```    
- Effect: 'Allow'
      Action:
      - 'lambda:InvokeFunction'
      Resource:
      - Fn::Join:
        - ':'
        - - arn:aws:lambda
          - Ref: AWS::Region
          - Ref: AWS::AccountId
          - function:${self:service}-${opt:stage, self:provider.stage}-worker
```

schuyler1d:
>     * A `job_result` table that a job can post results to once done.

@hiemanshu FYI, this is pretty much what the `campaign_admin` table is for -- it currently has fields for contact-loaders, but if there are other campaign job result data, then you can add columns there.

matteosb:
@hiemanshu I think job failure tracking in the database is decoupled from the job running plugin, though I agree that it would be good to have some uniform failure handling for all jobs. I would recommend implementing that in `invokeJobFunction`. As @schuyler1d points out this is already implemented for a subset of jobs with `campaign_admin`. 

Re: empty payload. I'm not sure I understand the use case but I see no reason not to allow an empty payload. Where does this break exactly? Looking at the code it seems like it should already work .

hiemanshu:
> > ```
> > * A `job_result` table that a job can post results to once done.
> > ```
> 
> @hiemanshu FYI, this is pretty much what the `campaign_admin` table is for -- it currently has fields for contact-loaders, but if there are other campaign job result data, then you can add columns there.

Yeah. Just seems like too many DB changes for something that could be a little more generic.

hiemanshu:
> @hiemanshu I think job failure tracking in the database is decoupled from the job running plugin, though I agree that it would be good to have some uniform failure handling for all jobs. I would recommend implementing that in `invokeJobFunction`. As @schuyler1d points out this is already implemented for a subset of jobs with `campaign_admin`.

Yeah. Probably want to make changes to `invokeJobFunction` and the `dispatchJob` task to try and handle retries too. One with exponential backoff for certain things. 

> Re: empty payload. I'm not sure I understand the use case but I see no reason not to allow an empty payload. Where does this break exactly? Looking at the code it seems like it should already work .

https://github.com/MoveOnOrg/Spoke/blob/main/src/server/models/job-request.js#L15 Trying to send an empty payload doesn't work

schuyler1d:
> > > ```
> > > * A `job_result` table that a job can post results to once done.
> > > ```
> > 
> > 
> > @hiemanshu FYI, this is pretty much what the `campaign_admin` table is for -- it currently has fields for contact-loaders, but if there are other campaign job result data, then you can add columns there.
> 
> Yeah. Just seems like too many DB changes for something that could be a little more generic.

I feel like I'm mostly the one pushing for genericity :-), but here I'd like us to 'know' what the data is in the schema -- previously, we just didn't delete `job` records and had data in there, but that made things a lot more complex and kept data that didn't need to be around.  Maybe you could give some use-cases?

I also think this is the kind of thing that will often exist in several 'dispatcher' frameworks already (including backoff algorithms, e.g. in SQS) -- so maybe instead of building a table, we can make an optional API extension to the dispatch API that could lookup/store success-type data if/when it's available to integrate somewhere.

matteosb:
@schuyler1d I strongly agree with that as a long term plan: job tracking should be delegated to the job runner itself. Some runners like bull have it built in, others will probably need to implement it themselves using a db table like job_request. For now, I would like to limit scope creep on this PR and not touch the JobRequest model.

PR #1665, opened 2020-07-07
Add sorting by timezone to campaign list

### LINKED ISSUE #1007: Campaigns List sortable by time zone
Would be ideal to select campaigns list display by campaign number, candidate, or launch date. Please consult Dylan for preferred default setting. I prefer campaign # or launch date. By time zone would be awesome, too!
Order needs to accommodate campaigns ### and ###X; if that’s a problem, we can designate certain campaigns in a way other than by adding an X to the campaign number. 
@lperson @joemcl

alliejones:
# Fixes #1007 

## Description

You can now sort by timezone on the campaign list. I discussed this with Sky during the hacknight and we decided sorting by the timezone name was good enough (that's the only timezone info currently available in the client). The timezone column is only displayed when it is the active sort option to try and save some space on small screens.

<details>
  <summary>Screenshots</summary>

View with default sort (same as before)
image

Sorting by timezone with column visible
image

</details>

## Questions I have:
- Did I put these tests in the right place? Do you typically write tests for UI changes of this size? I also haven't used Enzyme before so I'm not sure about good patterns.
- I did try to test the Campaign view with in-browser mobile emulation but it seems like it isn't mobile-friendly, is that right?

# Checklist:

- [x] I have made any necessary changes to the documentation **(don't think any are required here)**

### DIFF
--- a/src/components/AdminCampaignList/CampaignTable.jsx
+++ b/src/components/AdminCampaignList/CampaignTable.jsx
@@ -11,6 +11,7 @@ import theme from "../../styles/theme";
 import Empty from "../Empty";
 import DataTables from "material-ui-datatables";
 import { CircularProgress } from "material-ui";
+import { SORTS, TIMEZONE_SORT } from "./SortBy";
 
 const inlineStyles = {
   past: {
@@ -62,7 +63,8 @@ export class CampaignTable extends React.Component {
     onPreviousPageClick: PropTypes.func,
     onRowSizeChange: PropTypes.func,
     campaignsToArchive: PropTypes.array,
-    campaignsWithChangingStatus: PropTypes.array
+    campaignsWithChangingStatus: PropTypes.array,
+    currentSortBy: PropTypes.oneOf(SORTS.map(s => s.value))
   };
 
   state = {
@@ -129,8 +131,21 @@ export class CampaignTable extends React.Component {
       });
     }
 
+    const timezoneColumn = [];
+    // only show the timezone column when we're currently sorting by timezone
+    if (this.props.currentSortBy === TIMEZONE_SORT.value) {
+      timezoneColumn.push({
+        key: "timezone",
+        label: "Timezone",
+        sortable: false,
+        style: {
+          width: "5em"
+        }
+      });
+    }
+
     return [
-      // id, title, user, contactcount, unassigned, unmessaged, due date, archive
+      // id, timezone (if current sort), title, user, contactcount, unassigned, unmessaged, due date, archive
       {
         key: "id",
         label: "id",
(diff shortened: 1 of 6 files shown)


src/components/AdminCampaignList/CampaignTable.jsx
```diff
@@ -278,7 +294,6 @@ export class CampaignTable extends React.Component {
     const { campaigns, pageInfo } = this.props.data.organization.campaigns;
     const { limit, offset, total } = pageInfo;
     const displayPage = Math.floor(offset / limit) + 1;
-    console.log("CampaignTable", campaigns);
```
alliejones, line 281: Should this stay in the code? I thought it was from my own debugging but apparently it was here before!

joemcl:
Cool and thanks @alliejones ! @lperson this would be great to have finally. Hey @suegibson wishes sometimes do come true :) .

schuyler1d (approved):
:100:

src/components/AdminCampaignList/CampaignTable.jsx
ibrand: I'm a big fan of removing console logs, but I also see them all over the codebase so I'm not sure. @schuyler1d ?

schuyler1d: I don't want a concerted efforts to delete console.logs (yet -- we have some plans around swapping out logging frameworks for better flexibility), we can delete specific instances, especially if they're annoying.

Anyway, this delete is good and fine.

joemcl:
Thanks and if this can be merged to a staging branch, I can deploy that on our staging instance and test that later today.

joemcl:
@alliejones finally have stage-main deployed on a staging instance to test this and a bunch of other stuff out. The time zones are sorted by name, which is cool. But having a default sort, or an optional sort, by timezone offset from UTC would be handy - so rather than the default sort order be alpha - Central, Eastern, Mountain, Pacific - time zones sorted by UTC offset would be handy - like Eastern, Central, Mountain, Pacific. @lperson would adding timezone UTC offset values to Texting Hours be doable, or is it already in there maybe?

joemcl:
Okay, rereading this and seeing that time zone sort by name is intentional, got it. @lperson If you can add the UTC offset value for each time zone to Texting Hours, that will be handy, and then sort by time zone could use that value, right @alliejones ?

alliejones:
@joemcl Yes, I think that should should be straightforward if we can get the UTC offset from the backend.

joemcl:
@alliejones cool. @lperson adding UTC offset to Texting Hours for timezone sorting by that would be handy.

PR #1675, opened 2020-07-09
job-runner dispatch changes: single lambda dispatch and campaignCache as a Task

schuyler1d:


src/server/models/cacheable_queries/campaign-contact.js
```diff
@@ -410,57 +411,39 @@ const campaignContactCache = {
...
-          stream.pipe(cacheSaver);
-        });
-        return result
-          .then(done => {
-            console.log("updateCampaignAssignmentCache Completed", campaignId);
-          })
-          .catch(err => {
-            console.log("updateCampaignAssignmentCache Error", campaignId, err);
-          });
+    //try {
```
matteosb, line 414: nit: can we get rid of the commented try/catch if it isn't being used?

matteosb (approved):

PR #1678, opened 2020-07-09
Per-campaign phone number management

matteosb:
## Description

Allows campaigns to select phone numbers from the inventory by area code. Each campaign gets its own Messaging Service, so this an effective way to scale beyond the 400 number limit and does not require interacting with Twilio. I also had to introduce the concept of permanently archiving campaigns to guard against a campaign being unarchived after having its numbers released.

The plan is to eventually move phone number allocation strategies into integrations/, but I think it's worth merging this before that.

# Screenshots
Campaign edit:
AdminCampaignEditPhoneNumbers
Archived campaign:
ArchivedCampaignReleaseNumbers

Archived campaign after numbers are released
ArchivedCampaignCantBeUnarchived

# Checklist:

- [X ] I have added tests that prove my fix is effective or that my feature works


schuyler1d (approved):
I'm ok with merging this but I have a couple of flags in terms of the direction:

* I'm hoping we can back out of adding new mutations/endpoints (e.g.release messaging service)

Both in terms of adding these and creating 'new workflow' for each of them worries me.  Why not just automatically delete it on-archive and disallow unarchiving campaigns that were configured that way rather than have an extra step -- that way it's connected to existing workflow which other strategies can work with.

*  I'm concerned with custom logic in AdminCampaignEdit on pending jobs -- I think it would be better for that to be a more generic endpoint similar to organization.fullyConfigured or something
* Many client-side tests are on globals which are across the instance -- this area specifically needs to be per-org (ideally per-campaign so folks can experiment before they turn a strategy on)

matteosb:
@schuyler1d responses below. I agree that there are some sub-optimal bits here, but getting this in a state where phone number allocation is "pluggable" is a really big task, so hopefully this gets us one step closer to that.

```
* I'm hoping we can back out of adding new mutations/endpoints (e.g.release messaging service)
```
This one is tough, I don't think we can offer as good an experience while staying clear of adding operations but I'm open to suggestions.
* One option would be to separate them from the core operations and types even more, e.g to create new gql types for CampaignPhoneNumbers, rather than extending Campaign and Organization types.
* Another option could be to use more generic fields: `Campaign.extensions` and `Organization.settings`, rather than continuing to add specific fields to each type.

In the future, I think the more principled way of doing this would be to use GQL `extend type` to load extensions from `integrations`. I think that probably requires upgrading graphql-js (and maybe apollo as well). 

```
Why not just automatically delete it on-archive and disallow unarchiving campaigns that were configured that way rather than have an extra step -- that way it's connected to existing workflow which other strategies can work with
```
The reason I did it this way is that archiving is the only way to temporarily pause a campaign and it's really easy to hit archive accidentally. In the Warren version we did release on archive but we also added a confirmation modal to "archive" and added states to temporarily pause a campaign.

```
* I'm concerned with custom logic in AdminCampaignEdit on pending jobs -- I think it would be better for that to be a more generic endpoint similar to organization.fullyConfigured or something
```

I think campaign starting has to be treated a bit differently from other jobs in the UI though. We could supply `isStarting` from the backend to abstract it out a bit?  I would prefer to see a `state` enum (NEW, STARTING, STARTED, ARCHIVED, ARCHIVED_PERMANENTLY) take the place of the boolean flags, but that's a bigger refactor.

```
* Many client-side tests are on globals which are across the instance -- this area specifically needs to be per-org (ideally per-campaign so folks can experiment before they turn a strategy on)
```

I think this is only the case for CONTACTS_PER_PHONE_NUMBER. Because this is largely driven by carrier policy I don't expect it to change per org, but I'm happy to move that to Organization like perCampaignPhoneNumbers. perCampaignPhoneNumbers is at the org rather than the campaign level for simplicity. I think we can make it per-campaign, but would really have to think through all the edge cases and the UX. I think it would take some time to get right.

schuyler1d:
> One option would be to separate them from the core operations and types even more, e.g to create new gql types for CampaignPhoneNumbers, rather than extending Campaign and Organization types.

I like this -- again, not nec. for this PR but let's plan to go in this direction, so it's at least 'contained' away from adding too much complexity into Campaign and Organization which should stay understandable and as simple as possible.

> The reason I did it this way is that archiving is the only way to temporarily pause a campaign and it's really easy to hit archive accidentally. In the Warren version we did release on archive but we also added a confirmation modal to "archive" and added states to temporarily pause a campaign.

Erik was suggesting we just remove the archive buttons along the right and only provide it through the multi-select process.  That seems like a good idea even more now for this reason.

Second, I guess it's the easiest way from the Campaign List, but someone can also change the dueBy date to being in the past to 'pause'.  But I think if we want to make it easy to pause a campaign then that should be a different feature than build a lot more complexity into an area that will already have a lot of complexity already.

> I think campaign starting has to be treated a bit differently from other jobs in the UI though. We could supply isStarting from the backend to abstract it out a bit? I would prefer to see a state enum (NEW, STARTING, STARTED, ARCHIVED, ARCHIVED_PERMANENTLY) take the place of the boolean flags, but that's a bigger refactor.

Agreed -- not for this PR, but I think this is the right way.

> I think this is only the case for CONTACTS_PER_PHONE_NUMBER. Because this is largely driven by carrier policy I don't expect it to change per org, but I'm happy to move that to Organization like perCampaignPhoneNumbers. perCampaignPhoneNumbers is at the org rather than the campaign level for simplicity. I think we can make it per-campaign, but would really have to think through all the edge cases and the UX. I think it would take some time to get right.

Ok, that's fair, though it's still a bit concerning that this abstraction is leaking.

PR #1684, opened 2020-07-13
7.1 release notes

schuyler1d:
cc @matteosb (added a line about lambda-async to the aws docs)


ibrand (approved):
Yay! These are wonderful. Really excited about all the community work.

The only thing is that historically we've done the sub-sections within a release with **bold** instead of full on
### headers

I don't mind too much either way, but thought I'd mention it since it changes the hierarchy of the doc

matteosb (approved):
:clinking_glasses:

PR #1688, opened 2020-07-14
VAN input type is optional

jeffm2001:
## Description

When creating canvass responses through the VAN API, `inputTypeId` is [optional](https://developers.ngpvan.com/van-api#people-post-people--vanid--canvassresponses), and defaults to 11 ("API"). The Spoke action handler was specifying this default value explicitly — which would be fine — except that for some reason VAN's API is not returning that default inputType is not always included as one of the available options when you [get them](https://developers.ngpvan.com/van-api#canvass-responses-get-canvassresponses-inputtypes) from VAN.

That was causing errors because it made the Action Handler think it was using an illegal value. The simplest solution is to omit `inputTypeId` entirely if it's not specified in the configuration.

# Checklist:

### DIFF
--- a/__test__/integrations/action-handlers/ngpvan-action.test.js
+++ b/__test__/integrations/action-handlers/ngpvan-action.test.js
@@ -552,6 +552,228 @@ describe("ngpvn-action", () => {
       allNocksDone();
     });
 
+    describe("when there's an input/contact type configured", () => {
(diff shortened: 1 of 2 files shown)


lperson (dismissed):
LGTM. Thanks!

schuyler1d (approved):
:+1:

PR #1690, opened 2020-07-15
Remove Documentation

JeremyParker:
It looks like this tip was necessary at one point, but the Message Review process UX now looks explanatory. Now there's a `REASSIGN ALL [N] MATCHING` button that will reassign all messages in the selected state.
I might not be understanding some subtlety in the system here that requires this tip, but I think we don't need it any more.


JeremyParker:
@joemcl is this documentation still relevant?

ibrand:
@JeremyParker I think I'd like to actually do a deeper write up how how people use message review rather than delete this. I'd call it more of an "outline" rather than an actual doc at this point. But I agree with you that the doc as it stands is not very useful. That said, new admin have often had problems knowing what the message review is for.

ibrand:
Experimenting with surfacing our "outline/unfinished docs" like this: https://github.com/MoveOnOrg/Spoke/pull/1760/files#diff-aee3e3a05f8f0a70aa7a8132bee2fddcR47

schuyler1d (dismissed):
:+1:

ibrand:
@schuyler1d and I talked, and I'm actually going to 👍 this too let's go ahead and delete the doc.

@JeremyParker would you mind also removing it from [the microsite's table of contents](https://github.com/MoveOnOrg/Spoke/blob/main/docs/_sidebar.md)?

JeremyParker:
Could/should we put the existing doc into a Github Issue? That tends to be where we keep details about other work to be done.

Another offhand thought (no response necessary) - it would be cool to have a user forum where admins could trade tips on how they use the tool. Also, where users could search people's comments for the answer to their problem.

ibrand:
Yes we should totally make an issue for this! I think the doc we'd really want is "how to use message review"

 I love the idea of the forum. I really can't tell what the right approach to that would be. I really wish I could be connected to github in some way.

schuyler1d (approved):
:+1:  Agree we should delete defunct docs, though it would still be nice to have some campaigner docs that include how to use the Message Review feature

PR #1699, opened 2020-07-29
Simplify the README by changing .env.example

ibrand:
# Description

Changes `.env.example` to be fully usable out of the box if a dev is using sqlite by setting the default service to `fakeservice`. Then tighten up the README to cut out unnecessarily confusing options for new devs who want to get started. Also modify the docker and postgres instructions to more accurately fit the system we have now.


schuyler1d (approved):
Happy to merge this (even before the next version) -- (if you saw an earlier version of this comment, is was because github was showing me a single commit's changes instead of the whole set)

matteosb (approved):
:tada:

PR #1702, opened 2020-07-30
Dynamic assignment batches

### LINKED ISSUE #1533: RFC: Request/Release-style dynamic assignment + Texter Sidebox extensibility

schuyler1d:
# Fixes https://github.com/MoveOnOrg/Spoke/issues/1533

## Description

Changes Dynamic Assignment

TODO:
- [x] extensible changes to AssignmentSummary (on the todos page) from texter-sideboxes
- [x] extensible changes to assignment.hasUnassignedContactsForTexter so a choice for when/who can request another batch (e.g. vetted texters, or after replies are finshed, etc)
- [ ] change/add tests

# Checklist:

- [NO] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#your-first-code-contribution), or has a documented reason in the description why it’s longer


hiemanshu:
@schuyler1d One change that was made after we open sourced the patches was to change the `findNewCampaignContact` mutation to only reassign when `message_status` is `needsMessage`, otherwise it would reassign replies, and other side effects. 

Here: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/mutations/findNewCampaignContact.js#L70

Issue #1645, opened 2020-06-25
Feature Request: Canned response tag association


ibrand:
**Problem**
It will finish out the [tagging story](#1473) to allow canned responses to be associated with a tag so that when a texter clicks a canned response it automatically applies the tag that the admin wanted applied.

**Solution**
Create a field in canned responses for a tag that allows auto-completing with the tags that an admin has added. Allow 1-many tags to be filled in. Automatically apply those tags when that canned response is selected.

schuyler1d:
Some good examples for many-to-many canned response-tag relationships:
* Wrong Number + Strong Support. This came up a lot in recent M4BL texting (edited) 
* For GOTV: you might want responses for Moved + Already Voted, Moved + Needs Absentee Ballot, etc..
* Multiple tags per canned response was almost the default for event recruitment campaigns, since you often have different responses based on the reason that people are/are not attending an event ("Won't Attend + Strong Support + Interested in Future Events", "Won't Attend + Strong Support + No Future Events", etc...)

(from Matteo)

arena:
Just found this issue, is it supposed to be open? 
I'm about to add a similar issue about showing tags alongside the canned response they are associated with

Frydafly:
Closing since adding multiple tags to canned responses can already be done! Thanks @agreenspan24

PR #1705, opened 2020-07-31
Feature #1645: Canned response tag association

aschneit:
# Fixes # (issue)
#1645 
## Description

Built ui to add/remove tags to canned responses and have them appear in canned response list. Added join table tag_canned_response and built db mutations and queries to edit/fetch canned response data with corresponding tag id data.

(This does not include the application of the tags when a texter uses a canned response, which it has been suggested should be separated into another pr)

# Checklist:

### DIFF
--- a/__test__/components/CampaignCannedResponseForm.test.js
+++ b/__test__/components/CampaignCannedResponseForm.test.js
@@ -14,13 +14,38 @@ describe("CampaignCannedResponseForm component", () => {
     defaultValue: {
       id: 1,
       title: "Response1",
-      text: "Response1 desc"
-    }
+      text: "Response1 desc",
+      tagIds: [1, 2]
+    },
+    tags: [
+      {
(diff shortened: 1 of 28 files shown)


schuyler1d (commented):
looking good! Let's just get the tests passing.

If you hide the admin interface changes behind EXPERIMENTAL_TAGS enabled, then I can merge this without waiting for the frontend part of adding the tags when a canned response is sent.  Otherwise, that's the next step!

aschneit:
@schuyler1d I fixed the issue causing the canned response test to fail ("should save campaign canned responses across copies and match saved data" in campaign.test.js), but there are still some tests failing (in CampaignInteractionStepsForm, ConversationPreviewModal) that don't seem to have anything to do with the work that I did. I'm looking into is as best I can, but if you happen to have a chance to take a look, that would be great, thanks.

PR #1709, opened 2020-07-31
Add an environment variable for database debugging

tekkamanendless:
## Description
This introduces the `DB_DEBUG` environment variable, which toggles
on the `debug` value for Knex.  When this is true, the database
queries will be logged to the console, making debugging and
troubleshooting easier.

The default value for the new environment variable is false.


src/server/knex-connect.js
```diff
@@ -26,6 +27,8 @@ const useSSL = DB_USE_SSL === "1" || DB_USE_SSL.toLowerCase() === "true";
 if (useSSL) pg.defaults.ssl = true;
 // see https://github.com/tgriesser/knex/issues/852
 
+const debug = DB_DEBUG === "1" || DB_DEBUG.toLowerCase() === "true";
```
jeffm2001: A small nitpick, but I don't see a need for an extra variable here. I think it would be cleaner to make the line down below `config.debug = DB_DEBUG === "1" || DB_DEBUG.toLowerCase() === "true";`

tekkamanendless: Fixed :+1:

jeffm2001 (approved):
Looks good!

PR #1710, opened 2020-08-01
Receive Twilio messages from behind a proxy using TWILIO_MESSAGE_CALLBACK_URL

tekkamanendless:
## Description

When running behind a proxy (for example, using an Apache server to proxy a virtual host to the Node server running Spoke), the default Twilio header verification fails because the URL is detected incorrectly.
Instead of discovering the external-facing URL that Twilio used, the code will only find the internal URL.

This change introduces `TWILIO_MESSAGE_CALLBACK_URL`, which is analogous to `TWILIO_BASE_CALLBACK_URL` (which is used for report status callbacks).

In principle, we could merge those two into a single environment variable that covers the protocol and domain, but that wouldn't be backward-compatible, and a single additional environment variable to (optionally) tweak isn't bad.

In addition, this also enables Twilio header verification on the `/twilio-message-report` endpoint, too.  Previously, this endpoint was uncovered and it could have been used as an attack point for an attacker to pretend to be Twilio and publish fake status updates.


schuyler1d (approved):
:+1:

PR #1711, opened 2020-08-02
Remove the whole-page scrollbar by fixing the sidebar height

tekkamanendless:
## Description

The header height is defined as 65px in `src/components/TopNav.jsx`, but the sidebar height calculation incorrectly used 56px (in `src/components/AdminDashboard.jsx`).

This fixes the sidebar height calculation by using the correct header height.

### DIFF
--- a/src/components/AdminDashboard.jsx
+++ b/src/components/AdminDashboard.jsx
@@ -13,7 +13,7 @@ const styles = StyleSheet.create({
     ...theme.layouts.multiColumn.container
   },
   sidebar: {
-    minHeight: "calc(100vh - 56px)"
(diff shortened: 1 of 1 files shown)


JeremyParker (approved):
This looks reasonable to me. A corresponding issue with a screen shot of the issue would make reviewing easier, but in this case it seems pretty clear that it's a reversal of the digits.
👍

tekkamanendless:
Oh, sure, for posterity:

Original (bad; unnecessary scrollbar on right-hand side):
height-broken

Proposed (good, no scrollbar needed):
height-fixed

Issue #207, opened 2017-09-12
Auto-refresh replies and other ToDo-screen data


schuyler1d:
Use Cases:
* User is 'staring at the screen' of ToDos -- waiting for things to reply to.
* When the user 'clicks back', does the content refresh then?

Currently, texters often have to hard-refresh to see replies. The number of replies does not update either (#178)

Tech context:
#162 improved this by increasing poll frequency, and #198 added an automatic refresh after the user returns to the ToDo list from the messaging interface. But some other things should update.

UI: Should we 'ask to notify user' (via desktop) and push a notification when there are new replies found?

## Possible solution (for someone wanting to do this work):
* Assignments are too complex to cache, however when caching is available we *could* cache a notification when a new message came in.  Then, instead of polling the whole TodosList query, we could poll whether any notifications were present, and when they are they could request stats on the whole todos page.  Somewhere around here: https://github.com/MoveOnOrg/Spoke/blob/main/src/server/models/cacheable_queries/message.js#L320 on message.save we could add a notification flag if the campaignContact assignment (and user id) was cached (https://github.com/MoveOnOrg/Spoke/blob/main/src/server/models/cacheable_queries/campaign-contact.js#L43).
* A new mutation would need to be made that gets back whether there are notifications, and when there are both clears them and returns a true result.
* A `todos` query should also clear notifications from cache for the user.

sandramchung:
I've looked into and tested a couple of options explained here: http://dev.apollodata.com/react/receiving-updates.html and here: http://dev.apollodata.com/react/cache-updates.html
We're currently doing polling, refetch, and normalization with dataID from object. Polling addresses updating the ToDo list while someone's staring at it (changes when admin edits assignments and when user sends text or receives replies); refetch updates the ToDo list when we close out of the texting interface and return to the ToDo list. 

We could replace the first two with subscriptions in the reply/todos interface, and that would probably have the most impact on DB tax; but implementing the websocket required by subscriptions isn't a good option for us at this time.

Next I'll try monitoring the load on the DB during campaign usage. Polling can get pretty harsh, but it may be that not enough users leave the ToDo list screen open long enough for it to matter most of the time.

sandramchung:
Last weekend I didn't see db activity at problematic levels, but it's worth checking on again as we scale up on Spoke usage.
During Monday's demo Sky and I noted that polling on the ToDo List appears to have ceased.  Suspect pollInterval may be incompatible with refetch if implemented on the same query/container.

[...]

PR #1713, opened 2020-08-03
Fix polling on the todos screen

### LINKED ISSUE #1712: Bug: Texter TODO list no longer auto-reloads every 5 seconds
**Describe the bug**
The intent of the texter TODO list is to present a kind of dashboard of the texts that they have available to deal with.  For example, after replying to all of the text messages, the texter shoudl be able to remain on that screen, and as replies come in, the screen should update with the appropriate buttons and badges (for example, "Respond (1)").

After combing through the source code, it looks like this functionality was broken on June 3, 2019, with a change to how user data is cached.

Short version of the story:
1. [src/containers/TexterTodoList.jsx](https://github.com/MoveOnOrg/Spoke/blame/main/src/containers/TexterTodoList.jsx#L57) contains this line to control the auto-refresh:
    ```if (this.props.data.currentUser.cacheable && !this.state.polling) {```
2. Thus, it will begin auto-refresh if (1) `cacheable` is true and (2) auto-refresh hasn't started yet.
3. [src/server/api/user.js](https://github.com/MoveOnOrg/Spoke/blame/main/src/server/api/user.js#L274) forces `cacheable` to be false no matter what.
4. Thus, auto-refresh will never begin.

**To Reproduce**
1. Create a texting campaign, assign a user 2 contacts.
2. As that user, send the "first message" to 1 of the contacts.
3. Go back to the `todos` screen.
4. See that there is 1 "first text" still to send, and nothing to reply to.
5. As the contact, reply to that first message.
6. Wait 10 seconds.
7. As the user, note that the `todos` screen has not changed, even though a reply has come back.

**Expected behavior**
Step (7) above should instead show a "Respond (1)" button, telling the user that there has been a response to that first text message.

**Screenshots**
n/a

 - OS: Linux (but applies to all OSes)
 - Browser: Chrome (but applies to all browsers)
 - Desktop or Mobile: Desktop (but applies to all architectures)
 - Version: 84 (but applies to all versions)

**Additional context**
This is the first relevant change (on April 24, 2019): https://github.com/MoveOnOrg/Spoke/commit/4f9585db3cf6564282264cf07d6f8410619e5fb4
This change introduces the requirement around `cacheable` needing to be true in order to start auto-refresh ("polling").

This is the second relevant change (on June 3, 2019): https://github.com/MoveOnOrg/Spoke/commit/dbbb20acb2bde3ca3daf5936dc115dadcfb2590c
This change forces `cacheable` to always be false.

My take is that the `todos` screen shouldn't care at all about `cacheable`, since we know for a fact that the data should be refreshed no matter what on that screen.

On my machine, I've simply removed the `this.props.data.currentUser.cacheable && ` part, and it's working as expected now.

tekkamanendless:
# Fixes #1712 
For more information, see: https://github.com/MoveOnOrg/Spoke/issues/1712

## Description

This fixes polling on the texter `todos` screen, making it so that the information is reloaded every 5 seconds as long as the screen is up.  This seems to have been broken by an older commit.

The fix here is to simply ignore the user's `cacheable` property, since (1) it's hard-coded to false right now, and (2) it doesn't matter.  The data here, by definition, needs to be queried every time so that we can see a fairly "live" view of the text messages that need to be dealt with.

# Checklist:

### DIFF
--- a/src/containers/TexterTodoList.jsx
+++ b/src/containers/TexterTodoList.jsx
@@ -54,7 +54,7 @@ class TexterTodoList extends React.Component {
   componentDidMount() {
     this.props.data.refetch();
     // stopPolling is broken (at least in currently used version), so we roll our own so we can unmount correctly
-    if (this.props.data.currentUser.cacheable && !this.state.polling) {
+    if (!this.state.polling) {
       const self = this;
       this.setState({
         polling: setInterval(() => {


schuyler1d:
Hi, unfortunately, it's not that easy.  "Just" turning polling on will tax the spoke database too much if there are not even too many texters.  Let's discuss further in the issue:
https://github.com/MoveOnOrg/Spoke/issues/207

PR #1715, opened 2020-08-03
fix GSScriptEditor for Safari browsers that auto-focus

schuyler1d:
frustratingly, without onFocus, this editor breaks when tabbing into
 the field -- no editor dialog comes up
 However, on Safari, when the field is created, Safari seems to auto-focus
 which triggers a disruptive (early) dialog open, e.g. in Admin Interactions

### DIFF
--- a/src/components/forms/GSScriptField.jsx
+++ b/src/components/forms/GSScriptField.jsx
@@ -92,7 +92,17 @@ export default class GSScriptField extends GSFormField {
             this.handleOpenDialog(event);
           }}
           onFocus={event => {
-            this.handleOpenDialog(event);
+            // HACK
+            // frustratingly, without onFocus, this editor breaks when tabbing into
+            // the field -- no editor dialog comes up
+            // However, on Safari, when the field is created, Safari seems to auto-focus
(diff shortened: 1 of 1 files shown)


src/components/forms/GSScriptField.jsx
```diff
@@ -92,7 +92,17 @@ export default class GSScriptField extends GSFormField {
             this.handleOpenDialog(event);
           }}
           onFocus={event => {
-            this.handleOpenDialog(event);
+            // HACK
+            // frustratingly, without onFocus, this editor breaks when tabbing into
+            // the field -- no editor dialog comes up
+            // However, on Safari, when the field is created, Safari seems to auto-focus
+            // which triggers a disruptive (early) dialog open, e.g. in Admin Interactions
+            const isSafari = /^((?!chrome|android).)*safari/i.test(
```
ibrand, line 100: one nit but don't let this stop you from merging: I'm curious if we could have this definition of `isSafari` live somewhere library-like so that this form of testing for safari could be reusable? Maybe in `src/lib/is-client.js`

PR #1722, opened 2020-08-06
Remove the scroll bars in every cell in the incoming-messages table

tekkamanendless:
## Description

The incoming-messages table was set so that every cell had `overflow: scroll` set, meaning that every cell would have both horizontal and vertical scroll bars, even when they weren't necessary.

The styles also had `text-overflow: ellipsis`, so there's really no reason to force scrolling in the cell when the text will be cut off appropriately.

For all of these `overflow: scroll` styles, I've replaced them with `overflow: hidden`, as it is in a couple other places in that screen.

### Screenshots

Before (bad, scrollbars everywhere):
table-broken

After (good, no scrollbars):
table-fixed

# Checklist:

### DIFF
--- a/src/components/IncomingMessageList/index.jsx
+++ b/src/components/IncomingMessageList/index.jsx
@@ -123,7 +123,7 @@ export class IncomingMessageList extends Component {
       label: "Texter",
       style: {
         textOverflow: "ellipsis",
-        overflow: "scroll",
+        overflow: "hidden",
         whiteSpace: "pre-line"
       }
     },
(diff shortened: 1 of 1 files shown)


schuyler1d (approved):
interesting that I've never seen the scrollbars.  Would be interested to know what browser you're using.

In any case, happy to make things better on every platform! Thanks!

tekkamanendless:
Are you using a Mac, by any chance?  They tend to hide scroll bars unless you're mousing over them.  We tend to run into the opposite problem at work: Mac customers can't find the scrollbar that's obvious on every other OS.

PR #1723, opened 2020-08-06
Log requests using morgan

tekkamanendless:
## Description

This makes Spoke log like a normal web server, including the source address, target URL, status code, byte count, and user agent.  This information is invaluable when troubleshooting web server issues.

### Example

You can see the standard morgan `combined` logs that were requested as part of the normal refresh process in the UI.

```
2:16:07 AM server.1  |  backend info Checking if zip code is needed
2:16:07 AM server.1  |  backend info Node app is running on port 8090
2:16:16 AM server.1  |  192.168.86.1 - - [06/Aug/2020:02:16:16 +0000] "POST /graphql HTTP/1.1" 200 984 "https://spoke.XXXX/admin/1/campaigns/1/edit" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.105 Safari/537.36"
2:16:27 AM server.1  |  192.168.86.1 - - [06/Aug/2020:02:16:27 +0000] "POST /graphql HTTP/1.1" 200 7160 "https://spoke.XXXX/admin/1/campaigns/1/edit" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.105 Safari/537.36"
```

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,6 +127,7 @@
     "minilog": "^3.0.1",
     "moment": "2.14.1",
     "moment-timezone": "^0.5.14",
+    "morgan": "^1.10.0",
     "nexmo": "^2.3.2",
     "node-abort-controller": "^1.0.4",
     "node-fetch": "^2.6.0",
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -1,6 +1,7 @@
 import "babel-polyfill";
 import bodyParser from "body-parser";
 import express from "express";
+import morgan from "morgan";
 import appRenderer from "./middleware/app-renderer";
 import { graphqlExpress, graphiqlExpress } from "apollo-server-express";
 import { makeExecutableSchema, addMockFunctionsToSchema } from "graphql-tools";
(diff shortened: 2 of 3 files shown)


tekkamanendless:
If you want something backward compatible and with more options, then we could add an environment variable (defaulting it to empty):

```
ACCESS_LOG_FORMAT=
```

And then if you wanted to turn on logging, you would do:

```
ACCESS_LOG_FORMAT=combined
```

or:

```
ACCESS_LOG_FORMAT=':method :url :status :response-time ms - :res[content-length]'
```

(for example)

Then, when we go to set up logging, we'd check to see if `ACCESS_LOG_FORMAT` was set to something; if so, then turn on logging and use that format; otherwise, don't do any logging, which is what we used to do.

schuyler1d:
> This makes Spoke log like a normal web server...

I think this is an argument against including this dependency.  Spoke is most commonly deployed in Heroku and AWS both of which have sophisticated request logging.  I would strongly recommend against deploying Spoke as a bare server -- if for no other reason than it should be wrapped by an https protocol.

Issue #510, opened 2018-02-15
Support pre-approved emojis in scripts/canned messages


schuyler1d:
Note: we should only use emojis which have been standardized across platforms:
https://www.twilio.com/blog/2015/08/common-sms-problems-unicode-twilio.html

To start, we would like to add some emoji insert buttons on the Admin page for canned answers and when editing a script.

After that, we would like to make it so admins for can white-list specific emojis for a particular campaign (i.e. it's a campaign setting) where it would expose the same insert buttons to the campaign's texters.

The first part is a `good-first-issue` though this will require a little familiarity with unicode

ibrand:
I'm cleaning up our `good first issues`s and want to check in on the status of this. This hasn't been implemented has it?

ibrand:
@schuyler1d 
Sounds like this should actually be broken into 3 issues:

2 Good First Issues:
1. Add an emoji palette to the interaction script editor in the Campaign creation page
2. Add an emoji palette to canned responses in the Campaign creation page

and then a meatier issue:
3. Add ability to add an emoji safe-list to a campaign.
*with the subtasks of:*
- Acquire design for where in the create campaign page we would add this safelist and how it could be constructed (perhaps a task for a Progressive HackNight design inclined person)
- Implement the admin facing safelist creator
- If that safelist exists, provide a safe-list approved emoji palette to the texter view

Does that seem right? If so, can I close this with links out to the 3 new smaller issues?

schuyler1d:
I think that's the right breakdown of tasks.  Except maybe the third -- For the third, it's probably that we should error or warn on saving (or writing) emoji that aren't cross-platform, if that's possible (and forward-compatible).

As far as splitting the tasks -- I'm ambivalent -- e.g. I don't think we need two separate emoji palette  components for (1) and (2) so they seem pretty interconnected.  If a single one is done first, and then the other uses the work from that one to move forward, then that's fine -- but why not just have 3 PRs that link to this issue (and have checkboxes or something)

ibrand:
Totally. I'll make them sub tasks!

michaelxh:
Taking a first pass at this issue here:
https://github.com/michaelxh/Spoke/tree/510-support-white-listed-emojis

Arique1104:
Here is an example of how Elizabeth Warren's Campaign hooked up emojis.

https://github.com/Elizabeth-Warren/Spoke/blob/master/src/containers/ConversationTexter/ConversationTexterContact.jsx

PR #1729, opened 2020-08-07
[WIP] 510 emoji picker in campaign interactions and canned responses

michaelxh:
# Fixes https://github.com/MoveOnOrg/Spoke/issues/510

## Description
Adds emoji selector to script editor in campaign page allowing users to add emojis to scripted interactions and canned responses.

emoji-editor

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "dataloader": "^1.2.0",
     "dotenv": "^2.0.0",
     "draft-js": "^0.7.0",
+    "emoji-mart": "^3.0.0",
     "express": "^4.14.0",
     "fs": "^0.0.2",
     "google-libphonenumber": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
     "camelcase-keys": "^4.1.0",
     "chart.js": "^1.1.1",
     "cookie-session": "^2.0.0-alpha.1",
+    "css-loader": "^4.2.1",
     "dataloader": "^1.2.0",
     "dotenv": "^2.0.0",
     "draft-js": "^0.7.0",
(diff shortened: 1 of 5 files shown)


michaelxh:
Happy to add tests but unsure of what would be helpful. Also unsure if there is documentation available but I think something around how messages with emojis are bigger/more 'expensive' could be useful, though the twilio segment documentation does discuss that.

package.json
```diff
@@ -155,13 +157,14 @@
...
     "rollbar": "^2.4.4",
+    "style-loader": "^1.2.1",
     "thinky": "^2.3.3",
     "timezonecomplete": "^5.5.0",
     "twilio": "^3.40.0",
     "url-join": "^4.0.1",
     "uuid": "^3.1.0",
     "wait-for-expect": "^1.1.1",
-    "webpack": "^3.8.1",
+    "webpack": "^4.44.1",
```
schuyler1d, line 167: will need to confirm that this version of webpack works in deploying, but if so, this upgrade would be very welcome!

src/components/EmojiPicker.jsx
```diff
@@ -0,0 +1,129 @@
...
+    this.state = {
+      x: 0,
+      y: 0
+    };
+
+    this.handleClickOutside = this.handleClickOutside.bind(this);
+  }
+
+  componentDidMount() {
+    const bounds = ReactDOM.findDOMNode(this.picker).getBoundingClientRect();
```
schuyler1d, line 22: rather than go so low-level like this could we use:
`import Popover from "material-ui/Popover";`
See src/components/AssignmentTexter/Controls.jsx for an example.

schuyler1d (changes requested):
Great feature!! Let me know if you have time to use `Popover` from material-ui, otherwise we can work with what you have.  Thanks!

Frydafly:
hey @schuyler1d  & @michaelxh -- checking on the status of this! Looks like a few changes were in order to make this possible?

schuyler1d:
I think the webpack upgrade was a blocker at the time, but FWIW, another [react upgrade PR](https://github.com/MoveOnOrg/Spoke/pull/1930) also has that upgrade with fixes for the issues we hit before which might unblock this.

Frydafly:
@schuyler1d does the react upgrade unblock this?

PR #1733, opened 2020-08-07
opt out record column in export: distinguishes between whether they opted out in this campaign or any campaign

schuyler1d:
optOut in export is whether the contact opted out in THAT campaign -- instead of whether in any one during the campaign's duration.

### DIFF
--- a/src/workers/jobs.js
+++ b/src/workers/jobs.js
@@ -759,6 +759,7 @@ export async function exportCampaign(job) {
         "contact[optOut]": optOuts.find(ele => ele.cell === contact.cell)
           ? "true"
           : "false",
+        "contact[optOutRecord]": contact.is_opted_out,
         "contact[messageStatus]": contact.message_status,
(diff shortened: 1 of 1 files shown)


src/workers/jobs.js
```diff
@@ -759,6 +759,7 @@ export async function exportCampaign(job) {
         "contact[optOut]": optOuts.find(ele => ele.cell === contact.cell)
           ? "true"
           : "false",
+        "contact[optOutRecord]": contact.is_opted_out,
```
bdatkins, line 762: Wondering if "contact[optOutRecord]" might not clear up the confusion at it's not much different than "contact[optOut]".  Maybe rename to contact[optOutCampaign] or similar to differentiate as the campaign contact opt out flag.

bdatkins (changes requested):

schuyler1d:
closing because this exists in another branch that will be merged -- good comment @bdatkins -- will tweak in that branch

PR #1738, opened 2020-08-11
Copy all relevant campaign properties when copying a campaign

tekkamanendless:
## Description

(If you take this one, then you can ignore #1736, which just fixes the interaction step ordering.)

I typically segment my data a bit and then copy the same campaign over and over, changing the list and the intro text a bit each time, but everything else generally stays the same.

I noticed some issues when copying a campaign, however:

1. There is no notification that the "Copy Campaign" button did anything; I mashed it many times before I realized what was happening.
2. Nothing is copied from "Basics" except the title.
3. Nothing is copied from "Dynamic Assignment".
4. Nothing is copied from "Texter Experience".
5. Nothing is copied from Texting Hours.
6. The questions and responses are not copied in the right order, but the order seems close to the correct order at times.

This patch addresses all of those issues.  To the best of my knowledge, all of a campaign's configuration is copied over to the new campaign.

In addition, we show a "confirm" box upon successful copy completion offering to take the user to the edit screen for the new campaign. If she accepts, then we'll re-route to that screen.  If she does not, then she'll stay where she is, but at least she'll know that the copy operation succeeded.

## Screenshots

Here is what it looks like once the copy has completed.
copy-message

# Checklist:

### DIFF
--- a/src/containers/AdminCampaignStats.jsx
+++ b/src/containers/AdminCampaignStats.jsx
@@ -166,17 +166,6 @@ class AdminCampaignStats extends React.Component {
     );
   }
 
-  renderCopyButton() {
-    return (
-      <RaisedButton
-        label="Copy Campaign"
-        onTouchTap={async () =>
-          await this.props.mutations.copyCampaign(this.props.params.campaignId)
-        }
-      />
-    );
-  }
-
   render() {
     const { data, params } = this.props;
     const { organizationId, campaignId } = params;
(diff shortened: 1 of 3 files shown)


src/containers/AdminCampaignStats.jsx
```diff
@@ -272,11 +261,26 @@ class AdminCampaignStats extends React.Component {
...
-                          onTouchTap={async () =>
-                            await this.props.mutations.copyCampaign(
+                          onTouchTap={async () => {
+                            let result = await this.props.mutations.copyCampaign(
                               this.props.params.campaignId
-                            )
-                          }
+                            );
+                            if (
+                              window.confirm(
```
schuyler1d: I support adding feedback on-completion but it should not be through `window.confirm()` which forces an interaction.  Many campaigners will click this button 10 times from a campaign template so they can then go edit them all at once -- they shouldn't be interrupted with a popup window -- please add a `setState()` call that updates the render with a SnackBar link to the new campaign.

schuyler1d (changes requested):
This'll be very useful and helpful -- especially the deterministic ordering.

Please tweak the feedback and then we're good to merge!  Thank you!

src/containers/AdminCampaignStats.jsx
tekkamanendless: Okay, the `Snackbar` is in, and I added what I believe are the appropriate tests under `__test__/backend.test.js`.

tekkamanendless:
This is ready to go for the v7 train, but if you want to merge it into v8, I want to add support for:

1. `response_window`

(I think that that's the only new field that would need to be copied.)

PR #1744, opened 2020-08-13
Add more info about safely using Redis with multiple Spoke instances

tekkamanendless:
## Description

This adds more information and clarity to the Redis docs, including:

1. How to safely use Redis with multiple Spoke instances
1. How to do some basic Redis maintenance operations

# Checklist:

### DIFF
--- a/docs/HOWTO_CONNECT_WITH_REDIS.md
+++ b/docs/HOWTO_CONNECT_WITH_REDIS.md
@@ -1,26 +1,74 @@
-Redis is a caching technology that can greatly speed up the application.  We recommend every
-production instance of Spoke to use Redis -- even if it's a very small size -- 'micro' or 'hobby-dev'
-will still have a very positive impact on scaling.
+[Redis](https://redis.io/) is a caching technology that can greatly speed up the application.
+We recommend every production instance of Spoke to use Redis -- even if it's a very small size -- 'micro' or 'hobby-dev' will still have a very positive impact on scaling.
 
-# Instructions for using Redis in Development and Production
-- To add a caching layer to your development and production environments, follow the below instructions to get setup.
+# Instructions For Using Redis in Development and Production
+To add a caching layer to your development and production environments, follow the below instructions to get setup.
 
 ## Development
-- Using Redis in development can be achieved by connecting a local Redis server or using 'fake Redis'. Note that fake Redis is an easy to use simulation of Node Redis. Initial Redis setup is included in `src/server/models/thinky.js`.
+Using Redis in development can be achieved by connecting a local Redis server or using 'fake Redis'.
+Note that fake Redis is an easy to use simulation of Node Redis.
+Initial Redis setup is included in `src/server/models/thinky.js`.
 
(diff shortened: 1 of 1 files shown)


docs/HOWTO_CONNECT_WITH_REDIS.md
```diff
@@ -1,26 +1,74 @@
...
-- If you are running on a scale of ~100K contacts or more per-day, then we recommend increasing your redis instance with memory ~ 1kbytes * (number-of-contacts-per-day) and enabling REDIS_CONTACT_CACHE=1
-- We recommend setting the "maxmemory policy" to "volatile-lru" -- in the off-chance Redis runs out of memory it will discard the oldest cache data first -- the application should handle this event gracefully as long as there's some discard policy to allow new keys to be added.
+1. Point `REDIS_URL` in production environment variables configuration to production Redis instance url. For example `redis://[url]:6379`.
+1. If you are running on a scale of ~100K contacts or more per-day, then we recommend increasing your redis instance with memory ~ 1kbytes * (number-of-contacts-per-day) and enabling REDIS_CONTACT_CACHE=1
+1. We recommend setting the "maxmemory policy" to "volatile-lru" -- in the off-chance Redis runs out of memory it will discard the oldest cache data first -- the application should handle this event gracefully as long as there's some discard policy to allow new keys to be added.
+
+## Running Multiple Spoke Databases On The Same Redis Server
```
schuyler1d, line 29: Maybe also document CACHE_PREFIX environment variable (as another way to set a different prefix per-instance)

schuyler1d (approved):
:+1:

docs/HOWTO_CONNECT_WITH_REDIS.md
tekkamanendless: Honestly, I'd rather not encourage anyone to use it.  Redis will guarantee the prefix when using its native support (and Redis is the only cache option right now), and anything else is just asking for a bug.

There are also no alternative cache implementation options available (for example, `memcache`); if there were, I'd recommend creating a cache wrapper.  For example:

```
function Cache(engine, connectionString, options)
```

```
// Redis
let cache = new Cache("redis", "redis://[...]", { "prefix": process.env.CACHE_PREFIX })
// If a prefix was given in the options, then we will tack on `prefix=[...]` to the connection string to use the native prefix support.
cache.Set("key1", "value1");
let value = cache.Get("key1");
```

```
// Memcache
let cache = new Cache("memcache", "127.0.0.1", { "prefix": process.env.CACHE_PREFIX })
// If a prefix was given in the options, then we will prefix all keys accordingly.
// Note that some Node JS memcache clients have support for this already.
// Otherwise, we could roll our own prefix support in the `get` and `set` functions.
cache.Set("key1", "value1");
let value = cache.Get("key1");
```

----

Anyway, it's your call on what to do with `CACHE_PREFIX`.  My vote is to leave it undocumented until there is a compelling use case for it.

schuyler1d (approved):
thanks.  TIL!

PR #1756, opened 2020-08-18
Make VAN database mode and timeout configurable; send phoneId with people/canvassResponses

lperson:
## Description

- We had previously hard coded database mode to `0` (my voters). But sometimes people want to use my campaign. I added the environment variable `NGP_VAN_DATABASE_MODE` which can be set to 0 for my voters or 1 for my campaign.  If it's not specified, Spoke will default it to 0.
- Timeout for all VAN calls was hardcoded before, in some cases to 5 seconds. Now it defaults to 32 seconds for all calls to the VAN API, and it's configurable with `NGP_VAN_TIMEOUT`.
- Send phoneId in canvassContext in people/canvassResponses call. This is necessary because some updates disqualify the person, rather than the affected phone number, if we don't pass the phoneId. Confirmed with VAN that it's innocuous to always send it. VAN will use it only if applicable.

# Checklist:


schuyler1d (approved):
Thanks Larry

PR #1768, opened 2020-08-22
AWS event/metric telemetry

schuyler1d:
# Telemetry infrastructure esp. for AWS CloudWatch

## Description

Copies Telemetry infrastructure from the Warren campaign.  Thanks to @matteosb 

# Checklist:

### DIFF
--- a/src/server/api/errors.js
+++ b/src/server/api/errors.js
@@ -26,7 +26,11 @@ export async function accessRequired(
   // require a permission at-or-higher than the permission requested
   const hasRole = await cacheableData.user.userHasRole(user, orgId, role);
   if (!hasRole) {
-    throw new GraphQLError("You are not authorized to access that resource.");
+    const error = new GraphQLError(
+      "You are not authorized to access that resource."
+    );
+    error.code = "UNAUTHORIZED";
+    throw error;
   }
 }
 
--- a/src/server/api/mutations/findNewCampaignContact.js
+++ b/src/server/api/mutations/findNewCampaignContact.js
@@ -1,4 +1,5 @@
 import { log } from "../../../lib";
+import telemetry from "../../telemetry";
 import { Assignment, Campaign, r, cacheableData } from "../../models";
 import { assignmentRequiredOrAdminRole } from "../errors";
 import { getDynamicAssignmentBatchPolicies } from "../../../extensions/dynamicassignment-batches";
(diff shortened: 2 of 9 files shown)


src/server/telemetry.js
```diff
@@ -0,0 +1,198 @@
...
+const stage = process.env.STAGE || "local";
+const functionName = process.env.AWS_LAMBDA_FUNCTION_NAME || "NOT_SET";
+
+const reportEventCallbacks = [];
+const reportErrorCallbacks = [];
+const expressMiddlewareCallbacks = [];
+const formatRequestErrorCallbacks = [];
+
+const TELEMETRY_IGNORED_ERROR_CODES = [
+  "FORBIDDEN",
```
matteosb, line 15: Don't need all the Warren error codes

src/server/telemetry.js
```diff
@@ -0,0 +1,198 @@
...
+const makeCloudwatchErrorEvent = (err, details) => {
+  const errDetails = {
+    traceback: err.stack,
+    message: err.message,
+    ...details
+  };
+  return makeCloudwatchEvent("Application Exception", errDetails);
+};
+
+// Specific to the Warren AWS deploy: report a cloudwatch event to "Mission Control"
```
matteosb, line 54: nit: remove this comment

src/server/api/mutations/findNewCampaignContact.js
```diff
@@ -122,6 +123,12 @@ export const findNewCampaignContact = async (
       "assignedCount",
       updatedCount
     );
+
+    await telemetry.reportEvent("Assignment Dynamic", {
```
matteosb, line 127: this seems like it is going to be triggered a lot, imo it would be better as a metric rather than an event

src/server/api/errors.js
```diff
@@ -26,7 +26,11 @@ export async function accessRequired(
   // require a permission at-or-higher than the permission requested
   const hasRole = await cacheableData.user.userHasRole(user, orgId, role);
   if (!hasRole) {
-    throw new GraphQLError("You are not authorized to access that resource.");
+    const error = new GraphQLError(
```
matteosb, line 29: We should throw regular JS `Error`s rather than `GraphQLError`

src/server/telemetry.js
```diff
@@ -0,0 +1,198 @@
+import moment from "moment";
+import AWS from "aws-sdk";
+import { log } from "../lib";
+import _ from "lodash";
+
+const stage = process.env.STAGE || "local";
```
matteosb, line 6: Does MoveOn set the STAGE env var?

matteosb (commented):
Looks good, I would suggest separating metrics and events in a future iteration, with events being reserved for things you want to do something about in real time, e.g. on the Warren campaign we posted every event to slack with a simple Lambda SNS consumer.

PR #1769, opened 2020-08-23
Lp clear org and extension caches

lperson:
## Description

This PR provides a button in the UI and a mutation for administrators to remove the cached organization object from the redis cache, which is necessary when 

* they use a SQL client to modify `organization.features`
* metadata changes in an external resource, such as NGP VAN (the use cases in VAN is that a saved list was added or survey questions or activists code were modified; Spoke's VAN integration won't go back to VAN through the API if those metadata are present in the cache).

Screen Shot 2020-08-24 at 9 52 15 AM

# Checklist:

### DIFF
--- a/src/components/OrganizationFeatureSettings.jsx
+++ b/src/components/OrganizationFeatureSettings.jsx
@@ -1,22 +1,10 @@
 import type from "prop-types";
 import React from "react";
-import orderBy from "lodash/orderBy";
-import Slider from "./Slider";
-import Divider from "material-ui/Divider";
-import AutoComplete from "material-ui/AutoComplete";
-import IconButton from "material-ui/IconButton";
-import RaisedButton from "material-ui/RaisedButton";
 import GSForm from "../components/forms/GSForm";
 import yup from "yup";
 import Form from "react-formal";
-import OrganizationJoinLink from "./OrganizationJoinLink";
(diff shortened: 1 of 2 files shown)


src/server/api/mutations/clearCachedOrgAndExtensionCaches.js
```diff
@@ -0,0 +1,39 @@
...
+import { clearCacheForOrganization as clearActionHandlerCaches } from "../../../extensions/action-handlers";
+import cacheable from "../../../server/models/cacheable_queries";
+import { r } from "../../../server/models";
+
+export const clearCachedOrgAndExtensionCaches = async (
+  _,
+  { organizationId },
+  { user }
+) => {
+  await accessRequired(user, organizationId, "SUPERVOLUNTEER", true);
```
schuyler1d: I think this should be an ADMIN

schuyler1d (dismissed):
going to tweak the permission to be ADMIN, but :+1:

lperson:
@schuyler1d I moved the `clear cached organization and extension caches` into a separate expandable card on the settings panel which will be shown only when `props.params.adminPerms` is true. Why even tempt them with the button if they're a super volunteer?

schuyler1d:
merged into 9.1

PR #1770, opened 2020-08-24
Add importing of response tags

jeffm2001:
## Description

Based off #1705, adds importing of associated tags with canned responses for the google doc importer.

One flag: I'm using _italicized_ text to indicate the tags in the Canned Response section, which is the same formatting used to indicate action handlers in the Interactions section. If we were to ever add a feature for directly associating action handlers with canned responses, that could be an issue. But I'm assuming that won't be the case and this will be fine.

# Checklist:

### DIFF
--- a/docs/HOWTO_IMPORT_GOOGLE_DOCS_SCRIPTS_TO_IMPORT.md
+++ b/docs/HOWTO_IMPORT_GOOGLE_DOCS_SCRIPTS_TO_IMPORT.md
@@ -70,7 +70,7 @@ It's a horrible idea to publish live secrets to Github. You should never do that
 
 ## Create script documents
 
-1. Create a script from a Google Doc Spoke script template - see for example https://docs.google.com/document/d/1zKRbDU9vwetJlfWYgSGd1taBEVj_QrI7bhu4snvHlmg/edit
+1. Create a script from a Google Doc Spoke script template - see for example https://docs.google.com/document/d/1gFji2Vh_0svb4j7VUtmJZkqNhRWt3htp_bdGoWh6S6w/edit
 2. Copy the template, create a new draft doc with that and save it, then edit the new script doc, not the template :)
 3. Share the script document with your API user.
    - Go to the document in Google Docs.
--- a/src/server/api/lib/import-script.js
+++ b/src/server/api/lib/import-script.js
@@ -340,7 +340,7 @@ const makeCannedResponsesList = cannedResponsesParagraphs => {
       if (textParagraph.isParagraphItalic) {
         // Italic = tag.
         const tagId = textParagraph.text.match(/^\d*\b/);
-        if (!!tagId[0]) {
+        if (tagId && !!tagId[0]) {
           cannedResponse.tagIds.push(tagId[0])
         }
       }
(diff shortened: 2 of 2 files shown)


src/server/api/lib/import-script.js
```diff
@@ -336,7 +337,17 @@ const makeCannedResponsesList = cannedResponsesParagraphs => {
       !cannedResponsesParagraphs[0].isParagraphBold
     ) {
       const textParagraph = cannedResponsesParagraphs.shift();
-      cannedResponse.text.push(textParagraph.text);
+      if (textParagraph.isParagraphItalic) {
+        // Italic = tag.
+        const tagId = textParagraph.text.match(/^\d*\b/);
+        if (!!tagId[0]) {
```
schuyler1d: this will fail when there is no match with a type error null can't index [0] == tweak condition to `if (tagId)`

schuyler1d (changes requested):
need at least one tweak on the regex

src/server/api/lib/import-script.js
jeffm2001: Are you sure? When I was testing, `match()` always returns an array.
E.g.: `'foo'.match(/^\d*\b/)`
Returns: `[ '', index: 0, input: 'foo', groups: undefined ]`

jeffm2001: @schuyler1d - I couldn't reproduce the error you mentioned, but I added the extra condition just to be safe. The `!!tagId[0]` needs to be there as well. Without it, when you try to import a script with a non-matching tag, you get an error:
`insert into "tag_canned_response" ("canned_response_id", "tag_id") values ($1, $2), ($3, $4), ($5, $6) - invalid input syntax for type integer: ""`

schuyler1d:
```
$ nvm use
Found '/home/sky/sb/moveon/spoke/.nvmrc' with version <10>
Now using node v10.19.0 (npm v6.13.4)
$ node
> 'asdf'.match(/xxx/)
null
> 'asdf'.match(/asd/)
[ 'asd', index: 0, input: 'asdf', groups: undefined ]
> 'asdf'.match(/as(d)/)
[ 'asd', 'd', index: 0, input: 'asdf', groups: undefined ]
```

in browser: https://jsfiddle.net/3uwj8La1/

docs/HOWTO_IMPORT_GOOGLE_DOCS_SCRIPTS_TO_IMPORT.md
```diff
@@ -70,7 +70,7 @@ It's a horrible idea to publish live secrets to Github. You should never do that
 
 ## Create script documents
 
-1. Create a script from a Google Doc Spoke script template - see for example https://docs.google.com/document/d/1zKRbDU9vwetJlfWYgSGd1taBEVj_QrI7bhu4snvHlmg/edit
+1. Create a script from a Google Doc Spoke script template - see for example https://docs.google.com/document/d/1gFji2Vh_0svb4j7VUtmJZkqNhRWt3htp_bdGoWh6S6w/edit
```
schuyler1d, line 73: Thanks -- this is a great new template!

schuyler1d (approved):

PR #1775, opened 2020-08-24
Allow admins and supervols to see organization settings

lperson:
## Description

Supervols and admins were getting a graphql error when they tried to navigate to the organization | settings panel. That was happening because we weren't handling the case where someone with less privilege than an owner requested TexterUIConfig on the organization object. Now, that doesn't cause an error, and attempting to request it will return null. In that case, `Texter UI Defaults` doesn't render on organization settings.

This PR also restricts the new cache reset button to owners.  I made that change based on feedback from one of the Spoke user organizations.

# Checklist:

### DIFF
--- a/src/server/api/organization.js
+++ b/src/server/api/organization.js
@@ -208,16 +208,17 @@ export const resolvers = {
     texterUIConfig: async (organization, _, { user }) => {
       try {
         await accessRequired(user, organization.id, "OWNER");
-        const options = getConfig("TEXTER_UI_SETTINGS", organization) || null;
-        // note this is global, since we need the set that's globally enabled/allowed to choose from
-        const sideboxChoices = getSideboxChoices();
-        return {
-          options,
-          sideboxChoices
-        };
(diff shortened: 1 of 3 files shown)


src/server/api/organization.js
```diff
@@ -206,14 +206,18 @@ export const resolvers = {
...
+      try {
+        await accessRequired(user, organization.id, "OWNER");
+        const options = getConfig("TEXTER_UI_SETTINGS", organization) || null;
+        // note this is global, since we need the set that's globally enabled/allowed to choose from
+        const sideboxChoices = getSideboxChoices();
+        return {
+          options,
+          sideboxChoices
+        };
+      } catch (caught) {
```
schuyler1d, line 211: can you move this catch up to just cover the accessRequired?

lperson: @schuyler1d I made this change.

schuyler1d:
merged in 9.1

PR #1788, opened 2020-08-27
8.2 PR -- scaling improvements

schuyler1d:
Wut!  We just (stealth) released 8.1 -- why the quick second release?  Well, we deploy 8.1 at MoveOn on production and it was doing great for two days.  On the third day there was a final set of tweaks and we deployed that and we got a product validation from our campaign folks -- and thus we cut the release for 8.1 on Wednesday 8/26.  Well, Murphy's law -- two hours after we finished up the release we started hitting production issues.  We have not yet scaled up for our "hockeystick" (where the participation graph looks like a hockey stick and surges) period and honestly, I asked our campaign manager "try to break Spoke this week" -- well, the texting team sent over 1million messages with 70K sent in a 5minute period and our database started failing.

We will be upgrading this weekend and the great thing about Spoke is it scales "linearly" -- if you double the resources, then you can roughly double the contacts/texters-per-day metrics.

But because of the timing it was ambiguous whether we 'just' hit scaling issues yesterday or whether there were bugs in Spoke 8.1.  I spent the day tracking everything that went wrong and have determined that it was scaling.  I did some urgent improvements to make our queries more efficient and I'll be focusing on more improvements so the application can do its part in keeping scaling costs as low as possible.  But these last set are pretty impactful -- even without upgrading our system with almost the same texting volume as yesterday we are seeing half the database load.

Changes:
* This does include a schema change on the `opt_out` table (please see instructions for migrating)
* Drastically improves the query efficiency for the Texter Todos page
* Removes some liability of thrashing with auto-optout updating.

### DIFF
--- a/docs/HOWTO-use-message-handlers.md
+++ b/docs/HOWTO-use-message-handlers.md
@@ -68,7 +68,7 @@ The best way to make a new message handler is something like this in the codebas
 
 ```
 cd src/extensions/message-handlers
-cp -rp initaltext-guard <NEW_STRATEGY_NAME>
+cp -rp initaltext-guard <NEW_NAME>
 ```
 
 Then edit index.js in that directory.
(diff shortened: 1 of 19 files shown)


ibrand (dismissed):
PR for release notes is out too!

ibrand (approved):
🥳

PR #1789, opened 2020-08-27
Release notes for 9.0

ibrand:
🎉


schuyler1d (dismissed):
++ -- maybe one thing to do is link to the 8.1 release notes and say something like "the community has a great batch of improvements and you should look at 8.1 release notes which have several major features."

I don't want to  'steal the thunder' of 8.1 -- honestly, I'd be fine with just copying the release notes there or at least highlighting a few of the major improvements and copying the "thanks" list in it in 9.0 so folks can see them both and not miss 8.1's changes.

schuyler1d (approved):

PR #1790, opened 2020-08-27
Canned response search

JeremyParker:
# Fixes #1790 
## Description
This change adds a search field to the Survey/Responses popover in the Texter experience. The search field updates the available Survey Answers and Canned Responses as the user enters characters. Search terms are checked against the title, and the response text.

Screen Shot 2020-08-31 at 12 06 41 PM

Screen Shot 2020-08-31 at 12 27 20 PM

# Checklist:

### DIFF
--- a/src/components/AssignmentTexter/Controls.jsx
+++ b/src/components/AssignmentTexter/Controls.jsx
@@ -89,6 +89,13 @@ export class AssignmentTexterContactControls extends React.Component {
     window.addEventListener("orientationchange", this.onResize);
   }
 
+  componentWillUnmount() {
+    document.body.removeEventListener("keyup", this.onKeyUp);
+    document.body.removeEventListener("keypress", this.blockWithCtrl);
+    window.removeEventListener("resize", this.onResize);
+    window.removeEventListener("orientationchange", this.onResize);
+  }
+
   componentWillUpdate(nextProps, nextState) {
     // we refresh sideboxes here because we need to compare previous state
     const newPopups = [];
@@ -103,13 +110,6 @@ export class AssignmentTexterContactControls extends React.Component {
     newPopups.forEach(sb => {
       nextState.sideboxOpens[sb] = (nextState.sideboxOpens[sb] || 0) + 1;
     });
-  }
-
-  componentWillUnmount() {
-    document.body.removeEventListener("keyup", this.onKeyUp);
-    document.body.removeEventListener("keypress", this.blockWithCtrl);
-    window.removeEventListener("resize", this.onResize);
-    window.removeEventListener("orientationchange", this.onResize);
   }
 
   getStartingMessageText() {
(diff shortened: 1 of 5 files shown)


schuyler1d (changes requested):
I would like us only to show search if canned response count > ... 6 or so (however many bump below the popup scrolling on a reasonable-sized device).

Maybe I'm just missing the conditional.

This doesn't have to block this PR, but it might be the right time/place to also do a search on canned response tags -- I'm not sure we're pulling down that data in the texter view yet, but maybe *if it's there* then search through it? -- then we can just do a PR to add the data to the query.

JeremyParker:
@schuyler1d you did not miss the conditional. It wasn't there. It's added in the newest commit.

Also, the "Other Responses" link disappears when the its conditions are no longer met in the filtered popup contents. This way someone can't click on "Other Responses" if there are no other responses in the filtered results.

src/components/AssignmentTexter/Controls.jsx
```diff
@@ -312,29 +344,27 @@ export class AssignmentTexterContactControls extends React.Component {
 
   handleMessageFormChange = ({ messageText }) => this.setState({ messageText });
 
-  handleOpenAnswerPopover = event => {
+  handleOpenAnswerResponsePopover = event => {
```
inorvig, line 346: So I understand this is combining handleOpenAnswerPopover and handleOpenResponsePopover, but I'm not seeing where the response parts were being called before so just unclear on the reasoning/how this changes things.

src/components/AssignmentTexter/Controls.jsx
```diff
@@ -228,6 +236,30 @@ export class AssignmentTexterContactControls extends React.Component {
     }
   };
 
+  handleSearchChange = searchValue => {
+    // filter answerOptions for this step's question
+    const answerOptions = this.state.currentInteractionStep.question
+      .answerOptions;
+    const keysToSearch = ["value", "nextInteractionStep.script"];
```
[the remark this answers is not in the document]
JeremyParker: No reason really. Might be a little more readable that way, but I changed it to the more succinct syntax in 5142a14c21f9ab6c7a5abeda9072d8f928decfc0

src/components/AssignmentTexter/Controls.jsx
JeremyParker: Combining the two popovers' handlers like this should have no behavioral impact - just code simplification. Previously responsesPopover stuff was being handled [here](https://github.com/MoveOnOrg/Spoke/pull/1790/commits/20e6575c129b26b83be43d5d54cb626851acc9e8#diff-891594c99eab8026a02071b140273077L369). Now those state values are combined with the state values for the AnswerPopover in the same `setState()` call.

(Further code simplification is definitely still possible: e.g. we have `answerPopoverOpen ` and `responsePopoverOpen` but we only have one popover, whose parts can't be open independently.)

PR #1796, opened 2020-08-30
Sqs batching and dispatchProcesses improvements

schuyler1d:
This PR improves dispatchProcesses so they are not run sequentially but in-parallel.  It also fixes some bugs in processing messages from an SQS (in AWS) queue.


schuyler1d:
merged in 9.1

PR #1812, opened 2020-09-05
Mobilize event shifter sidebox

agreenspan24:
## Description

Adds a texter sidebox for Mobilize (America) event scheduling. User clicks the button and the MOBILIZE_EVENT_SHIFTER_URL opens up in a Dialog as an iframe, which is meant to be an organization's main event list. If contact has a zip, it is included to filter the mobilize events. If the user provides an event_id column in the CSV, or provides a default event id for the campaign/organization, it adds a tab for the mobilize event in an iframe and prefills the first name, last name, cell, email, and zip into the fields via the query string. Texter can then switch between tabs of the specific event or the general event list.

Screenshots:
- image
- image
- image

# Checklist:


schuyler1d (changes requested):
Can you make the shifter URL another admin form variable rather than a global? -- this will allow it to be set per-organization and through the UI.

Other than that, looks good!

ibrand:
@agreenspan24 would love to get this into our next release and I know this will make some community member's day!

codygordon:
hey @agreenspan24 we're trying this out and getting the following error in Chrome:

**Refused to display 'https://www.mobilize.us/ny-wfp' in a frame because it set 'X-Frame-Options' to 'sameorigin'.**

edit: turns out you need to send a request to mobilize support to enable embedding for each dashboard lol..

agreenspan24:
@codygordon that's so weird, did they change their CORS policy? Maybe we were already approved through another process? 

While you're talking to Mobilize Support, maybe ask if they have an embed version of the feed and events? We just i-framed the whole page, header included.

arena:
This is a great feature! 
It looks like you are implementing this by embedding the event feed html that comes from mobilize as opposed to using the Mobilize api to get the event details. It seems like a straight-forward strategy. How does this work/display on mobile devices?

arena:
Also, I think the Warren campaign had an implementation of this that got open sourced if that's helpful (I'm sure you love helpful suggestions about how you might do work you've already done.. sorry) but I think they were using the API to get it done if that's interesting to you.

agreenspan24:
Both my version and the Warren version which I based it on used iframes to embed the event page/feed. If I remember correctly, it looks fine on mobile: Mobilize's page is fairly responsive, and it just shows up in a popup that's most of the page. I added a width value that's 100% with a max-width of 768px.

PR #1813, opened 2020-09-08
Per-org VAN settings

### LINKED ISSUE #1693: Feature Request: Set per-org NGP_VAN_API_KEY in UI, similar to Twilio
Based on reading code I am pretty sure (but would love confirmation!) that:

--You can have per-organization differing VAN API keys
--The current way to achieve this is to edit the features column in the Organizations table directly
--Unlike the Twilio secret there's no encryption handling

If someone can confirm I can do the DB edit I would really appreciate that, but having it in the UI similar to the Twilio option would be great as well...thank you!

jeffm2001:
# Fixes #1693 

## Description

This PR will add a form for VAN API settings to the organization settings page, and encrypt the API key before storing it in the database. It also attempts to make it easier to add other encrypted configuration in the future, by doing two things:
1. Make the `OrganizationFeatureSettings` component more reusable by taking fields as a prop (so that different fields can be shown in different blocks). It also moves the logic and submit handling fully into the component.
2. Transparently encrypting/decrypting org features variables that end in `_ENCRYPTED`

Also made some minor tweaks to fix a bunch of React warnings/errors that have plagued the Settings page.

# Checklist:

### DIFF
--- a/src/api/organization.js
+++ b/src/api/organization.js
@@ -77,6 +77,7 @@ export const schema = gql`
     twilioAuthToken: String
     twilioMessageServiceSid: String
     fullyConfigured: Boolean
+    vanEnabled: Boolean!
     phoneInventoryEnabled: Boolean!
     campaignPhoneNumbersEnabled: Boolean!
     pendingPhoneNumberJobs: [BuyPhoneNumbersJobRequest]
--- a/src/components/OrganizationFeatureSettings.jsx
+++ b/src/components/OrganizationFeatureSettings.jsx
@@ -8,6 +8,7 @@ import { dataTest } from "../lib/attributes";
 
 const configurableFields = {
   ACTION_HANDLERS: {
+    category: 'defaults',
     ready: false, // TODO: let's wait for better interface
     schema: ({ formValues }) =>
       formValues.settings.actionHandlers
@@ -56,6 +57,7 @@ const configurableFields = {
     }
   },
   ALLOW_SEND_ALL_ENABLED: {
+    category: 'defaults',
     schema: () => yup.boolean(),
     ready: true,
     component: props => {
@@ -96,6 +98,7 @@ const configurableFields = {
     }
   },
   DEFAULT_BATCHSIZE: {
+    category: 'defaults',
     schema: () =>
       yup
         .number()
@@ -119,6 +122,7 @@ const configurableFields = {
     }
   },
   DEFAULT_RESPONSEWINDOW: {
+    category: 'defaults',
     schema: () => yup.number().notRequired(),
     ready: true,
     component: props => {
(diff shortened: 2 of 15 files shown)


mayefsky:
Just in case it resonates--the big thing for my users that the current VAN setup doesn't allow for (other than encrypting the key, which sounds great) is handling VAN's multiple modes (i.e. MyCampaign vs MyVoters) in the UI within a single organization, i.e. as a per-campaign setting. Doesn't sound like it's necessarily in scope here but just wanted to mention it. (I also have a non-awful workaround currently so definitely no need to up the priority on my account.)

jeffm2001:
> Just in case it resonates--the big thing for my users that the current VAN setup doesn't allow for (other than encrypting the key, which sounds great) is handling VAN's multiple modes (i.e. MyCampaign vs MyVoters) in the UI within a single organization, i.e. as a per-campaign setting. Doesn't sound like it's necessarily in scope here but just wanted to mention it. (I also have a non-awful workaround currently so definitely no need to up the priority on my account.)

This adds a setting in the UI to switch between MC/MV for the org. Switching per-campaign is a good idea, but I think it should be a separate PR.

bxjw:
I'd love to see movement on this PR! It will be super helpful for us, as we have organizations on our Spoke instance that swap regularly between My Voters and My Campaign, and also a user base of admins that aren't necessarily that familiar with SQL.

mayefsky:
Hey Bella,
Not sure how helpful this is but I hacked together this functionality last
year for my own fork and have been using it pretty painlessly ever since. I
literally just made copies of the VAN connectors (for contact loader,
action handler and message handler) and added the "|1" and "|0" to the end
of the database-stored key to force usage of MyCampaign and MyVoters
respectively. So the end user result is two additional contact
loader/action handler/message handlers you can enable via environment
variable.
https://github.com/mayefsky/leftapps_spoke/tree/10-0-plus-connectors
I never opened a PR for this because it's bad/repetitive code and not the
right long term solution, and I lack the bandwidth and JavaScript expertise
to actually do the right thing here, I make no guarantees etc. etc. but
wanted to mention it just in case it's helpful.

Frydafly:
thanks @jeffm2001! would love to see a screenshot of how this looks in the UI so that others (maybe @arena ) can also review?

I also see that now there are some merge conflicts that require resolution.

schuyler1d:
Hi, this PR represents the challenges of wanting to deeply support specific vendors like VAN while avoiding complexity for instances that don't have a VAN connection.  I don't think we can merge this specific PR, but I am hoping to accomodate the ability to create an extension that supports this kind of behavior with this (still in-development) PR: https://github.com/MoveOnOrg/Spoke/pull/1987
which includes "extensions/service-managers" -- which allows for both organization and per-campaign interfaces and saveable settings and can include hooks like onCampaignContactsLoad, onCampaignStart, etc.

jeffm2001:
Closing this in favor of #2198

PR #1815, opened 2020-09-08
Hide branched scripts

agreenspan24:
## Description

Users can add the HIDE_BRANCHED_SCRIPTS config to hide all of the interactions workflow. The user will instead select an "Initial Outbound" message and then populate Canned Responses. This is to help simplify the texter's UI, make the conversation flow more freely, and simplify the campaign creation process. 

In total, this:
- Renames some help text
- Hides the interaction/question UI on campaigns
- Hides the question/responses in the texter view
- Makes the message text bar effectively a search bar: shortcuts are always shown and get filtered to what's been typed (case insensitive).

*This last point on filtering shortcuts based on typed text and effectively bypassing the +/- functionality when this mode is enabled should be discussed and perhaps standardized regardless of this mode.*

image
image
image
image
image
image

# Checklist:


ibrand:
I want to note that #1790 has a canned response search and we'll probably merge that in soon, so if you're able to resolve conflicts with that one that would be great.

In this experience, do canned responses record data? Or do you leverage tags that link to canned responses for that?

schuyler1d:
Yeah, I'd prefer search filter connected to the number of canned responses (which I believe #1790 does) rather than this option so it's relevant to both, but I like that this config option can simplify the interface for folks wanting to set it up like this.

agreenspan24:
@ibrand, there's some separate work that I'm adding to allow the actions we use on questions to also work on canned responses. That handles our data sync for canned responses. But the work there is extensive, plus requires new DB changes, so I figured I'd try to get that incorporated after the election. 

I'm happy to use the search bar work in #1790 rather than my own. Any sense of when that will be merged? Should I merge it in now or later? Also, do you want to try and rush this for the next release?

CC: @schuyler1d

src/components/AssignmentTexter/Controls.jsx
```diff
@@ -147,10 +147,18 @@ export class AssignmentTexterContactControls extends React.Component {
...
+      sanitizedText.slice(0, limit) +
+      (sanitizedText.length > limit ? "..." : "")
+    );
+
   toggleContactList = () => {
     this.setState({ contactListOpen: !this.state.contactListOpen }, () => {
       setCookie("assignmentContactListOpen", this.state.contactListOpen, 1);
     });
+
   };
```
crayolakat:
```suggestion
  getShortButtonText = (text, limit) => {
    var sanitizedText = text.replace(/^(\+|\-)/, "");
    return (
      sanitizedText.slice(0, limit) +
      (sanitizedText.length > limit ? "..." : "")
    );
  };

  toggleContactList = () => {
    this.setState({ contactListOpen: !this.state.contactListOpen }, () => {
      setCookie("assignmentContactListOpen", this.state.contactListOpen, 1);
    });
  };
```

src/containers/AdminCampaignStats.jsx
```diff
@@ -401,7 +403,15 @@ class AdminCampaignStats extends React.Component {
...
         </div>
+
+        {global.HIDE_BRANCHED_SCRIPTS ? (
+          ""
+        ) : (
+          <div className={css(styles.header)}>Survey Questions</div>
+        )}
+
         <Typography variant="h5">Survey Questions</Typography>
+
```
crayolakat:
```suggestion
          <Typography variant="h5">Survey Questions</Typography>
        )}
```

crayolakat (changes requested):
Hey @Arique1104 , here is my feedback regarding the merge conflict resolutions done in  commit 6e99e566e8fb870fc1104f61750b67bd231f5022:

`src/components/AssignmentTexter/Controls.jsx`:
The syntax is invalid and causing compilation issues. `getShortButtonText` and `toggleContactList` should be 2 separate functions.

`src/containers/AdminCampaignStats.jsx`:
The line `<Typography variant="h5">Survey Questions</Typography>` should replace `<div className={css(styles.header)}>Survey Questions</div>` instead of being added below it.

Could you review my suggested changes and update as needed?

src/components/AssignmentTexter/Controls.jsx
agreenspan24: Looked good to me!

src/containers/AdminCampaignStats.jsx
agreenspan24: Looked good to me!

crayolakat (approved):

PR #1826, opened 2020-09-09
for backwards compatibility we need to validation when TWILIO_STATUS_CALLBACK_URL

schuyler1d:
9.1 has a regression for TWILIO_MULTI_ORG=1 setup where we are now validating twilio-message-report/

This goes back to not validating but only does so if an additional variable is set TWILIO_VALIDATION=1 (or the callback url)

It also moves twilio calls into the single endpoint -- hoping to get closer to where twilio/nexmo/fakeservice are src/extensions!

This has not been tested yet.


mayefsky (approved):
I tested simply merging this with no ENV change and I seem to be getting the callbacks in a quick test. Thanks!

PR #1828, opened 2020-09-10
Fix org creation

jeffm2001:
## Description

With the new Organization Dashboard, creating new org was still not working with `SUPPRESS_SELF_INVITE` on. The cause was a missing parameter in the `createInvite` mutation that resulted in it not receiving the user, so it couldn't check for `is_superadmin`.

I also changed it to use the `is_valid` value from the invite input (which is now being passed in).

# Checklist:


schuyler1d (approved):
:+1:

PR #1836, opened 2020-09-11
Additional fixes for Twilio validation

jeffm2001:
A few additional fixes related to #1826 

- Messaging services created by Spoke include the org ID
- Status callback url is no longer included in the twilio `sendMessage` function. The callback URL does not vary per-message, so it's better to allow it to use the callback set in the messaging service
- Add the org callback URL to the Setting page

### DIFF
--- a/src/components/DisplayLink.jsx
+++ b/src/components/DisplayLink.jsx
@@ -3,23 +3,39 @@ import React from "react";
 import TextField from "material-ui/TextField";
 import { dataTest } from "../lib/attributes";
 
-const DisplayLink = ({ url, textContent }) => (
+const inlineStyles = {
+  displayLabel: {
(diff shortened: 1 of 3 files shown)


schuyler1d:
Thanks -- I'm not going to target this for merge right now to avoid conflict with incoming https://github.com/MoveOnOrg/Spoke/pull/1931 service-vendor refactors, but I'm going to look during the merge to bring this in at that point,.

engelhartrueben:
Hi @jeffm2001, I'm Ruby, one of the Spoke maintainers at State Voices. We are going to close this due to it becoming stale. Feel free to reach out if this becomes a priority again!

jeffm2001:
I don't even remember what the purpose of this was anymore.

PR #1863, opened 2020-10-05
Run Cypress end-to-end tests in CI/Github Actions

asalant:
## Description

This PR gets the existing Cypress end-to-end tests running as a Github workflow. You can see them running at https://github.com/asalant/Spoke/actions.

There are a few things to discuss and some decisions to make before moving forward. Anywhere below that I say "I recommend" I am looking for agreement or a request to take a different approach. 

## Approach

My approach has been to get the existing Cypress tests running in CI before fixing them. They won't get consistently used or maintained until they run in CI and I figured it would be the tricky part so I started there.

It did turn out to be the tricky part as it required gaining a solid understanding of the startup and configuration of the app under different environments.

This PR includes:

1. Get Cypress integration tests running in CI/Github Actions
1. Fix failing tests so that they are passing in CI
1. Refactor tests to remove smoke test support (see rationale below)
1. Refactor tests to use test factory data methods from unit test suite
1. Remove legacy e2e test suite (Saucelabs & Travis CI)
1. Run integration tests in development with webpack/bundler on-the-fly compilation for better developer experience
1. Rename references to "end-to-end tests" and "cypress tests" to "integration tests" to highlight the concept of integration testing over the specific tool being used

## To Do

Next steps following this PR are:

- [ ] Refactor existing integration tests to be good examples for future tests (e.g. introduce [page objects], introduce assertions based on database state)
- [ ] (nice to have) Start node from within Cypress process to enable stubbing and mocking services on the back end 
- [ ] (nice to have) Cache webpack and bundler assets in CI for faster runs
- [ ] Create first smoke test and smoke test runner

## Integration (end-to-end) v smoke tests

@schuyler1d called out their desire to have the Cypress test suite serve double duty as a smoke test for new installations. I advise against this approach. I think having a smoke test suite is a great idea and that Cypress is an appropriate tool, but I think you need to have full liberty with the database in the integration test suite so that you can set up the system in whatever state you need it in order to run your test, often resetting the database with new data for each test.

Smoke tests, on the other hand, need to be mindful that they are running against a production instance. Done well, the smoke tests will be useful for verifying new installations and for running against production instances after new deployments. 

I recommend creating smoke tests in `__test__/cypress/smoke` and having different Cypress config for running these tests than the end-to-end tests. The integration tests should be free to do anything they want with the database while the smoke tests should have little-to-no impact on the system-under-test. Any impact they do have should be well documented. I'm happy to create a first smoke test to demonstrate and so that the infrastructure is there.

With this approach, there's some simplification of the existing e2e tests available as they are currently written to either create a new test org or use one that exists in the system under test. I believe all e2e tests should be run against data created in the test suite itself, ideally using test data fixtures.

## NODE_ENV and environment switches

I recommend running both the application-under-test and the Cypress test suite with `NODE_ENV=test`. The suite was running against the development instance with `NODE_ENV=development` and the different behaviors that come with that like using the webpack proxy to recompile assets. That config also requires a local `.env` file which does not exist in CI and was actually my biggest issue to resolve when figuring out how to move forward.

Switching to having the application-under-test use `NODE_ENV=test` required two changes to files that run in production so please review those changes closely [here](https://github.com/MoveOnOrg/Spoke/pull/1863/files#diff-727d7d85bce5c6c73597e64340f1c88bR13) and [here](https://github.com/MoveOnOrg/Spoke/pull/1863/files#diff-d878d2e3214b3e9fae02a34123a87635R5). These changes allowed me to start the app with `NODE_ENV=test` and no `.env` file which I needed for CI.

With this configuration, assets can be compiled by webpack for the integration tests just as they are in production or created on the fly for local testing & development (TDD). 

## Running locally

With these changes, to run locally you:

0. (Optional, but necessary to replicate CI environment) Reset database with `docker-compose down -v`
1. Start postgres with `docker-compose up`
1. Create test database with `./dev-tools/create-test-database`
1. Start application-under-test on port 3001 with `yarn start:test`
1. Run cypress test suite with `yarn test-cypress`

The last step opens the Cypress interactive runner. Switching the cypress command in package.json to `cypress run` will run the suite headless as it does in CI.

With this setup you can run both your development server and the server-under-test at the same time as they use different ports and the same postgres instance with two distinct databases and sets of credentials. The server-under-test starts with babel and webpack hot reloading so you can use Cypress to do TDD when developing new features.

## Github workflow

It looks like the path forward for CI for this project is Github actions. Thanks for suggesting - I hadn't used them before and they seem solid. Nice that I can just enable them in my own fork to get CI support out of the box.

When running as a Github action, the workflow precompiles assets before starting the server so the tests run against an instance that starts up like a production server.

It was confusing to see the "passing" badge on the project link to Travis but to hear that Github was the place to look. Once the integration tests are integrated into the main project, I recommend removing the Travis references and pointing everyone at the Github actions.

## Minor details

A few less important questions or things to point out...

*  Script `prod-build` could be renamed to `build` as it doesn't do anything production-specific (e.g. I use it to build assets for the test instance)

# Checklist:


__test__/cypress/plugins/utils.js
```diff
@@ -8,15 +8,14 @@ export async function getOrCreateTestOrganization() {
...
     .first();
-  if (!org) {
-    org = await r
-      .knex("organization")
-      .insert({
-        name: DEFAULT_ORGANIZATION_NAME,
-        uuid: uuid.v4(),
-        features: JSON.stringify({ EXPERIMENTAL_PHONE_INVENTORY: true })
-      })
-      .returning("*");
```
asalant, line 19: This code path was failing with an error every time in CI. `returning(*)` doesn't return an object. It was never being hit in development because these tests were never run against an empty database.

webpack/config.js
```diff
@@ -22,6 +21,11 @@ const jsxLoaders = [{ loader: "babel-loader" }];
 const assetsDir = process.env.ASSETS_DIR;
 const assetMapFile = process.env.ASSETS_MAP_FILE;
 const outputFile = DEBUG ? "[name].js" : "[name].[chunkhash].js";
+console.log("Webpack starting with config", {
+  assetsDir,
+  assetMapFile,
+  outputFile
+});
```
asalant: It was helpful to see this when sorting out config in different environments. I can delete if it seems overly verbose.

__test__/cypress/plugins/index.js
```diff
@@ -1,7 +1,6 @@
 // plugins run in the cypress Node process, not the browser. Define tasks
 // operations like seeding the database that can't be run from the browser.
 // See: https://on.cypress.io/plugins-guide
-require("dotenv").load();
```
asalant, line 4: I scratched my head over this for a while. Is `.env` used by webpack in ways that I don't understand? This line was causing webpack to fail to build in CI where `.env` does not exist. Why doesn't the Heroku Dockerfile build fail in the same way?

schuyler1d (dismissed):
this all looks good.

If env vars aren't coming from the .env file (good that they wouldn't), where do they come from?

asalant:
> If env vars aren't coming from the .env file (good that they wouldn't), where do they come from?

The come from the process environment when there is no .env as they do in production.

webpack/config.js
```diff
@@ -1,9 +1,8 @@
-require("dotenv").config();
```
asalant, line 1: I scratched my head over this for a while. Is .env used by webpack in ways that I don't understand? This line was causing webpack to fail to build in CI where .env does not exist. Why doesn't the Heroku Dockerfile build fail in the same way?

asalant:
@schuyler1d I've spent some more time on these tests this week and I believe am at the point where I am running up against my limited knowledge of Spoke and the development work that has been done since May when @matteosb implemented the Cypress tests currently in the project.

I could use your help to troubleshoot the last failing test. I'm not sure if Spoke has changed since the e2e tests were written or if the test server instance isn't configured quite right. I've done a fair bit of my own debugging but would really benefit from a more experienced set of eyes.

You can see [`__test__/cypress/integration/basic-campaign-e2e.test.js` failing in CI](https://github.com/asalant/Spoke/runs/1254496309?check_suite_focus=true). To reproduce locally, follow the steps in the PR above to run `yarn test-cypress`. When the Cypress UI comes up, click on `basic-campaign-e2e.test.js` to run the test.

The test is failing when the "texter" user clicks on “Send first texts (2)" but then gets a screen that says “You've messaged all your assigned contacts.” instead of the texting UI.

When I reproduce the campaign setup manually in the same Cypress Spoke server instance, I get the expected behavior of seeing the texting UI when I log in as a user with "TEXTER" role.

If you have the time to reproduce and investigate it would be a huge help in unblocking me. I'm happy to find a time to pair in real time to resolve this if you prefer. Would probably go faster since I can catch you up on my understanding of the e2e test setup. Just hit me up on Slack to coordinate.

Once this last test is passing, I'll move on to the cleanup I outline in the PR.

__test__/cypress/integration/basic-campaign-e2e.test.js
```diff
@@ -18,24 +23,36 @@ describe("End-to-end campaign flow", () => {
...
+    // cy.get("input[data-test=dueBy]").focus();
+    cy.get("input[data-test=dueBy]").type(campaignDueBy, { force: true });
+    cy.get("button")
+      .contains("Cancel")
+      .click();
+    cy.wait(200);
+    cy.pause();
+    cy.get("input[data-test=dueBy]").then(input => {
+      input.val(campaignDueBy);
+    });
```
asalant: @schuyler1d If you have a chance, check this out in the Cypress runner. Something is setting the dueBy back to 1/1/2000 after I set it programmatically. This code sets it once by typing into the input before/while the picker displays and then is reset when the picker renders. After the picker is closed, it sets it again by directly accessing the form input and then in reset when the form submit is clicked.

I can't see what may be resetting the value to 1/1/2000. Any ideas?

asalant: Resolved.

matteosb (dismissed):
Looks good to me. Thanks for picking this back up!

asalant:
@schuyler1d I believe we are good to go with this PR. There's more I would like to do but suggest doing in follow on PRs:

- [ ] Remove smoke test support in favor of separate smoke tests
- [ ] Use server unit test fixture data generators for e2e tests
- [ ] Refactor existing e2e tests to be good examples for future tests (e.g. introduce [page objects](https://martinfowler.com/bliki/PageObject.html))
- [ ] Run e2e tests in development with webpack/bundler on-the-fly compilation for better developer experience
- [ ] Start node from within Cypress process to enable stubbing and mocking services on the back end
- [ ] Cache webpack and bundler assets in CI for faster runs

schuyler1d (dismissed):
:+1: Thanks, this is really valuable work!

PR #1867, opened 2020-10-08
fix prepareTwilioCampaign knex .map chain bug

codygordon:
# Fixes

In jobs.prepareTwilioCampaign you get `TypeError: trx(...).select(...).where(...).map is not a function` 

## Description

Chaining `.map` on knex queries seems like a bad practice either way.


schuyler1d:
This should work.  I solved the same thing (I think) with just wrapping the await in parens:
https://github.com/MoveOnOrg/Spoke/commit/ff861eb8bede83e3176d60b90b09f9714180b653

schuyler1d:
(sorry i missed this pr)

PR #1872, opened 2020-10-12
[WIP] Multi org todo screen

### LINKED ISSUE #1871: Feature Request: View all todos across orgs
**Problem**
There are many scenarios where texters are working on 1 Spoke instance but in multiple orgs and then it gets easy for a texter to miss one of their responses.

**Solution**
Create a Texter Todo screen view where a texter sees all assignments across orgs. Start with showing this view when the visiting the todo screen with the query param `?org=all` and as a follow up create a button for exposing it.

ibrand:
# Fixes #1871

## Description
This allows for appending the query param `?org=all` to the Texter Todo screen to get all texter assignments cross org. I'm posting this WIP because I wanted to check in on my proposal here to run a `filter` whenever we display the Todo screen. My understanding is that this page is not one that is particularly slow normally, and that filter will in the majority of cases be pretty fast since most texters shouldn't have a huge number of assignments. That said, I want to be mindful of adding any scaling concerns to the texter side.

This also doesn't have tests yet, but I'm working on those.

# Checklist:

### DIFF
--- a/src/containers/TexterTodoList.jsx
+++ b/src/containers/TexterTodoList.jsx
@@ -17,7 +17,6 @@ class TexterTodoList extends React.Component {
   }
 
   renderTodoList(assignments) {
-    const organizationId = this.props.params.organizationId;
     return assignments
       .sort((x, y) => {
         const xToText = x.unmessagedCount + x.unrepliedCount;
@@ -34,7 +33,7 @@ class TexterTodoList extends React.Component {
         ) {
           return (
             <AssignmentSummary
-              organizationId={organizationId}
+              organizationId={assignment.campaign.organization.id}
               key={assignment.id}
               assignment={assignment}
               texter={this.props.data.user}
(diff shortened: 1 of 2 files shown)


schuyler1d:
Unfortunately, this can be an expensive query -- fine for a single texter, but since texters can come in a 'thundering herd' it can have bad effects on the db/application -- so I think we should at least filter the backend by-default.  I would recommend the following:

* https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/TexterTodoList.jsx#L201 should conditional on ! orgs=all in query
* https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/user.js#L262 should be added conditional on whether organizationId has a value.
* campaign.organization_id is indexed -- This probably should be a separate PR since we don't want to add an index during this period, but we probably want to add an index for `[campaign.is_archived, campaign.is_started]` to campaign.  OTOH, campaign is a pretty small (relative to contacts/messages) table and just adding an index shouldn't be too bad.  So maybe ok to add the migration to this PR.
* Conditional on orgs=all we probably want to indicate the org somewhere in the summary: https://github.com/MoveOnOrg/Spoke/blob/main/src/components/AssignmentSummary.jsx
* This needs to be https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/TexterTodoList.jsx#L37 assignment.campaign.organization.id now

PR #1881, opened 2020-11-06
Delete phone numbers from inventory and Twilio

jeffm2001:
## Description

Adds delete buttons to the phone number inventory page, which lets you bulk remove unallocated numbers for an area code from both the Spoke inventory and you Twilio account.

2020-11-06 at 9 03 AM
2020-11-06 at 9 03 AM
 

# Checklist:

### DIFF
--- a/src/containers/AdminPhoneNumberInventory.js
+++ b/src/containers/AdminPhoneNumberInventory.js
@@ -3,6 +3,7 @@ import React from "react";
 import gql from "graphql-tag";
 
 import ContentAdd from "material-ui/svg-icons/content/add";
+import DeleteIcon from "material-ui/svg-icons/action/delete-forever";
 import DataTables from "material-ui-datatables";
 import Dialog from "material-ui/Dialog";
 import FloatingActionButton from "material-ui/FloatingActionButton";
@@ -13,7 +14,12 @@ import Form from "react-formal";
 import theme from "../styles/theme";
 import { dataTest } from "../lib/attributes";
 import loadData from "./hoc/load-data";
-import { CircularProgress, FlatButton, Toggle } from "material-ui";
+import {
+  CircularProgress,
+  FlatButton,
+  RaisedButton,
+  Toggle
+} from "material-ui";
 
 const inlineStyles = {
   column: {
@@ -50,8 +56,9 @@ class AdminPhoneNumberInventory extends React.Component {
       buyNumbersFormValues: {
         addToOrganizationMessagingService: false
       },
-      sortCol: 'areaCode',
-      sortOrder: 'asc'
+      deleteNumbersDialogOpen: false,
+      sortCol: "areaCode",
+      sortOrder: "asc"
     };
   }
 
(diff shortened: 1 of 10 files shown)


src/containers/AdminPhoneNumberInventory.js
```diff
@@ -138,6 +171,18 @@ class AdminPhoneNumberInventory extends React.Component {
         label: "Available",
         style: inlineStyles.column
       },
+      {
+        key: "deleteButton",
+        label: "",
+        style: inlineStyles.column,
+        render: (columnKey, row) =>
+          this.props.params.ownerPerms ? (
+            <FlatButton
```
agreenspan24, line 180: Why not an `IconButton`?

src/containers/AdminPhoneNumberInventory.js
```diff
@@ -264,6 +309,29 @@ class AdminPhoneNumberInventory extends React.Component {
         >
           {this.renderBuyNumbersForm()}
         </Dialog>
+        <Dialog
+          title="Delete Numbers"
+          modal={false}
```
agreenspan24, line 314: I believe `false` is the default here.

src/workers/jobs.js
```diff
@@ -1201,6 +1201,36 @@ export async function buyPhoneNumbers(job) {
...
 }
 
+export async function deletePhoneNumbers(job) {
+  try {
+    if (!job.organization_id) {
+      throw Error("organization_id is required");
+    }
+    const { areaCode } = JSON.parse(job.payload);
+    if (!areaCode) {
+      throw new Error("areaCode and is required");
```
agreenspan24: Typo?

agreenspan24 (dismissed):
Maybe not a task for this first iteration, but would be nice to be able to control how many numbers are being deleted - add some sort of limit there.

Otherwise looks mostly good!

src/containers/AdminPhoneNumberInventory.js
jeffm2001: I initially tried `IconButton` but it didn't look right to me. No hover state and the tool tip was getting cut off by the row, which I'm sure is fixable but `FlatButton` seemed fine.

src/workers/jobs.js
jeffm2001: yep, good catch

matteosb (dismissed):
:raised_hands:

src/server/api/lib/twilio.js
```diff
@@ -702,6 +702,44 @@ async function addNumbersToMessagingService(
   );
 }
 
+/**
+ * Release a phone number and delete it from the owned_phone_number table
+ */
+async function deleteNumber(twilioInstance, phoneSid, phoneNumber) {
+  const response = await twilioInstance.incomingPhoneNumbers(phoneSid).remove();
+  if (response.error) {
```
schuyler1d: I think my one request here is that instead of throwing if deleting a number fails, we should just count/track the errors here.  A use case that came up for us and at least one other campaign was deleting numbers in Spoke when at least some of them were already deleted from the Twilio console.

Other than that, looks good.

Issue #1824, opened 2020-09-09
Bug: Relying on PHONE_NUMBER_COUNTRY default breaks CSV uploads


mayefsky:
**Describe the bug**
If you haven't set PHONE_NUMBER_COUNTRY your CSV uploads will not assume US numbers. https://moveonorg.github.io/Spoke/#/REFERENCE-environment_variables currently suggests US should work as a default.

**To Reproduce**
1. Don't set PHONE_NUMBER_COUNTRY
2. Upload a valid CSV with US numbers
3. Get a "upload at least one number" error

**Expected behavior**
Either the csv processing should honor the US default if not set in the environment, or the environment variable reference should say that it's a required variable, and various deploy .env templates should include it.

This was also the culprit in https://github.com/MoveOnOrg/Spoke/issues/1653

mayefsky:
Edit: .env.example does, it's just the lambda-env that doesn't

maxkc:
Linking to [#1754](https://github.com/MoveOnOrg/Spoke/issues/1754) too because (I think?) the proposed workflow from @bobbyoderek may break that PR workaround.

Seems like there's some work that could be done around country assignment if anyone wants some help tackling it!

Frydafly:
Checking in -- does PR https://github.com/MoveOnOrg/Spoke/pull/1883 resolve this issue?

maxkc:
Yes, #1883 should fix this 🎉

Frydafly:
Woohoo! Closing this.

Issue #1754, opened 2020-08-17
Bug: Can't buy puerto rican numbers with phone number buying


ibrand:
**Describe the bug**
Currently phone number buying is only guaranteed to work for US numbers. I was specifically trying to purchase Puerto Rican numbers when I saw a problem emerge.

**To Reproduce**
1. Go to the phone numbers tab
2. Click on the plus button to buy numbers and type in a non-us area code
3. See the spinner indicating an attempt to buy but a failure

**Expected behavior**
We should allow the purchasing of all types of phone numbers.

**Additional context**
A few notes here:
- [Switchboard](https://github.com/MoveOnOrg/switchboard-twilio) is also only able to buy US numbers
- Some of the numbers I eventually bought through switchboard hacking were only able to text domestically. Maybe worth eventually adding a toggle for purchasing numbers that can only text internationally?
- [Here is the code for searching for numbers](https://github.com/MoveOnOrg/Spoke/blob/main/src/server/api/lib/twilio.js#L489). One note here is that when I spoke to a Twilio rep, they said that for puerto rican numbers, you can't use the `areaCode` key as part of the [availablePhoneNumber](https://www.twilio.com/docs/phone-numbers/api/availablephonenumberlocal-resource) resource. Instead I ended up needing to use the `contains` like this:
```
  return client
    .availablePhoneNumbers('PR')
    .local.list({
      contains: '<the-actual-digits-of-the-area-code>*******',
      smsEnabled: true,
    })
```
It'd be nice to find a way to write this code so that it is maximally versatile and not full of logic forks.

ibrand:
@hiemanshu have you tried out the phone number buying UI? Have you seen problems trying to buy numbers for other countries?

maxkc:
@ibrand `AvailablePhoneNumberLocal` assumes you're buying in the United States unless told otherwise. Even though Puerto Rico numbers are +1, ISO 3166 alpha-2 assigns it PR.

Using `contains: +1787` would work but then you'd have to look again for `+1939` if 787 isn't available.  It'll probably be easier to just rely on the country code instead.

This should work:
```
client.availablePhoneNumbers('PR')
      .local
      .list({smsEnabled: true})
```

It might be worth reminding users in-app that PR numbers cost more than US and don't support MMS.  Because this is "domestic" traffic based off the country calling code of +1 sending an SMS from US --> PR won't error out but it'll cost more (and sending an MMS will error out).

maxkc:
[...]

PR #1883, opened 2020-11-08
Clarify role of PHONE_NUMBER_CONUNTRY environment variable

mayefsky:
##Description

Existing documentation claims that PHONE_NUMBER_COUNTRY defaults to US; in practice when it is not set, uploads of US numbers fail without the country code.

In https://github.com/MoveOnOrg/Spoke/issues/1824 Max suggested that actually making the behavior default to US might cause a problem. I don't understand why, but this at least brings the documentation in line with current behavior and adds the variable to the lambda example .env (it's already in the other ones) which should prevent this gotcha for most new folks setting up Spoke.

Meta note, this is my first PR so please let me know if I'm doing something off-protocol; thanks!


deploy/lambda-env.json
```diff
@@ -39,5 +39,6 @@
   "ROLLBAR_ACCESS_TOKEN": "set_this_in_production",
   "ROLLBAR_ENDPOINT": "https://api.rollbar.com/api/1/item/",
   "DST_REFERENCE_TIMEZONE": "America/New_York",
-  "TZ": "America/New_York"
+  "TZ": "America/New_York",
+  "PHONE_NUMBER_COUNTRY": "US"
```
ibrand, line 43: Should we be careful about defaulting this to US after what Max said?

schuyler1d (approved):
I would rather us fix this bug (it should default rather than fail), but I agree until we fix it, we should document it.

deploy/lambda-env.json
mayefsky: I think the concern was about changing the behavior when unset, versus a concern about the template? FWIW the other templates all to set it--this was just updating the lambda template to match.

maxkc: I'll write a doc outlining the handful of changes you'd want to make for supporting non-US numbers soon.  [Few ideas are here](https://github.com/MoveOnOrg/Spoke/issues/1754#issuecomment-803264867) if you think of others @mayefsky!

PR #1891, opened 2020-11-18
NGPVAN user guide v0

mayefsky:
This is a basic admin user guide for using @lperson's NGPVAN integration with Spoke.
One outstanding item: I do not know or know how to confidently get at all the specific keys one needs to flush from Redis when one adds or changes their VAN API key. If one of you knows what they are, that would be helpful to add.

# Checklist:

(This change is just documentation)

### DIFF
--- a/docs/HOWTO-use-action-handlers.md
+++ b/docs/HOWTO-use-action-handlers.md
@@ -14,7 +14,7 @@ external systems:
 
 - [ActionKit](HOWTO_INTEGRATE_WITH_ACTIONKIT.md)
 - [Mobile Commons](HOWTO_INTEGRATE_WITH_MOBILE_COMMONS.md)
-- NGPVAN and everyaction ***(coming soon)***
+- [NGPVAN and everyaction](REFERENCE-NGPVAN_user_guide.md)
 - [Revere](HOWTO_INTEGRATE_WITH_REVERE.md)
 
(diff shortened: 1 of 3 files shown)


maxkc:
@bobbyoderek let me know if we can be helpful at all in building out this documentation more!

schuyler1d:
@lperson @Frydafly 
Could one of you take a look at this PR?

Frydafly:
Cool! Something that might make this documentation even more useful is if we could include the json variables needed in order to integrate with MyCampaign in VAN, which is different from MyVoters but still used by some organizations. The below goes within the {} brackets in the "features" column for the corresponding organization in the database.

"NGP_VAN_DATABASE_MODE":1,"MESSAGE_HANDLERS":"ngpvan"

Also, some organizations run into the issue of having too many survey questions in VAN, and then Spoke fails to pull them up in the action handler section. The way we have resolved this is by adding this to features: 

"NGP_VAN_ELECTION_CYCLE_FILTER":"2021"

Filtering to the current year limits the options and fixes this issue.

Frydafly (approved):
This looks great!

Frydafly:
@schuyler1d what do you think about merging this documentation! I think it'll be really helpful. 

(cc: @lperson & @maxkc ?)

PR #1905, opened 2020-12-22
for heroku + pg 8+ need ssl -> rejectunauthorized: true

codygordon:
upgrading to pg8+ breaks heroku PG deployments without the following fix

see: https://help.heroku.com/MDM23G46/why-am-i-getting-an-error-when-i-upgrade-to-pg-8

NOTE: using the simpler method of just turning `ssl: true` in the config object into `ssl: { rejectUnauthorized: false }` doesn't work – the only way i can get it working is putting the ssl key in a `connection` _object_, thus needing to use this `DATABASE_URL` parser package recommended by heroku (edited)

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,6 +142,7 @@
     "passport-local": "^1.0.0",
     "passport-local-authenticate": "^1.2.0",
     "pg": "^8.0.3",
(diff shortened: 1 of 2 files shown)


schuyler1d (approved):
thanks!

PR #1917, opened 2021-01-27
Experimental Sticky Sender Prod

agreenspan24:
Note: should be merged to Main with Skip Messaging Service work: https://github.com/agreenspan24/Spoke/pull/36

## Description

Twilio has a feature called Sticky Sender in its messaging services. It saves a mapping between a from number (Spoke/Twilio) and a recipient's number. This is an essential piece of having a conversation, so all future outgoing messages come from the same number. But it also has the utility of making future texts to the recipient appear like they're coming from the same number even after the conversation is over. Many campaigns want this utility to make their relationship with voters/supporters feel more personal. 

Because this is on the messaging service level, any organization in Spoke using only one messaging service has always had the ability to have sticky sender. Sticky Sender is turned on by default. However, this has its limits. For example, messaging services can only contain 400 phone numbers. It is also recommended by Twilio and Spoke that one from number only be used to send messages to 200 recipients per day, which is a rough guideline to avoid carrier spam filtering. This caps the amount of daily unique recipients to roughly 80,000 for a given messaging service. 

The new campaign phone numbers feature adds the ability to text more than 80,000 per day by creating new messaging services for each text campaign. However, you lose the Sticky Sender ability with this framework because the sticky sender mappings are contained within the Twilio messaging service that is being created and destroyed at the beginning and end of the campaign. This means the long-term relationship with the same phone number is lost.

This features adds a simple ability to turn on sticky sender more directly using the `EXPERIMENTAL_STICKY_SENDER` config variable. The approach is relatively simple: the mappings between from and to phone numbers get saved per organization in the `organization_contact` table. Mappings are added whenever a new recipient is texted. Once the mapping exists, we skip the messaging service entirely and send the message while specifying the `from` number. 

The main challenge of this work, besides creating and storing the mappings, was handling incoming messages. Incoming messages don't contain a reference to the message they are responding to. Spoke previously used the combination of the recipient number and the message service sid to find the last message the recipient received. Now, without the message service id being used, we rely on the user number to query the message table. In keeping with that, we needed to add the user number in various other places.

# Checklist:


schuyler1d:
Hey @agreenspan24 this looks like a good start.  Just wanted to checkin and see if you were looking for feedback at this stage or are planning to work more on this before we should review.

agreenspan24:
@schuyler1d I'm marking this as ready for review/feedback. 

I do still intend to write some unit tests to capture some of the logic, though.

Frydafly:
Making a note here that this is blocked by https://github.com/MoveOnOrg/Spoke/pull/1931

PR #1929, opened 2021-02-18
No yellow badges

### LINKED ISSUE #1874: Bug: Yellow badges on buckets of handled conversations prompt unexpected opt-outs by new volunteers
**Problem**
Volunteer texters see the yellow badges on buckets of handled conversations and want to get rid of the badges so they opt people out to “clear” those conversations leading to lots of admin clean-up.

**To Reproduce**
This is a design issue --- the software is working as expected but the humans are confused. 

**Solution**
1. remove the yellow badges
2. reword the labels to complete the sentence “I want to ____” with: “view past” and “view skipped” 
3. either ignore the count or insert it into the label of the button (e.g. “view 2 skipped” or “view 22 handled”)

**Screenshots**
https://www.figma.com/file/mfDrcffzFnb0Iwsjq6Srdz/We-don-t-need-yellow-badges?node-id=0%3A1

**Additional context**
"We’re seeing a pattern of people going back into Past Messages and Skipped Messages and trying to get rid of them somehow (in some cases by opting people out :sob: which creates a ton of clean-up work).
Is there any reason to have an alert badge on these buttons? I can’t think of one and would like to see those yellow circles with a number in them go away." -Dinah, Admin

stefanhayden:
# Fixes #1874

## Description

Volunteer texters see the yellow badges on buckets of handled conversations and want to get rid of the badges so they opt people out to “clear” those conversations leading to lots of admin clean-up.

This PR removes those yellow badge icons and inlines the number. This make it so folks don't think they are a number that needs to be addressed.

image

# Checklist:

### DIFF
--- a/__test__/components/AssignmentSummary.test.js
+++ b/__test__/components/AssignmentSummary.test.js
@@ -170,16 +170,20 @@ describe("AssignmentSummary actions inUSA and NOT AllowSendAll", () => {
     const actions = create(0, 0, 0, 9, 0, false);
     expect(
       actions
-        .find(Badge)
+        .find(RaisedButton)
         .at(0)
-        .prop("badgeContent")
-    ).toBe(9);
+        .prop("label")
+    ).toBe("Past 9 Messages");
+  });
+
+  it('renders "skipped messages (n)" with messaged', () => {
+    const actions = create(0, 0, 0, 0, 8, false);
     expect(
       actions
         .find(RaisedButton)
         .at(0)
         .prop("label")
-    ).toBe("Past Messages");
+    ).toBe("Skipped 8 Messages");
   });
 });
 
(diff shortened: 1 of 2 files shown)


stefanhayden:
# Question! 
the `renderBadgedButton` function seems to let the counts be passed in as `undefined`. Is that on purpose? It would seem that if we want to hide buttons on `0` then we would also want to hide them on `undefined`? I also ask this because in the `spec` file we seem to not pass some counts causing some to be `undefined` and then the test seem to check if those buttons that are supposed to be hidden if `0` count actually exist?

I would want to / expect 0 and undefined to be treated basically the same way unless we are doing it this way on purpose.

schuyler1d (approved):
I'm not sure about the `undefined` use case, so I'm tempted to leave it be as-is.  I know I was doing some work a while ago about sometimes not getting counts and just getting 'some exist vs. none exist' but I don't think that's what the undefined thing is about.

PR #1930, opened 2021-02-19
Upgrade react and react-formal

### LINKED ISSUE #479: Remove dependences that use React.PropTypes

stefanhayden:
## Description

Our dependancies are way out of date. Some repos have been deprecated. Many have had many upgrades but have no CHANGELOG to guide how to upgrade.

Types of updates made:
- react is a lot noisier about passing props that are not supported and so I've tried to only pass supported props where possible.
- react-formal has a new way that inputs are passed in. No longer can we get a global type map and instead we need to use the `as` prop to pass in the component like `<Form.Field as={GSTextField}`
- `react-tap-event-plugin` is deprecated and is said to no longer be an issue in browsers. So all `onTouchTap` are changed to `onClick`
- yup has a new import pattern `import * as yup from 'yup'`
- react expects `key` to be on components in arrays
- react update stoped `react-chartjs` from working and it has been deprecated so I moved to `react-chartjs-2` Updated `chart.js` requires a new data format passed in.
- `Form.Context` no longer exists as used on `ScriptList.jsx` so components have been moved around. Its seems to function correctly now but react throws an error for having a form inside a form. The design might need to be altered to get rid of that error in the future.

Package Upgrades:
- react `^15.6.1` -> `16.14.0`
- react-dom `^15.6.1` -> `16.14.0`
- react-formal `^0.20.0` -> `2.2.2`
- yup `^0.24.0` -> `0.32.3`
- enzyme `^3.3.0` -> `^3.11.0`
- react-formal `^0.20.0` -> `2.2.2`
- enzyme-adapter-react-15 `^1.0.5` -> `DELETED`
- enzyme-adapter-react-16 `^1.15.6` 
- react-hot-loader `^1.3.0` -> `4.13.0`
- react-tap-event-plugin `^2.0.0` -> `DELETED`
- chart.js `^1.1.1` -> `^2.9.4`
- react-chartjs `^0.7.3` -> `DELETED`
- react-chartjs-2 `^2.11.1`
- uglifyjs-webpack-plugin `^0.4.6` -> `1.3.0`

# Checklist:


__test__/components/IncomingMessageList/ConversationPreviewModal.test.js
```diff
@@ -25,7 +25,7 @@ import {
   createStartedCampaign
 } from "../../test_helpers";
 
-describe("ConversationPreviewModal", async () => {
+describe.skip("ConversationPreviewModal", async () => {
```
stefanhayden, line 28: currently throws this error which I don't know how to fix:
```
Invariant Violation: Could not find "client" in the context of ApolloConsumer. Wrap the root component in an <ApolloProvider>.

      93 |       onRequestCloseMock = jest.fn();
      94 |
    > 95 |       component = mount(
         |                   ^
      96 |         <ApolloProvider client={ApolloClientSingleton}>
      97 |           <MuiThemeProvider>
      98 |             <ConversationPreviewModal

      at new InvariantError (node_modules/ts-invariant/lib/invariant.esm.js:12:28)
      at invariant (node_modules/ts-invariant/lib/invariant.esm.js:24:15)
      at node_modules/@apollo/react-common/lib/react-common.cjs.js:55:132
      at updateContextConsumer (node_modules/react-dom/cjs/react-dom.development.js:18304:19)
      at beginWork (node_modules/react-dom/cjs/react-dom.development.js:18661:14)
```
The error originates in the `MessageResponse` component at the line:
```
export default loadData({ mutations })(MessageResponse);
``` 

removing the `loadData` call like this lets the test pass: 
```
export default MessageResponse
```

But I don't know what the issue is with loadData or how to solve it ATM.

stefanhayden:
The updated `react-chartjs-2` wrapper to `chart.js` library has a built in way to show labels.

image

schuyler1d:
This is awesome work! Thank you!  Running it locally, I haven't found any issues at all.
In trying to deploy it, I'm running into a bit of trouble.
First I needed to change `.babelrc` to:
```
{
  "presets": ["react", "env", "stage-0"],
  "only": ["*.js", "*.jsx"],
  "plugins": [
    ["transform-runtime", {
      "polyfill": false,
      "regenerator": true
    }]
  ],
  "env": {
    "dev": {
      "plugins":["react-hot-loader/babel"]
    }
  }
}
```
because hot-loader isn't available in dev dependencies.

Second, I'm getting this (or similar) [Invalid Assignment error](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/187)  when building the client (running `ASSETS_DIR=./build/client/assets ASSETS_MAP_FILE=assets.json ./node_modules/.bin/webpack --config ./webpack/config.js`) -- I'm going to look at these closer over the next day or so, but if you get to it before me, then that would be very welcome.

stefanhayden:
I was able to get `ASSETS_DIR=./build/client/assets ASSETS_MAP_FILE=assets.json ./node_modules/.bin/webpack --config ./webpack/config.js` to work by upgrading `uglifyjs-webpack-plugin` to `1.3.0`. After `1.3.0` (which is `2.0.0`) they require webpack 4.

I also added your `.babelrc` config change.

schuyler1d:
ok, getting closer!  With these additional tweaks (please pull this in) https://github.com/MoveOnOrg/Spoke/commit/0850841bb9cd01b74613c912a2c65b799b8baff1 I was able to deploy.  Clicking around most things worked -- I will ask folks to QA further now, but I did encounter one issue myself:

On the Settings page for the org, I got this [minified React Error](https://reactjs.org/docs/error-decoder.html/?invariant=130&args[]=undefined&args[]=). I was unable to reproduce in my local dev so far, but maybe it's worth noting that THERE ARE SO FEW ERRORS/WARNINGS (WOW!!!) that used to plague the material UI component renders, but am seeing a couple of "componentWillMount has been renamed" and "componentWillReceiveProps has been renamed" on that page -- that's probably a red herring but thought I'd mention it just in case.

schuyler1d:
Another issue we've located and is reproducible in development environment is manually adding a texter (non-dynamic assignment) -- I can add the texter, but can't enter a number into the new field.  I don't see any console errors.
spoke-manualassign

ibrand:
> Another issue we've located and is reproducible in development environment is manually adding a texter (non-dynamic assignment) -- I can add the texter, but can't enter a number into the new field. I don't see any console errors.
> spoke-manualassign

Noting that we just pushed a fix for this one, but haven't debugged the settings page crashing yet!

schuyler1d:
> Another issue we've located and is reproducible in development environment is manually adding a texter (non-dynamic assignment) -- I can add the texter, but can't enter a number into the new field. I don't see any console errors.
> spoke-manualassign

This was resolved with this commit: https://github.com/MoveOnOrg/Spoke/commit/9237c9a1dbb45938191dbaac1a4f4722d895de9f
namely just removing the `mapToValue` method which is no longer needed in the new versions.

So we still need to fix the cypress test and Settings page.

stefanhayden:
not sure how to best coordinate this between this branch I have from my fork and the local copy you have made. I pulled in your 2 commit back in to this PR. 

I see the crash on `spoke-stage`. Seems that error is about importing modules incorrectly (https://stackoverflow.com/questions/42813342/react-createelement-type-is-invalid-expected-a-string) but I'm unclear what is the difference between prod and dev. I'm trying hard to find a way to run the app locally in prod mode but not getting too far. `webpack/congif.js` seems to suggest `DEBUG` could be `false` and minimize the app but I can't see to get that to work.

schuyler1d:
> not sure how to best coordinate this between this branch I have from my fork and the local copy you have made. I pulled in your 2 commit back in to this PR.
> 
> I see the crash on `spoke-stage`. Seems that error is about importing modules incorrectly (https://stackoverflow.com/questions/42813342/react-createelement-type-is-invalid-expected-a-string) but I'm unclear what is the difference between prod and dev. I'm trying hard to find a way to run the app locally in prod mode but not getting too far. `webpack/congif.js` seems to suggest `DEBUG` could be `false` and minimize the app but I can't see to get that to work.

I think you merging our branch works -- if we have more we could make PRs against your repo/branch explicitly if you prefer that for the future -- yours is still the canonical branch.

I think just remove DEBUG= in your .env altogether or set it to empty (setting to `false` will evaluate true because it's a non-empty string).  Not sure how much that helps simulate the exact prod problem -- I also haven't found a way to fully simulate prod environment in dev -- when we figure it out, we should document that!

stefanhayden:
blarg. I figured out how to run the app locally with minimized js but I still didn't get the settings page to crash like it does on staging. 

To run minimized js locally first you need to build the app: 

```
ASSETS_DIR=./build/client/assets ASSETS_MAP_FILE=assets.json  yarn prod-build
```

then to run the app you need a way to run `yarn start` but get it to include your `.env` file. I didn't think of a smart way to do that so instead I just put everything in the `.env` file on one line and add it before the `yarn start` command like this: 
```
NODE_ENV=production SUPPRESS_SELF_INVITE=1 JOBS_SAME_PROCESS=1 AWS_ACCESS_KEY_ID= AWS_SECRET_ACCESS_KEY= AWS_S3_BUCKET_NAME= APOLLO_OPTICS_KEY= DEV_APP_PORT=8090 OUTPUT_DIR=./build ASSETS_DIR=./build/client/assets ASSETS_MAP_FILE=assets.json CAMPAIGN_ID='campaign-id-hash' DB_HOST=127.0.0.1 DB_PORT=5432 DB_NAME=spokedev DB_USER=spoke DB_PASSWORD=spoke DB_TYPE= DB_MIN_POOL=2 DB_MAX_POOL=10 DB_USE_SSL=false DB_DEBUG=false WEBPACK_HOST=localhost WEBPACK_PORT=3000 AUTH0_DOMAIN=XXXX.auth0.com AUTH0_CLIENT_ID= AUTH0_CLIENT_SECRET= BASE_URL=http://localhost:3001 SESSION_SECRET=set_this_in_production DEFAULT_SERVICE=fakeservice NEXMO_API_KEY= NEXMO_API_SECRET= TWILIO_ACCOUNT_SID= TWILIO_AUTH_TOKEN= TWILIO_MESSAGE_CALLBACK_URL= TWILIO_MESSAGE_SERVICE_SID= EXPERIMENTAL_TWILIO_PER_CAMPAIGN_MESSAGING_SERVICE=false TWILIO_STATUS_CALLBACK_URL= TWILIO_SQS_QUEUE_URL= PHONE_NUMBER_COUNTRY=US ROLLBAR_CLIENT_TOKEN= ROLLBAR_ACCESS_TOKEN= ROLLBAR_ENDPOINT=https://api.rollbar.com/api/1/item/ ALLOW_SEND_ALL=false EMAIL_HOST= EMAIL_HOST_PASSWORD= EMAIL_HOST_USER= EMAIL_HOST_PORT= EMAIL_FROM= TWILIO_MESSAGE_VALIDITY_PERIOD= DST_REFERENCE_TIMEZONE='US/Eastern' PASSPORT_STRATEGY=local TEXTER_SIDEBOXES=celebration-gif,default-dynamicassignment,default-releasecontacts,contact-reference,tag-contact,freshworks-widget,default-editinitial,take-conversations,hide-media,texter-feedback OWNER_CONFIGURABLE=ALL yarn start
```

and then the app runs at `http://localhost:8090/`

Madness? Yes!

perhaps an `.env` var I don't have set would cause something additional to be imported? I'll keep trying and reporting back.

stefanhayden:
Crash on the setting page is fixed! 

It was just a bad `react-formal` button. `<Form.Button>` no longer exists and only `<Form.Submit>` works.

To get it to crash I had to have `TWILIO_MULTI_ORG` turned on.

stefanhayden:
~it looks like the cypress tests are passing now too?~

scratch that. I got over excited and was looking in the wrong place.

schuyler1d:
VERY EXCITING!!  I've deployed your changes and can confirm that the Settings page is working now.

schuyler1d (dismissed):
Amazing work!

schuyler1d:
@stefanhayden we found one more issue with the auto-completes in CampaignTextingHoursForm.jsx -- the auto-completes don't work and the times don't show am/pm but just accept absolute number text values. Similarly setting Texting Hours in (organization) Settings only shows number values without autocomplete.

If I am able to work on this, I will update this PR, but if you have a chance to check it out, that would be wonderful.

stefanhayden:
@schuyler1d Should be fixed now! Thanks for the amazing help with bug reports.

schuyler1d:
@stefanhayden That change didn't seem to fix it when I was testing -- current values weren't being shown and it didn't seem to trigger a save.  After some futzing, I think there's some real issue with AutoComplete component mixed with the react-formal update when the labels differ from the values. So I have a commit that switches them to fixed menus here:
https://github.com/MoveOnOrg/Spoke/commit/2146886c41337371c60cae8d3627faa2ec02367b

If that works for you, then I can merge that in. Otherwise, if you want to try getting it to work with the autocomplete, lmk and you can try your hand at it.

stefanhayden:
@schuyler1d hm. I'm confused. I double checked what I had and it seemed to work for me. Here is a gif of what I see:

texting hours working


When I pulled your code down just opening the section throws this error: `Uncaught TypeError: this.addFormField is not a function`. But when I rename `addAutocompleteFormField` to `addFormField` then I see it also working but with the dropdown covering the page: 

image

so I pulled that in.

schuyler1d:
> @schuyler1d hm. I'm confused. I double checked what I had and it seemed to work for me. Here is a gif of what I see:
It's not the timezone selection but if you choose 'override texting hours' and then the start/end-time hour autocompletes that fail

One more issue we've discovered (Haven't had time to investigate)
* Clicked 'EDIT' button from the People page fails.  I think this might be related to a very weird pattern using the apolloClient singleton in UserEdit: https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/UserEdit.jsx#L59-L77  I think these should be moved down to queries/mutations maybe with some specific apollo policy not to load unless asked. -- definitely related to the `loadData` in UserEdit wrapping the component.

stefanhayden:
hmmm. I'm not an expert is setting up graphql. It seems like `useLazyQuery` would let you call it conditionally but required hooks which our version of react-apollo doesn't have. I can try and upgrade it but I'm worried about causing more problems them I fix.

stefanhayden:
On further investigation is seems that just wrapping that UserEdit in the <Dialog> is what cause it to die. if I remove it from the dialog it works. But I can't yet figure out what about the dialog kills it.

In my branch where I'm upgrading to material-ui 4 the new dialog doesn't have this problem. https://github.com/stefanhayden/Spoke/pull/1

stefanhayden:
@schuyler1d Fixed the edit button not working. seems like any place with that kind of graphql error we can fix by just wrapping it with the provider again. It seems the old material ui dialog just does something to not pass the context along.

schuyler1d:
ah ha!  I pushed your change into load-data.jsx so we don't need to worry about it within components.
https://github.com/MoveOnOrg/Spoke/commit/06cd6e9b4098f1c0854fce43564fc36d4123597e

schuyler1d:
This was merged (with some additional changes) into Spoke 10.1 release.  Post-release a regression was reported in not being able to choose the rowsize in Campaigns and Message Review DataTables -- seems to be a bug/issue with material-ui/DropDownMenu.  Reporting this here for tracking and for testing when e.g. material-ui is updated.

stefanhayden:
thanks!

schuyler1d:
And this was merged so I'm going to close!  This was AMAZING work that's really helping us remove some technical debt. Thank you!!!!! (and onward to material-ui!)

PR #1932, opened 2021-02-22
Return informative errors from GraphQL endpoint

lperson:
## Description

I noticed while working on a web component that all errors coming back from the graphql endpoint were empty objects. When an error was thrown in a query or mutation, [this code](https://github.com/lperson/Spoke/blob/a4cc5a19998f87e88377e48573784950f5e22a61/src/network/apollo-client-singleton.js#L17-L26) would be called with `graphQLErrors` equal to an array consisting of an empty object -- `[{}]`. The error thrown in the client would be simply `error Error: GraphQL error: undefined` which doesn't let us give the user any useful information.

This PR has two commits. The first commit could be a solution on its own depending on what we think should be disclosed in the web app.

The [first commit](https://github.com/MoveOnOrg/Spoke/pull/1932/commits/c2a5117e9ecd8a1df1cc32bf4adb9b430f999328) improves the situation a little, resulting in an error with this message: `error Error: GraphQL error: Internal server error`.

The [second commit](https://github.com/MoveOnOrg/Spoke/pull/1932/commits/a4cc5a19998f87e88377e48573784950f5e22a61) returns this (as an example) `error Error: GraphQL error: Invalid Twilio credentials`.

Also, I believe [this is dead code](https://github.com/lperson/Spoke/blob/a4cc5a19998f87e88377e48573784950f5e22a61/src/network/errors.js#L1-L35)  now.

# Checklist:

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -3,7 +3,7 @@ import bodyParser from "body-parser";
 import express from "express";
 import appRenderer from "./middleware/app-renderer";
 import { graphqlExpress, graphiqlExpress } from "apollo-server-express";
-import { makeExecutableSchema, addMockFunctionsToSchema } from "graphql-tools";
+import { makeExecutableSchema } from "graphql-tools";
 // ORDERING: ./models import must be imported above ./api to help circular imports
 import { createLoaders, createTablesIfNecessary, r } from "./models";
 import { resolvers } from "./api/schema";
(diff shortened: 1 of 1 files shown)


schuyler1d (changes requested):
The point of the logic there is to restrict what kind of error is revealed.  If you want to make changes, then do so up here:
https://github.com/MoveOnOrg/Spoke/pull/1932/files#diff-846e308afe343fa0238d38943beed033bec2ab41582b904029a4ef0106d634a1R194-R196
and enable SHOW_SERVER_ERROR=1 or DEBUG=1 on your instance

lperson:
https://github.com/MoveOnOrg/Spoke/pull/1932/files#diff-846e308afe343fa0238d38943beed033bec2ab41582b904029a4ef0106d634a1L196

Do you agree that line should be a GraphQLError, not Error.

schuyler1d:
Yes!

schuyler1d (approved):
:+1:

PR #1935, opened 2021-02-27
Fix #1910: Bump Auth0 Lock to v11.28

### LINKED ISSUE #1910: The login fields and log in button should all be visible on the auth screen without scrolling on mobile devices
**Problem**
This is a pretty minor problem that may cause a minor amount of confusion for users on mobile devices. However, it would set a nice first impression (and eliminate any minor confusion) to make the "LOG IN" button appear closer to the login fields so that users on mobile devices do not need to scroll down to get to the button. 

**Solution**
a. keep the same design but make the need to scroll go away
b. change the layout so the "LOG IN" button is free standing and right below the email and password fields (instead of glued to the very bottom of the view like it is now).

**Screenshot**
This screenshot was taken on a mobile device. Note: the issue is not apparent if you select a mobile device emulator in a desktop browser, but when the mobile browser chrome is added to the screen (or something) the LOG IN button ends up below the fold. 
image

marzvrover:
# Fixes # (issue)
Fixes #1910

## Description

After failing to duplicate this error in a dev environment, I noticed a production environment using Auth0 has this issue. I found the issue is with https://github.com/auth0/lock/issues/1935.

By updating the version number in the documentation file, future configurations should no longer have this issue. There is [a comment on the Auth0/lock issue](https://github.com/auth0/lock/issues/1935#issuecomment-709349436) indicating that older Android phones may still run into this bug, but most devices shouldn't have this bug.

To fix this issue in existing deployments, one should follow the docs/HOWTO-configure-auth0.md guide to customize the login page and bump the version.

# Checklist:
_I do not believe the checklist is applicable for this case._

### DIFF
--- a/docs/HOWTO-configure-auth0.md
+++ b/docs/HOWTO-configure-auth0.md
@@ -51,7 +51,7 @@ callback(null, user, context);
       <script src="https://cdn.auth0.com/js/base64.js"></script>
       <script src="https://cdn.auth0.com/js/es5-shim.min.js"></script>
       <![endif]-->
-      <script src="https://cdn.auth0.com/js/lock/11.11/lock.min.js"></script>
+      <script src="https://cdn.auth0.com/js/lock/11.28/lock.min.js"></script>
       <script>
         // Decode utf8 characters properly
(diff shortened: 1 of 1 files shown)


schuyler1d (approved):
:+1:

PR #1941, opened 2021-03-08
Create HOWTO_setup_twilio_amazon_SQS

### LINKED ISSUE #384: Document TWILIO_SQS_QUEUE_URL setup
Start with a link to:
https://www.twilio.com/blog/2017/07/handling-high-volume-inbound-sms-and-webhooks-with-twilio-functions-and-amazon-sqs.html

* Include warning that TWILIO_SQS_QUEUE_URL should be the 'https://sqs.CHANGEME.amazonaws.com/CHANGEME` url (not the twilio function)
* Include instructions that will work when spoke is hosted out of amazon (e.g. settingup AWS creds in the app)

Include our twilio function source here (must change the line `const INCOMING_SMS_URL = 'https://sqs.CHANGEME.amazonaws.com/CHANGEME` and maybe `region: 'us-west-1' `):
```
/* global exports, require, console, process, Twilio */
'use strict'
 
// Some Node.js modules are preinstalled in the system environment.
// As of this writing, the third party modules are not configureable, but
// they should be soon. For now, though, you can take advantage of
// the AWS SDK being preinstalled. Require and initialize it here with the
// IAM credentials in your system environment.
const AWS = require('aws-sdk')
AWS.config.update({
  accessKeyId: process.env.AWS_KEY,
  secretAccessKey: process.env.AWS_SECRET
})
 
// Get a handle to SQS in the AWS region you created it in
let sqs = new AWS.SQS({ region: 'us-west-1' })
 
// Define SQS queue URLs from your AWS account
const INCOMING_SMS_URL = 'https://sqs.CHANGEME.amazonaws.com/CHANGEME
// const STATUS_CALLBACK_URL = 'https://sqs.us-east-2.amazonaws.com/XXX/status-callbacks'
 
// Implement handler function for incoming messages and status callbacks
exports.handler = function(context, event, callback) {
  // SQS send params - assume it's a status callback to a standard queue
  let sendParameters = {
    MessageBody: JSON.stringify(event)
    //QueueUrl: STATUS_CALLBACK_URL
  }
  
  // If the MessageStatus parameter is not passed, this is an incoming SMS
  // message - add appropriate SQS parameters and change the URL
  if (!event.MessageStatus) {
      sendParameters.QueueUrl = INCOMING_SMS_URL
      // FIFO queues use the message group ID to return messages to consumers
      // in logical groups. For an SMS app, a good group ID is the recipient
      // phone number
      // sendParameters.MessageGroupId = event.From.replace(/\D/g,'')


    // Add the message to the appropriate SQS queue
    sqs.sendMessage(sendParameters, (err, res) => {
      // For now, we'll just log any errors SQS throws us
      console.log(err)
      console.log(res)

      // Send a TwiML response with no reply message - we'll handle 
      // any responses from our workers
      let twiml = new Twilio.twiml.MessagingResponse()
      callback(null, twiml)
    })
  }
}
```

Frydafly:
# Fixes #  384

Documentation based off of @schuyler1d 's write up in this issue https://github.com/MoveOnOrg/Spoke/issues/384

## Description

Feedback / edits welcome since I am not personally familiar with the steps / process outlined

# Checklist:

Not applicable.


schuyler1d:
I think we need an extra step documented here which may be a bit of a blocker since there might be a current bug turning this on, but basically we need to ask them to ensure they also have a recurring Lambda (technically CloudWatch) trigger enabled to reguarly look for incoming messages:
https://github.com/MoveOnOrg/Spoke/blob/main/docs/HOWTO_DEPLOYING_AWS_LAMBDA.md#setting-up-scheduled-jobs
Because of the bug, I'm almost tempted to say we should update the docs to say DO NOT create the scheduled job now :-(

It also might be worth adding that instead of making this the primary means of receiving messages, that you can also set the Twilio messaging service *fallback* to the SQS queue which then means that if your Spoke application is down/not-responding, then Spoke doesn't lose messages and they are added to the SQS queue.  In this scenario (which MoveOn does), we do NOT automate it, but when service is restored, we can manually run the trigger and load in the sqs messages from Twilio without having lost any incoming messages.

Frydafly:
@schuyler1d added more commits to the `Frydafly-patch-2` that would hopefully address your latest comment. Not entirely sure though since I am going off your instructions & haven't familiarized myself with the process.

schuyler1d (approved):
:+1:

PR #1946, opened 2021-03-16
Update REFERENCE-environment_variables.md and Update and rename HOWTO_BUY_NUMBERS_IN_TWILIO.md to HOWTO_BUY_NUMBERS_IN_SPOKE.md

Frydafly:
Now that you can can request a Spoke API key in VAN (and not a Hustle one), I've changed the documentation accordingly.


docs/REFERENCE-environment_variables.md
```diff
@@ -64,7 +64,7 @@
...
-| NGP_VAN_API_KEY                                  | API key. Request an API key on the API Integrations section of VAN. Select `Hustle` API key in VAN. _Required_ for VAN integration_.                                                                                                                                                                                                                                                                                        |
+| NGP_VAN_API_KEY                                  | API key. Request an API key on the API Integrations section of VAN. Select `Spoke` API key in VAN. _Required_ for VAN integration.                                                                                                                                                                                                                                                                                        |
```
schuyler1d: I think NGPVAN is maybe changing some of this -- it would be great to link to their documentation if they have a place that points out how to get this info.  Would it also be useful to link here to the action handler, message handler, contact loader integrations for NGPVAN for those setups?

maxkc: Not sure if it's worth expanding on this (there are technically two different Spoke API types) but here's some additional info in case you want to use any of it @Frydafly.

```
| NGP_VAN_API_KEY                                  | API key. Generate an API key on the [API Integrations section of VAN](https://everyaction.readme.io/docs/key-generation-and-usage#how-do-i-get-a-production-client-api-key). Select `Spoke` or `Spoke Basic` API key type. Depending on a user's permission level the key may be generated automatically or need to be approved by their admin (users should contact their organization's VAN admin to request additional permissions as needed). `Spoke Basic` only supports writing data back into VAN, it does not have access to Export Jobs, while `Spoke` can export. Most users will only see `Spoke` but VAN clients who do not allow exports via API can request to have `Spoke Basic` added. _Required_ for VAN integration.                                                                                                                                                                                                                                                                                        |
```

Frydafly: Thanks @maxkc! That's real helpful.

Frydafly:
@schuyler1d I made a few more commits to this that are unrelated to the VAN API, so just flagging that I think you should also review this. thank you!!

Frydafly:
hey @schuyler1d flagging this documentation change for you! I've made some more commits based off of conversations with organizations that have prompted me to find inconsistencies, duplicates, and more in these documents.

docs/REFERENCE-environment_variables.md
Frydafly: Incorporated your suggestions @maxkc !

schuyler1d (dismissed):
:+1:

schuyler1d (approved):
yei!

PR #1944, opened 2021-03-13
Update HOWTO_DEPLOYING_AWS_LAMBDA.md

Frydafly:
# Fixes https://github.com/MoveOnOrg/Spoke/issues/1909

## Description

Updated documentation to regarding the `--role SpokeOnLamba` to specify that this role has to be created to give the baseline Lambda permissions and add AmazonS3FullAccess. Thanks @ffaristocrat 

# Checklist:

Not applicable.

- [ ] I have manually tested my changes on desktop and mobile
- [ ] The test suite passes locally with my changes
- [ ] If my change is a UI change, I have attached a screenshot to the description section of this pull request
- [ ] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#your-first-code-contribution), or has a documented reason in the description why it’s longer
- [ ] I have made any necessary changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] My PR is labeled [WIP] if it is in progress

PR #1947, opened 2021-03-17
Update HOWTO_DEPLOYING_AWS_LAMBDA.md (stac labs edition)

### LINKED ISSUE #1909: ">Claudia create --role SpokeOnLambda" significance
Hello,

Maybe this is a dumb question, but what is the significance to the --role SpokeOnLambda flag in the Deploy Step of the AWS Lambda Instructions? I'm currently struggling to get that working, and I'm wondering if this is why. The rest of the documentation never mentions creating a SpokeOnLambda role. Is this something that has to be done before deployment, or will executing the script in the "Deploy" section create it?

I'm currently getting Access Denies when running Claudia even though when I run AWS s3 or s3cmd directly it works fine.

ffaristocrat:
# Fixes # (issue)

## Description

Adds some lessons and clarifications to the Lambda deployment docs that we learned during our own deployment at stac-labs. There's obviously some overlap with #1944.

# Checklist:

- [N/A] I have manually tested my changes on desktop and mobile
- [N/A] The test suite passes locally with my changes
- [N/A] If my change is a UI change, I have attached a screenshot to the description section of this pull request
- [N/A] I have added tests that prove my fix is effective or that my feature works
- [N/A] My PR is labeled [WIP] if it is in progress

### DIFF
--- a/docs/HOWTO_DEPLOYING_AWS_LAMBDA.md
+++ b/docs/HOWTO_DEPLOYING_AWS_LAMBDA.md
@@ -31,7 +31,7 @@ Waiting for verification for a certificate may take a while so we will start tha
 
 ## S3
 
-Create a private S3 bucket by choosing all the default values as you go throuh the setup wizard. We will call this `textforcampaign`.
+Create a private S3 bucket by choosing all the default values as you go through the setup wizard. We will call this `textforcampaign`. Make sure the bucket is publicly accessible.
 
 ## VPC
 
@@ -73,6 +73,8 @@ Then, edit `Spoke - Private`'s routes to connect to the NAT you just created:
 | 0.0.0.0/0   | nat-xxxxxxxxxxxxx |
 ```
 
+Make sure all of the subnets are assigned to their respective route tables.
+
 ### Security Groups
 
 Security groups give you fine grained access control for resources.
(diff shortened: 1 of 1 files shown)


docs/HOWTO_DEPLOYING_AWS_LAMBDA.md
```diff
@@ -31,7 +31,7 @@ Waiting for verification for a certificate may take a while so we will start tha
 
 ## S3
 
-Create a private S3 bucket by choosing all the default values as you go throuh the setup wizard. We will call this `textforcampaign`.
+Create a private S3 bucket by choosing all the default values as you go through the setup wizard. We will call this `textforcampaign`. Make sure the bucket is publicly accessible.
```
schuyler1d, line 34: Is it worth noting here that thre s3 bucket used for S3_STATIC_PATH and STATIC_BASE_URL (same one) should definitely be publicly accessible.

As a policy, AWS_S3_BUCKET_NAME does need to allow objects to be publicly accessible (basically turn OFF 'block public access'), however there shouldn't be a blanket visibility -- in fact that would be very bad for security.
MoveOn doesn't use the same s3 bucket for all of those things (we have two -- one for the public code and one for the backend files (exports, and s3 contact loader)) but I suppose that's not necessary since the package.json `prod-static-upload` command adds acl-public

PR #1957, opened 2021-03-22
Optimize Campaign Interaction Step creation UI performance

codygordon:
## Description

Our folks often create very complex interaction step flows – compounded also by many van survey response options and type-to-search for these causing a re-render of all interaction steps with each letter typed – the result is when attempting to map a VAN survey response (or even type a question text, etc) the page just became effectively unresponsive.

I was able to dramatically improve the performance of the UI by extracting the interaction step into a separate component and preventing re-renders using `shouldComponentUpdate`.

Other small changes:
- Creating a new step automatically scrolls to that step – it was difficult to find new steps in complex trees.
- Added a `NONE` option to the Action Handler dropdown to clear it if you don't actually want to set one.

### DIFF
--- a/src/components/CampaignInteractionStepsForm.jsx
+++ b/src/components/CampaignInteractionStepsForm.jsx
@@ -1,54 +1,16 @@
 import type from "prop-types";
 import React from "react";
 import RaisedButton from "material-ui/RaisedButton";
-import IconButton from "material-ui/IconButton";
-import DeleteIcon from "material-ui/svg-icons/action/delete";
-import { Card, CardHeader, CardText } from "material-ui/Card";
-import theme from "../styles/theme";
 import CampaignFormSectionHeading from "./CampaignFormSectionHeading";
-import HelpIconOutline from "material-ui/svg-icons/action/help-outline";
-import Form from "react-formal";
-import GSForm from "./forms/GSForm";
-import yup from "yup";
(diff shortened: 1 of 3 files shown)


src/components/CampaignInteractionStepsForm.jsx
```diff
@@ -72,25 +34,9 @@ export default class CampaignInteractionStepsForm extends React.Component {
...
-    this.state.displayAllSteps is used to cause only the root interaction
-    node to render when the component first mounts.  ComponentDidMount sets
-    displayAllSteps to true, which forces render to run again, this time
-    including all interaction steps. This cuts half the time required to
-    render the interaction steps after clicking the header to expand the
-    interaction steps card.
-
-    FUTURE: reevaluate this after React >=16 upgrade
-  */
-  componentDidMount = () => {
```
codygordon, line 89: this wasn't doing anything so just removed it. i tried to re-create the intended functionality, but couldn't get tests to work properly with the delayed rendering so scrapped it.

schuyler1d:
Just a heads up, this will probably require (minor) changes due to our react/react-formal upgrade (also why I'm not currently adding it to 10.1) -- if you have time, maybe you can test against branch `stage-main-10-1c` I think it's pretty simple -- s/type/as/ for Form.Field or something -- take a look at other diffs

PR #1958, opened 2021-03-22
Export all contacts including those without assignment

### LINKED ISSUE #1954: Bug: Include Unsent Initials (Needs First Message) in Contacts Export
**Problem**
When exporting data from campaigns, contacts that were not sent an initial message are not currently exporting. 

**Solution**
Let's include uncontacted contacts (status = Needs First Message) in contacts export with appropriate status data

codygordon:
# Fixes https://github.com/MoveOnOrg/Spoke/issues/1954


schuyler1d (approved):
:+1:

PR #1959, opened 2021-03-23
Mobilize event shifter autofill data fixes

codygordon:
## Description

Need to parse stringified `customFields` and use `contact.zip`


schuyler1d (approved):
:+1:

PR #1960, opened 2021-03-23
Message review table overflow visible for long tags & truncate long tag text

### LINKED ISSUE #1894: Bug: In the Message Review table of messages, long tags get truncated so you cannot click the "X" button
**Describe the bug**
In the message list in Message Review, TagChips with long labels that overflow the "View Conversation" column are truncated, cutting off the "X" button to delete the tag.

In some cases, you can workaround this by closing the side navigation or resizing the browser window, but it would be nice not to depend on that and to be able to access the delete button and full tag text, especially since these workarounds won't work for some combinations of long tag labels, large text sizes, and small screens.

**To Reproduce**
1. Add a tag with a long label to a contact in a campaign.
2. Go to Message Review for that campaign.
3. Find the row for the tagged conversation in the incoming message table.
4. If necessary, resize browser window horizontally until the columns in the message table are narrower than the chip for the tag.
5. Observe that the tag is truncated, and you cannot scroll horizontally to get to the delete button.

**Expected behavior**
I expect to be able to scroll horizontally to both view the entire tag chip and access the delete button..

**Screenshots**
Screen Shot 2020-11-27 at 7 15 54 PM

 - OS: macOS
 - Browser: Firefox and Chrome
 - Desktop or Mobile? Desktop
 - Version: Firefox 83, Chrome 87

**Additional context**
There might be other feasible UI solutions. I think the minimum requirements are that the user be able to view the entire tag text and access the delete button. Horizontal scrolling is what I expected when I encountered this bug, though, and I was frustrated to notice that I couldn't scroll.

codygordon:
## Description
@navinsivakumar: In the Message Review incoming message list, allow long TagChips to overflow their table cells and allow horizontal scrolling of the table. This ensures that users can both see the entire tag text and access the delete button when using any combination of small enough screen, long enough tag text, and large enough text size such that the TagChip does not fit entirely inside its table cell.

@codygordon: I also set a max-width for tag text so they can't be infinitely long  (we should probably have a character limit on tag names anyhow). 

Also darkened the tag text to make them pop a bit more.

Resulting in:

Screen Shot 2021-03-23 at 8 31 26 AM

### DIFF
--- a/src/components/IncomingMessageList/index.jsx
+++ b/src/components/IncomingMessageList/index.jsx
@@ -231,7 +231,7 @@ export class IncomingMessageList extends Component {
       label: "View Conversation",
       style: {
         textOverflow: "ellipsis",
-        overflow: "hidden",
+        overflow: "visible",
         whiteSpace: "pre-line"
       },
       render: (columnKey, row) => (
@@ -389,6 +389,7 @@ export class IncomingMessageList extends Component {
           onRowSizeChange={this.handleRowSizeChanged}
           onRowSelection={this.handleRowsSelected}
           selectedRows={clearSelectedMessages ? null : this.state.selectedRows}
+          tableBodyStyle={{ overflowX: "auto" }}
         />
         <ConversationPreviewModal
           organizationTags={this.state.tags}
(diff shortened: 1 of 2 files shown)


schuyler1d (approved):
:+1:

PR #1961, opened 2021-03-23
Allow custom (extra) texter profile fields to be used in scripts

codygordon:
## Description

We need to insert custom texter profile fields into scripts which resulted some significant changes to the existing `profile_fields` implementation. cc @jeffm2001

- Pull in this setting from `getConfig` so we can use an instance-wide variable

- For consistency in variable naming, changed it to `TEXTER_PROFILE_FIELDS`, however using `profile_fields` setting var name will still work for backwards compatability

- Added an `isRequired` key to the profile settings objects – this is really necessary for WFP so users can optionally enter some customized data. For backward compatibility, if legacy extra profile fields are unchanged and `isRequired` is not defined, those fields will be treated as if they had `"isRequired": true` so expected behavior of existing fields doesn't change. For new fields added, `"isRequired": false` should be added if it is optional.

- ~~Added a conditional env var called `SCRIPTS_USE_TEXTER_PROFILE_FIELDS` that will make any extra texter profile fields defined in `TEXTER_PROFILE_FIELDS` populate in the script editor. These will show up in the script editor with a prefix of `texter_`~~

Example of working `isRequired` vs not fields:
Screen Shot 2021-03-27 at 10 31 44 AM

Showing up in script editor:
Screen Shot 2021-03-27 at 10 46 01 AM

# Checklist:


src/lib/scripts.js
```diff
@@ -89,8 +89,24 @@ const getScriptFieldValue = (contact, texter, fieldName) => {
   } else if (TOP_LEVEL_UPLOAD_FIELDS.indexOf(fieldName) !== -1) {
     result = contact[fieldName];
   } else {
-    const customFieldNames = JSON.parse(contact.customFields);
-    result = customFieldNames[fieldName];
+    // first try to match on texter extra fields
```
codygordon, line 92: It would be better to check here to see if the env var exists, but I'm not sure how to hook into an instance or org -level setting variable here.

Either way, this works and any issues that arise here would be an extreme edge case where the env var is not enabled, an admin uploads a file containing a custom contact column prefixed with `texter_`, and there happens to be a matching texter extra field with the same name as the contact custom field uploaded

jeffm2001:
I don't have time right now to test this thoroughly, but glancing through the code, it seems fine to me. My one question is do we actually need the `SCRIPTS_USE_TEXTER_PROFILE_FIELDS` env var? Wondering if we should just always make the tester profile fields available in the script. I'm trying to think of a scenario where that would be a problem, and not coming up with one.

codygordon:
@jeffm2001 I can't either but implemented it this way with the spirit of disrupting others as little as possible, so for most folks they wouldn't see any change unless they choose to (of course this time other than the breaking change regarding var name / `isRequired` I mentioned).

I was also considering adding another key to the fields themselves, like `useInScripts` or the like.

schuyler1d (changes requested):
++ let's remove the environmental variable -- I appreciate the care to avoid disrupting folks, but this has no backwards-incompatible implications so we don't need one.

We only need new variables if it would break previous things (e.g. if you were *changing* or removing the availability of those script choices).

schuyler1d (approved):
:+1:

src/server/api/organization.js
```diff
@@ -131,11 +131,25 @@ export const resolvers = {
         ? batchPolicies.split(",")
         : ["finished-replies", "vetted-texters"];
     },
-    profileFields: organization =>
-      // @todo: standardize on escaped or not once there's an interface.
-      typeof getFeatures(organization).profile_fields === "string"
-        ? JSON.parse(getFeatures(organization).profile_fields)
-        : getFeatures(organization).profile_fields || [],
+    profileFields: organization => {
+      let fields = getConfig("TEXTER_PROFILE_FIELDS", organization) || [];
```
schuyler1d: This is backwards incompatible -- We should test e.g. for `let fields = getConfig("TEXTER_PROFILE_FIELDS", organization) || getConfig("profile_fields", organization) || []`

src/server/api/user.js
```diff
@@ -367,13 +367,21 @@ export const resolvers = {
     profileComplete: async (user, { organizationId }, { loaders }) => {
       const org = await loaders.organization.load(organizationId);
       // @todo: standardize on escaped or not once there's an interface.
-      const profileFields = getFeatures(org).profile_fields;
-      const fields =
-        typeof profileFields === "string"
-          ? JSON.parse(profileFields)
-          : getFeatures(org).profile_fields || [];
+      let fields = getConfig("TEXTER_PROFILE_FIELDS", org) || [];
```
schuyler1d: same

src/server/models/cacheable_queries/campaign.js
```diff
@@ -49,10 +50,34 @@ const dbCustomFields = async id => {
         .where("campaign.id", id)
     )
     .first();
+
   if (firstContact) {
-    return Object.keys(JSON.parse(firstContact.custom_fields));
+    customFields = Object.keys(JSON.parse(firstContact.custom_fields));
   }
-  return [];
+
```
schuyler1d, line 57: is this part necessary?  This is caching the field list to be sent about the contact -- I don't think we need this at all as these are *texter* fields, not contact fields.

schuyler1d (changes requested):
I didn't realize this was backwards incompatible -- I would like to avoid that.

Further with a bit extra scrutiny I don't think the cacheable campaign thing needs to be included.

src/server/models/cacheable_queries/campaign.js
codygordon: @crayolakat to answer @schuyler1d's question here, this is necessary because it's grabbing the existing contact custom fields to be merged with the texter fields to be cached on the campaign, allowing them to be used in scripts.

see this line: 
https://github.com/MoveOnOrg/Spoke/pull/1961/files#diff-612db26be6b7f5a0cdd54b604ffa5d5afb3bd8208271fea120755796f1803972R77

crayolakat:
Hi @codygordon , while testing, I am seeing the following error message
image

That message appears when I click into a required field, and then click out without typing anything. Is the error message supposed to say that, or should it say "Zip Code is a required field" similar to what's in the screenshot provided in the description of your PR?

codygordon:
@crayolakat sorry this took so long, I totally missed this and thought it had already been deployed! 

I fixed the weird validation message, as well as a couple bugs I noticed while re-reviewing.

crayolakat (changes requested):
Thanks @codygordon !

Another small thing I noticed:

1. Create a new user and join an organization. You'll be taken to a page asking you to complete your profile
2. Don't fill out any additional fields and click "SAVE"
3. Issue: No validation errors appear telling the user they need to fill out required fields

Can we add validation errors at that step?

image

codygordon:
@crayolakat good flag, this wasn't working properly when `extra` was null or undefined. [Pushed a fix](https://github.com/MoveOnOrg/Spoke/pull/1961/commits/a0ead3e82ada0e3db0ea44a3df9fe3f55623018d)!

PR #1978, opened 2021-04-17
Upgrade material UI

stefanhayden:
# Fixes # (issue)

## Description

This monster PR will upgrade `"material-ui": "^0.20.1"` to `"@material-ui/core": "^4.11.3"`

# Checklist:


stefanhayden:
ATM I don't think pagination on the new data tables are working correctly. Some tables I didn't have enough data to test properly. :(

schuyler1d:
QA found a bug in the dynamicassignment section for campaigns.  see commit 7edfb11c065eafb4c59cd3d40027888b344e7c0c (in stage-main-11-0 -- you can cherry pick back in if you want).

While looking for other places to worry about that had onClick in the Form.Submit button instead of GSForm onSubmit prop, I found that saving in the OrganizationFeatureSettings.jsx doesn't work.  I first added `.transform( cv => isNaN(cv) ? undefined : cv)` to the yup fields after number, but I can't seem to trigger onSubmit or onClick at all there -- might be something silly.  Can you investigate?

schuyler1d:
Few other bugs:
* the admin people page's filters aren't selectable.  If you select "Filter by Last Name" e.g.  it just stays as Filter by First Name
* on Message Review, the Message Filter campaigns= parameter in the url is set to NaN instead of the campaign id (it works when there, but if you click 'reload' then it fails to load and you see the NaN's in the url)

Issue #1992, opened 2021-05-06
Bug: contact.external_id not available in TexterTodo container


johntwyman:
**Describe the bug**
The TexterTodo container loads in a range of data entities including `contact`. That object however does not include the `external_id` property.

**To Reproduce**
1. Modify/Create a texter sidebox so it makes use of `contact.external_id`. A good example is a simple sidebox with a button to execute `console.log(contact)` or `console.log(contact.external_id)`.
2. Enable your texter sidebox
3. As a Texter, load up a conversation such that your sidebox is displayed
4. Trigger your sidebox behaviour

**Expected behavior**
The `contact.external_id` property should be available in this container context.

**Platform (please complete the following information):**
N/A

**Additional context**
TexterSideboxes offer a lot of opportunities for integration with external systems. As such, it's important for the external_id of any Spoke contact to be accessible in the TexterTodo container.

johntwyman:
Closing following discussion with @schuyler1d regarding security/design decisions, etc.

PR #1993, opened 2021-05-06
Add contact.external_id to TexterTodo container for sidebox use

johntwyman:
# Fixes #
This fixes issue #1992 and exposes a contact's `external_id` property for use within Texter Sideboxes.

## Description

I've added external_id to the `contactDataFragment` defined in the TexterTodo container. This makes it visible to Texter Sideboxes that may wish to use it for third party system integrations/actions.

# Checklist:

### DIFF
--- a/src/containers/TexterTodo.jsx
+++ b/src/containers/TexterTodo.jsx
@@ -13,6 +13,7 @@ export const contactDataFragment = `
         lastName
         cell
         zip
+        external_id
         customFields
         optOut {
           id


schuyler1d:
If any texter could get campaign contact external_id, that would be a security leak, so we require that it's 'needed' before supporting.  This is currently, somewhat hacky, but see:
https://github.com/MoveOnOrg/Spoke/blob/main/src/server/models/cacheable_queries/campaign.js#L103-L108
Which texter-sidebox do you need it for?

johntwyman:
> If any texter could get campaign contact external_id, that would be a security leak, so we require that it's 'needed' before supporting. This is currently, somewhat hacky, but see:
> https://github.com/MoveOnOrg/Spoke/blob/main/src/server/models/cacheable_queries/campaign.js#L103-L108
> Which texter-sidebox do you need it for?

Thanks for the reply @schuyler1d. In our case we've been asked to make a few things easy/easier to do within Spoke, for which a custom sidebox feels most appropriate/user friendly:
- a "Set 'do not SMS' flag on CMS record" action
- a "Open this person's record in our `<insert external system here>`" action

We've written an action handler for the first but that approach feels sub-optimal; it has to be assigned to every possible "opt out" response/answer because (a) we can't add a custom action to the Spoke opt-out button nor can we (b) add one to canned responses.

The latter is highly requested - we're using SMS to recruit people to polling stations and want to be able to bring up their record in our polling roster tool to allocate them while the conversation is ongoing. 

"Hi John, last election you helped out at the Random Community Hall. Would you be happy to do so again on June 2nd? Yes? Great; I'll put you down for 8-10am like last time", etc.

schuyler1d:
Hi @johntwyman 

We'll be coming out with Spoke 11.0 soon which should have a new extension point where I added an onOptOut event that you could hook into.  See this commit: https://github.com/MoveOnOrg/Spoke/commit/da8d13708

I get your argument now for this PR

johntwyman:
Hi @schuyler1d, thanks for the update. I'm happy for this PR to be closed given the current architecture and security decisions you've made. The upcoming release certainly sounds encouraging and will address the more "critical" concern we have.

PR #1988, opened 2021-05-05
Separated Privacy Policy and TOS links and added a point to remind th…

Akash-Jairam:
…e user to replace the placeholders with their actual TOS and Privacy Policy

# Fixes #1392

## Description

_Separated Privacy Policy and TOS links and added a point to remind the user to replace the placeholders with their actual TOS and Privacy Policy_

# Checklist:

- [ ] I have manually tested my changes on desktop and mobile
- [ ] The test suite passes locally with my changes
- [ ] If my change is a UI change, I have attached a screenshot to the description section of this pull request
- [x ] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#your-first-code-contribution), or has a documented reason in the description why it’s longer
- [x ] I have made any necessary changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] My PR is labeled [WIP] if it is in progress

PR #1975, opened 2021-04-10
Bound keyboard shortcuts to the advance and go back arrows

Akash-Jairam:
## Fixes #1400 
## Description
- Binds the "<" and ">" buttons to "ctrl+<" and "ctrl+>" respectively on Windows and Linux as well as "command+<" and  "command+>" respectively on Mac
- [Here's a link to the Mousetrap documentation explaining my use of "mod"](https://craig.is/killing/mice)
-  I would really appreciate working with a senior developer to test this as I'm not quite confident navigating through the ui as well as making sure the tests are passed. The code did compile without any errors, however.
- I would also appreciate working with someone to create any needed tests for this

# Checklist:

- [ ] I have manually tested my changes on desktop and mobile
- [ ] The test suite passes locally with my changes
- [ ] If my change is a UI change, I have attached a screenshot to the description section of this pull request
- [X] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#your-first-code-contribution), or has a documented reason in the description why it’s longer
- [ ] I have made any necessary changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] My PR is labeled [WIP] if it is in progress

PR #2005, opened 2021-06-01
Stage-main 11.0 candidate

schuyler1d:
# Included
* #1978 
* #1975 
* #1988 
* #1998 

# Draft Release Notes

## v11.0

_August 2021:_ Version 11.0

This major release upgrades several backend libraries and significantly extends and refactors the way vendors (e.g. Twilio) are connected.  First, Stefan Hayden has helped us upgrade our Material UI library -- so all components may look *slightly* different in style, but nothing should look unfamiliar. This will make future UI improvements much easier and for developer-contributors to use current documentation and resources (and less buggy!) for continuing UI evolution.

More details below for a few migration steps depending on your deployment:
* There is a database migration for anyone upgrading -- instances with a very large message table have special instructions
* AWS Lambda deployments have slightly different deployment commands now
* Anyone using  EXPERIMENTAL_CAMPAIGN_PHONE_NUMBERS or EXPERIMENTAL_TWILIO_PER_CAMPAIGN_MESSAGING_SERVICE require setup changes (these *were* experimental features)

#### Service vendors and Service managers extensions

Based on work from Larry Person, there is a large refactor of "service-vendors" which makes it easier to contain the code to support a vendor connection like Twilio (and others -- there is an experimental Bandwidth.com implementation now, as well). Service Managers is in-turn an extension-api that allows one to hook into service-vendors and other campaign events easily. Adam Greenspan has created a Sticky Sender feature which allows one to keep the same phone number across conversations, so e.g. Twilio message services aren't necessary.

#### Additional changes

* Redis upgrade -- please report any issues -- new Heroku installs support Redis 6 which requires a TLS connection
* keyboard shortcuts for advancing left/right
* service-managers: carrier-lookup, scrub-bad-mobilenums, and per-campaign-messageservices (replacing EXPERIMENTAL_TWILIO_PER_CAMPAIGN_MESSAGING_SERVICE) 
* NGPVAN updates and fixes to use their changed/most recent API

#### Database upgrades
* This is a major release and includes a schema change.
* For small instances simply leave/set SUPPRESS_MIGRATIONS="" or for [AWS Lambda, see the db migration instructions](https://moveonorg.github.io/Spoke/#/HOWTO_DEPLOYING_AWS_LAMBDA?id=migrating-the-database). 
* For instances with a large `message` table, we recommend setting NO_INDEX_CHANGES=1 before running the migration, and then manually running two commands:
  * `CREATE INDEX CONCURRENTLY cell_msgsvc_user_number_idx ON message (contact_number, messageservice_sid, user_number);`
  * `DROP INDEX cell_messageservice_sid_idx;`

The schema changes include adding a new table `organization_contact` which will track contacts across campaigns in an organization -- for things like subscription_status (in future), whether the number is a landline or what number has been used to contact them in the past (for 'sticky' sending). We also add user_number at the end of already-indexed cell-messageservice, to make it easier to search for user_numbers (also for sticky sending features).

#### AWS Deployment changes

Instead of running a single 'claudia' command, You will need to tweak two things:
* Add an environment variable `ASSETS_DIR_PREBUILT` and set it to the absolute directory of your deployment checkout directory + "/build/client" (or whatever you have your ASSETS_DIR var set to).  For example, on a Mac it might be something like "/Users/Sky/spoke/build/client"
* Instead of a single deployment command, you will first need to run
  1. `ASSETS_DIR=./build/client/assets ASSETS_MAP_FILE=assets.json NODE_ENV=production yarn prod-build-client`
  2. and then for your `claudia update` command you need to include your usual command line parameters and ADD `--no-optional-dependencies`

These steps remove the client-side libraries from the server-side build, which is necessary now that the client-side libraries are too large to 'fit' into an AWS Lambda server deploy file.  This is documented in the [AWS setup/deploy steps](https://github.com/MoveOnOrg/Spoke/compare/main...stage-main-11-0#diff-548e8f704ad84645a42f2efaf1332490f6844d0a0dd50e9ac6b931c198d213f3)

#### Changes for Experimental Per-campaign phone numbers/message services

For those that used the experimental feature EXPERIMENTAL_CAMPAIGN_PHONE_NUMBERS, it has been moved and refactored into a  or Service Manager extension -- for these experimental installs (ONLY!), change to setting SERVICE_MANAGERS=per-campaign-messageservices

EXPERIMENTAL_TWILIO_PER_CAMPAIGN_MESSAGING_SERVICE is no longer supported.  Please create an issue if you still have a use-case for this -- there is tentative work to move its functionality into per-campaign-messageservices as well, but only if it still has users.

### Appreciations
* [Adam Greenspan](agreenspan24), [Akash Jairam](https://github.com/Akash-Jairam), [Arique Aguilar](https://github.com/Arique1104) (our new Community Manager -- Welcome!), [Cody Gordon](https://github.com/codygordon), [Fryda Guedes](https://github.com/Frydafly), [Schuyler Duveen](https://github.com/schuyler1d), [Stefan Hayden](https://github.com/stefanhayden),  and Mark Houghton and [Barbara Atkins](https://github.com/bdatkins) for QA


schuyler1d:
> @schuyler1d Looks like one of the test errors are coming from [this](https://github.com/MoveOnOrg/Spoke/blob/646d1c0b8407341c0b17bc6805842ade6760e860/__test__/workers/assign-texters.test.js#L71) not saving a user successfully. Then the user doesn't exist, violating a primary key. Can you think of a reason why that may fail on node 11 specifically?

Maybe if the user table isn't cleared, and then there's already an id=1 record?

Akash-Jairam:
@schuyler1d , I think I found the source of the cypress issue. It's not exactly a cypress issue, rather a Github issue. The easy fix appears to be using `actions/upload-artifact@v2` instead of `actions/upload-artifact@v1` in `cypress-tests.yaml`. If that doesn't work, we'd have to construct a path relative to `$GITHUB_WORKSPACE`.

[Source/Additional context](https://github.com/actions/upload-artifact/issues/43)

schuyler1d:
@Akash-Jairam , I [changed it to @v2](https://github.com/MoveOnOrg/Spoke/pull/2005/commits/3ebe2d7fbc2bb7d9dd5613e9a4c28bcb9afd1e50) but it looks like it's still failing.  So `$GITHUB_WORKSPACE` should change the paths to something like `$GITHUB_WORKSPACE/cypress/videos`?

Akash-Jairam:
@schuyler1d , it seems like the issue was resolved when I checked the recent test failure. Now there's a lockfile error?

crayolakat (approved):

Issue #2009, opened 2021-06-23
Feature Request: Auto-Opt Out responses


agreenspan24:
**Problem**
The auto opt-out message handler saves an opt out to our database. However, it doesn't let the contact know that we've removed them from our list. Plus, texters don't see the auto opted-out contact, so they can't send an opt-out message manually.

**Solution**
As part of the auto opt-out message handler, send the organization's default opt out message back to the contact being opted out.

**Context**
Twilio has their own effective auto opt-out, which blocks any additional messages after specific keywords are sent. We may need to exempt these specific keywords from this new feature.
https://support.twilio.com/hc/en-us/articles/223134027-Twilio-support-for-opt-out-keywords-SMS-STOP-filtering-

agreenspan24:
@schuyler1d @Frydafly I'm working on this, should be fairly easy. Let me know if you have any additional requirements or comments on it.

crayolakat:
This fix for this was merged into the `main` branch for the 12.2 release, but some rebasing caused the issue to not close automatically. Manually closing this issue.

PR #2010, opened 2021-06-23
Send auto opt out response

agreenspan24:
# Fixes

https://github.com/MoveOnOrg/Spoke/issues/2009

#### Problem
The auto opt-out message handler saves an opt out to our database. However, it doesn't let the contact know that we've removed them from our list. Plus, texters don't see the auto opted-out contact, so they can't send an opt-out message manually.

#### Solution
As part of the auto opt-out message handler, send the organization's default opt out message back to the contact being opted out.

#### Context
Twilio has their own effective auto opt-out, which blocks any additional messages after specific keywords are sent. We may need to exempt these specific keywords from this new feature.
https://support.twilio.com/hc/en-us/articles/223134027-Twilio-support-for-opt-out-keywords-SMS-STOP-filtering-

# Checklist:


docs/REFERENCE-environment_variables.md
```diff
@@ -97,6 +97,7 @@
...
 | ROLLBAR_ENDPOINT                                 | Endpoint URL for Rollbar error tracking.                                                                                                                                                                                                                                                                                                                                                    |
+| SEND_AUTO_OPT_OUT_RESPONSE                           | Send the organization's default opt-out message if a user is auto-opted out. Alternatively, set a shouldAutoRespond property in an item for the auto opt-out regex to true, for it to only apply on those regex matches.                                                                |
```
crayolakat, line 100: `Alternatively, set a shouldAutoRespond property in an item for the auto opt-out regex to true, for it to only apply on those regex matches.`

Won't the setting the SEND_AUTO_OPT_OUT_RESPONSE also only apply on the regex matches in the auto opt-out regex?

crayolakat (commented):
Looks great! I had 1 small question which I left in the comments.

docs/REFERENCE-environment_variables.md
agreenspan24: So if I remember correctly, we envisioned two paths: 
* For those using the default auto opt out regex, or those who wanted to send the message for every auto opt out, they could just set SEND_AUTO_OPT_OUT_RESPONSE to true and be done. 
* For those who only wanted to send the message for some of their opt outs (for example, not respond to the hostile opt outs), they could customize the behavior per auto opt out regex with the shouldAutoRespond property and set SEND_AUTO_OPT_OUT_RESPONSE to false. 

Does that answer your question?

crayolakat: Yes, I just discovered that you can have multiple regex for the customized auto-optouts :)

crayolakat (approved):

crayolakat:
This was merged into the `main` branch for the 12.2 release, but some rebasing caused the PR to not close automatically. Manually closing the PR.

PR #2017, opened 2021-07-27
error handling van contact loader

agreenspan24:
# Description

If an error occurs, just show no options, rather than breaking the whole page. This should point to a VAN auth issue for the user, who probably will be one step removed from making changes to the auth settings.

This is just the easier option to showing an error message directly, but I'm open to building that instead. This at least saves us from the page breaking.

# Checklist:


schuyler1d (approved):
:+1: awesome!  thanks for the follow-up

PR #2025, opened 2021-07-30
datefield: fix for better tab/keyboard support

schuyler1d:
I went down a bit of a rabbithole to avoid having a date-default (which would fail because it's "today's" date and it needs to be in the future), and I think arrived at something better, but I'm getting rid of a bunch of moment/date code and some other ....what I believe is cruft.

### DIFF
--- a/src/components/forms/GSDateField.jsx
+++ b/src/components/forms/GSDateField.jsx
@@ -4,59 +4,34 @@ import moment from "moment";
 import GSFormField from "./GSFormField";
 import { dataTest } from "../../lib/attributes";
 import DateFnsUtils from "@date-io/date-fns";
-import { MuiPickersUtilsProvider, DatePicker } from "@material-ui/pickers";
+import {
+  MuiPickersUtilsProvider,
+  KeyboardDatePicker
+} from "@material-ui/pickers";
 import theme from "../../styles/mui-theme";
 
-export default class GCDateField extends GSFormField {
-  render() {
-    const momentDate = moment(this.props.value);
-    let value = {};
-    let oldDate = null;
-    if (momentDate.isValid()) {
-      const fakeDate = momentDate
-        .add(this.props.utcOffset - moment().utcOffset(), "minutes")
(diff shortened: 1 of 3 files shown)


src/components/forms/GSDateField.jsx
```diff
@@ -4,59 +4,34 @@ import moment from "moment";
...
-                "minutes"
-              );
-              this.props.onChange(newDate.toDate());
-            }
-          }}
           style={{
             marginBottom: theme.spacing(2)
           }}
+          open={this.state.open}
+          {...this.props}
```
stefanhayden, line 34: `{...this.props}` this is an anti pattern. Modern react throws errors when you pass unexpected unexpected props to components. Also this makes making changes on components extra hard as you never know if a change you are making effect some prop you don't know about. If we can get rid of this and pick the props we need to pass in this will not only get rid of errors but also make future development much easier. 

Do you know what props you were trying to pass through?

stefanhayden: oh wait. I see where you moved this from. I see you're just kicking the can down the road that I also kicked down the road. So not a regression! But it would be nice to remove all of that spreading on props.

schuyler1d: Thanks for feedback.
I'm passing through 
https://github.com/MoveOnOrg/Spoke/pull/2025/files/6c525d51da1f44b2724a4ae7cdabea47eb4eca04#diff-ce3ce9d2ab22743073212dd164e9ba673418132730b0c9ad686ea4d3eefeeaf3R95
at least which since this is the only date instance we could push it into there, but I thought it made more sense in the first context.
I get the anti-pattern issue -- so if i tweak again, maybe i'll just pass explicit props, though datefield components have so many, it would be more annoying/confusing to 1. list them all explicitly in the date field or 2. only pass the ones i need, but if someone creates a datefield.
I guess we can do this.props and just remove the ones that were for the Form.Field component.

PR #2027, opened 2021-08-17
Get proper input value for Script Import url

codygordon:
The new Material inputs apparently provide event.target.value in the "old fashioned" way, this handler hadn't been updated. Are there others we may not yet be aware of?


schuyler1d (approved):
lol, sadly yes -- that's what Stefan and I have been debugging throughout the month ;-)

PR #2029, opened 2021-08-18
remove some UI logging

codygordon:
is there a good reason to leave UI logging in commits? i think it's probably a bad habit, but generally it's pretty harmless. however, one of these in particular (`createMenuItem`) i believe was causing some performance issues as it was spamming logs anytime anything was changed


schuyler1d (approved):
I like some console.logs in actions especially when they trace through several functions, but most of these are probably there from past debugging and forgotten to be removed

PR #2031, opened 2021-08-19
clear campaign cache when saving assigned phones

codygordon:
could solve our weird little issue, let's hope!


schuyler1d (approved):
thanks!

PR #2038, opened 2021-09-10
Skip messaging service prod

agreenspan24:
## Description

Sticky Sender: https://github.com/MoveOnOrg/Spoke/pull/1917

Most of the previous code uses messaging services to send to Twilio. But, with messaging services, we ultimately lose control over which user number is used to send to a recipient. It can also lead to overuse of phone numbers. Plus, a messaging service can hold a maximum of 400 phone numbers. Twilio recommends each phone number only send to 200 recipients per day, so a messaging service has a maximum of 80,000 recipients per day. This isn't enough for some campaigns. 

The campaign phone numbers feature establishes a new messaging service for every campaign. This does help expand the maximum for total texts per day. But it still adds a limit of 80,000 per text campaign, so on a high traffic day that number might be exceeded, especially for a national or statewide GOTV effort. Also, it removes the sticky sender effect across campaigns built into messaging services, because the sticky sender mappings are tied to the messaging service.

This feature is intended to work with the EXPERIMENTAL_STICKY_SENDER mode, which is how we know how to continue sending messages with the right user number. If someone wants to text without messaging services and without sticky sender, they would need to build in the ability to associate user numbers with campaign_contact records.

In addition, messaging services are a unique feature of Twilio. If we want to expand capabilities to other message senders, we will need to select user numbers ourselves instead.

In this feature, instead of using a messaging service to select the user number from which a message is sent, we select one from the owned_phone_number table. It must be used with EXPERIMENTAL_STICKY_SENDER and PHONE_INVENTORY. Prioritizes phone numbers by overuse (whether or not they've exceeded the CONTACTS_PER_PHONE_NUMBER limit), matching an area code, matching an overlay area code, matching an area code in the same state, current count of stuck phone numbers (in levels of 50), and randomization, in that order. 

Overlay area codes and state area codes are seeded in the database from a CSV built from the [NANPA](https://nationalnanpa.com/reports/area_code_relief_planning.html), which is an authority on area codes in North America. Overlays are georgraphic areas that share area codes, normally large metro areas.

As future work, we should probably use zip codes in some way, in case someone has moved out of their area since buying their phone number.

This PR is a total of 997 lines of code, but 801 lines of that are the new area code file, so it still meets the 300 line limit. 

# Checklist:


codygordon:
hey @crayolakat and @ibrand are you familiar with this PR, and if so do you know if it was superseded by any other PRs? It could be pretty important – esp with those who use Bandwidth – though it's unclear whether some of this is actually necessary (geomatching area codes, managing "overused" phones).

schuyler1d:
This code was merged/moved to 

https://github.com/MoveOnOrg/Spoke/tree/main/src/extensions/service-managers/sticky-sender

agreenspan24:
> This code was merged/moved to
> 
> https://github.com/MoveOnOrg/Spoke/tree/main/src/extensions/service-managers/sticky-sender

@schuyler1d that was sticky sender, this writes an algorithm for getting an outgoing phone number for the first send. I don't see this ever having been merged, but it would be cool if it was!

schuyler1d:
Ah, that part is here:
https://github.com/MoveOnOrg/Spoke/blob/main/src/extensions/service-managers/numpicker-basic/index.js

Id encourage you to review it though -- I think you might have a slightly better algorithm. If so, id encourage you to improve that one or create a numpicker-somethingsomething that is smarter

agreenspan24:
> Ah, that part is here: https://github.com/MoveOnOrg/Spoke/blob/main/src/extensions/service-managers/numpicker-basic/index.js
> 
> Id encourage you to review it though -- I think you might have a slightly better algorithm. If so, id encourage you to improve that one or create a numpicker-somethingsomething that is smarter

Oh cool! Didn't realize that was included! Does anyone use it? I'd love to get their thoughts to make one that considers area code/region.

schuyler1d:
It's necessary for anyone that uses Bandwidth instead of twilio since they don't have a message service system

agreenspan24:
> It's necessary for anyone that uses Bandwidth instead of twilio since they don't have a message service system

In that case, @codygordon: let me know if I can help implement a smarter num-picker for y'all!

In general, my philosophy (two years ago, mind you, not sure how relevant anymore) was that we would get more responses to send from numbers in the general area of the user, so seemed like a local presence. If you're a national organization, you may want to send from area codes in the recipient's state, for example. We used the recipient's area code to do that, but I now think it would make more sense to use their zip code, since people move across country without changing their phone number. 

But, once we started using more local numbers, we ran the risk of overusing phone numbers from certain areas, so the "stuck contacts" ranking was supposed to balance that out.

codygordon:
@inorvig @ryanjcrump @deasterdaywfp see the followup on this thread and let @agreenspan24 know if you are interested in a smart picker like was implemented here.

And thanks for replying here @schuyler1d hope you're doing well!

PR #2043, opened 2021-09-18
Assignment Contacts Sidebar

agreenspan24:
## Description

Let's make it easier to navigate between conversations.

This adds a toggleable left sidebar to the texter UI, giving them the list of their contacts. It bolds contacts that need a response and orders them first, does plain text for contacts that do not need a response, and grays and italicizes skipped contacts. You can search by name or multi-select filter by status, which also shows the count of each number of contacts. When more than 50 contacts have conversations, we paginate the list, for a cleaner UI. 

By default, the sidebar shows up when the ASSIGNMENT_CONTACTS_SIDEBAR setting is on and the device is not mobile. On mobile, it can still be displayed. We also cookie-fy the last toggle setting in the browser, to save the user's preference across immediate sessions. 

We also re-arrange the existing texter UI slightly to match wireframes - the texter sideboxes sidebar has a blank header bar, which could be used in the future. The Campaign toolbar is also split from the contact details toolbar, though no substantive change is made.

Lastly, we add a 30 second refresh of contacts, so we can see the latest of who needs to be contacted next. 

Figma Designs: https://www.figma.com/file/kLh6i9pKosI9MB9wqey5Fs/Spoke-ContactList

image
image
image
image
image

# Checklist:


stefanhayden:
this looks so good!

src/components/AssignmentTexter/AssignmentContactsList.jsx
```diff
@@ -0,0 +1,301 @@
...
+import FilterListIcon from "@material-ui/icons/FilterList";
+import Menu from "@material-ui/core/Menu";
+import Pagination from "@material-ui/lab/Pagination";
+
+const pageSize = 50;
+
+const messageStatusLabels = {
+  needsResponse: "Respond",
+  convo: "Past",
+  closed: "Skipped"
```
Arique1104, line 20: Hey @agreenspan24, 

@crayolakat and I were wondering what happens to the opted-out texts for this UI?

We can imagine opted-out texts being grayed out along with closed: "Skip" or being removed altogether from the conversations.

Just curious as to what is happening to the conversations that are `automatically opted-out` and the ones `opted-out` by our volunteers?  Does that contact completely disappear from the queue?

src/containers/TexterTodo.jsx
```diff
@@ -182,6 +186,19 @@ export class TexterTodo extends React.Component {
     }
   }
 
+  componentDidMount() {
+    // Get the latest data for the sidebar every 30 seconds to catch new messages
+    if (global.ASSIGNMENT_CONTACTS_SIDEBAR && !this.refreshInterval) {
+      this.refreshInterval = setInterval(this.refreshData, 30000);
+    }
+  }
```
Arique1104, line 194: @agreenspan24, could you confirm that pagination will assist in the performance levels of this call being requested every 30 seconds?  Just wanting to ensure that this type of call doesn't slow Spoke down.

Arique1104 (commented):
Overall, this code looks interesting!  I'm curious to see how it performs during QA.  If all goes well there, I think this would be a great addition to the functionality of the UI offered through Spoke.  Good work!

src/components/AssignmentTexter/AssignmentContactsList.jsx
agreenspan24: Good call out! I think the way it's currently implemented, opted out contacts are getting filtered out entirely, unless we're in some sort of "review" mode. I think that happens regardless of if they were automatically or manually opted out. I'm personally in favor of hiding them, especially since you can't see them currently, but open to your suggestions.

src/containers/TexterTodo.jsx
agreenspan24: The pagination would help with the performance on the front end- the whole list won't re-render, only a portion of it. We can probably improve the front-end performance slightly, by losing some of the clarity on the pagination, but I don't think it would be worth it.

The server and database, however, will get hit for the list of contacts and statuses every 30 seconds, so that could lead to a slow-down on the server side. I think this is necessary so texters can get to recent replies in a timely manner, but we could up this to every 60 seconds or 3 minutes if we're really worried about server capacity.

crayolakat (approved):

Issue #1504, opened 2020-04-23
Feature Request: Frameworks for Orgs to easily add their own theming and branding to Spoke


ibrand:
**Problem**
Orgs currently can't white label Spoke without significant contractor interventions. It would be awesome for orgs to be able to more easily change the colors and theming on their Spoke instance.

**Solution**
A 'short version' of theming could be to consolidate the colors/values into content-related names (instead of colors) into:
`src/styles/theme.js` -- and then maybe use an environment variable to make it directory loadable from a different name (similar to our import pattern for integrations)

That said, there are other ways it could be done. It'd be nice to make it as integration friendly as possible!

arena:
It seems like there are just a handful of styles and variables that could be organized and then controlled from an admin panel --- I'd be game to design one if somebody is going to be developing this. 
- background colors
- font colors
..I think there is no place to add a logo atm (except maybe the favicon)

DinahSanders:
It'd be great if that admin panel also included a "map legend" style display of what colors (if any) the system uses to indicate things. 
A super-fancy version might use that example display to preview the effect of admin-added color styling, just to prevent visually disastrous choices. :D

arena:
Here is a prototype of a color picker on the Settings screen that Admin's see
Note: the mockup uses bitmap screenshots which can be slow to load in Figma
The idea would be to do a first increment of this feature that lets admins change the primary color from their settings panel. In later increments we could add the ability to link to a logo file and pick a theme of colors, etc. For now, here's a preview of what it will be like to change the primary color:
https://www.figma.com/proto/PXVGzPnbizV5QyiD37FTOs/Spoke-Customization?node-id=165%3A2&viewport=-3404%2C872%2C0.18336185812950134&scaling=min-zoom


![image](https://user-images.githubusercontent.com/411070/81243924-f336b700-8fc5-11ea-8b76-e0e397ff9a5d.png)
in case you can't get the prototype to load there is a static mockup as of May 6 ...check the prototype for the most recent design.

arena:
Here is the array for the default color swatches
#000, #022746, #1F4A7C, #0D5FA4, #8BCEEB, #B8E4D0, #1EBEBD, #71B080, #FEDE3A, #FEDE3A, #E41588, #A02466, #A31F32, #443A8B, #254292

arena:
Here is the react color picker https://casesandberg.github.io/react-color/

arena:
[...]

PR #2047, opened 2021-09-26
Material UI theming

### LINKED ISSUE #1884: Feature Request: Indication of Which Organization You Are In
**Problem**
With a Multi-Org set up, once you select an org from the drop down, there is no indication of which org you are in besides the organization id in the url. Looking at the UI itself, however, I can't tell which organization I'm in.

**Solution**
Have a header that says the name of the org at the top or highlight the org you are in within the drop down menu.

**Context**
Header could be above the "Campaigns" title or some marker in the drop down menu
Screen Shot 2020-11-09 at 12 06 31 PM
Screen Shot 2020-11-09 at 12 06 41 PM

stefanhayden:
# Fixes #1884 #1504

## Description

This PR takes advantage of the move to use the most up to date Material UI and its ability to theme. There is now a ThemeProvider. If you want to access the theme from any component just wrap it in the `withMuiTheme` higher order component which will pass in `muiTheme` as a prop in to your component.

example

image

image

image

# Checklist:

- [ x ] I have manually tested my changes on desktop and mobile


crayolakat:
Hi Stefan! Here is a piece of feedback that we got from QA'ing your PR:
The secondary color is default set to #ffffff, and this causes several buttons with color="secondary" set to be invisible. E.g.: See attached screenshot where the "Add new canned response" button is invisible
Screen Shot 2021-10-06 at 2 31 32 PM

Is it possible to make it so that people can't set the secondary color to white or any other hard to see colors?

stefanhayden:
thanks for testing and thanks for the feed back. I've done a bunch of refactoring and I think the experience is much better now. One thing that is weird about this app is there really isn't much use of a secondary color. I think as we change that's something we can think more about. 

I've change the default secondary color from `#FFFFFF` to `#555555`. I've also grabbed a library call `color-difference` that finds the distance of two colors from 0 to 100 (back and white would be 100). I've eye balled what made sense to me and now show an error and stop form submission if that number is 20 or below. Originally I tried a library that measured contrast difference but actually the default theme of Spoke fails having enough contrast with the green on white. something to think about for the future. Also in the future this can be updated to account for dark mode as well (which with this change out app not fully supports if we had an interface for).

example

I've also updated the `GSColorPicker` to default have an error if a non-valid hex code is passed in. 

here is what that "Add new canned response" button looks like now: 
image

crayolakat:
Thanks Stefan! Some additional feedback:

There are still references to theme in `src/containers/Settings/index.jsx`, causing a blank screen when trying to view the Settings page in some organizations. I replaced the `theme` references with `muiTheme` in my QA branch (https://github.com/MoveOnOrg/Spoke/commit/7dacc052b1560b688aa1cb9e7fcffe44f5e5be9e), and the Settings page is loading now. There is this error in the console however: ```The prop `open` is marked as required in `ForwardRef(Dialog)`, but its value is `undefined`.```
image

stefanhayden:
@crayolakat I updated those colors and found more I also missed. I could not get that error to show. I tested both on this branch and your `stage-main-11-1-material-ui-theming2` branch. Do you have any additional steps to reproduce?

crayolakat:
@stefanhayden After doing a bit of digging, I figured out that this issue isn't related to your PR, and is also present in the current release of Spoke. I documented the issue and reproduction steps here: https://github.com/MoveOnOrg/Spoke/issues/2060

Thank you so much  for the updates. I will run your PR through another round of QA, and let you know if there's any additional feedback!

crayolakat (changes requested):
Hi Stefan! 2 pieces of feedback:

1. When creating a new campaign, the "Primary color" input field is marked red with the message "Invalid color" when the field is blank. However, a blank value is a valid value since the "Primary color" field isn't a required field, and if the campaign is created without a "Primary color" input, it uses the default color with no issues
image
2. The app now crashes in the Texter TODO view if an invalid "Primary color" value is entered. Can this behavior be modified to how it currently is on the `main` branch, and use the default color if the "Primary color" value is invalid?
image

stefanhayden:
oops! I'll have it fixed by Monday! thanks!

stefanhayden:
@crayolakat this should be fixed now.

crayolakat (approved):
Looks good, thanks!!

PR #2040, opened 2021-09-12
#2039 Fix twilio config buttons

lperson:
# Fixes #2039 

## Description

Update save buttons so they work as expected in the new material-ui world.

# Checklist:

- [x] I have manually tested my changes on desktop and mobile
- [x] The test suite passes locally with my changes
- [ ] If my change is a UI change, I have attached a screenshot to the description section of this pull request
- [x] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#your-first-code-contribution), or has a documented reason in the description why it’s longer
- [ ] I have made any necessary changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] My PR is labeled [WIP] if it is in progress

PR #2048, opened 2021-09-27
fix twilio credential saving -- onClick does not work on Form.Submit buttons

schuyler1d:


schuyler1d:
dupe of https://github.com/MoveOnOrg/Spoke/pull/2040

PR #2064, opened 2021-10-26
document --no-optional-dependencies for aws lambda

schuyler1d:
which was in release notes to 11.0 but not carried over to docs

### DIFF
--- a/webpack/config.js
+++ b/webpack/config.js
@@ -21,8 +21,8 @@ const plugins = [
   )
 ];
 const jsxLoaders = [{ loader: "babel-loader" }];
-const assetsDir = process.env.ASSETS_DIR;
-const assetMapFile = process.env.ASSETS_MAP_FILE;
+const assetsDir = process.env.ASSETS_DIR || "./build/client/assets";
+const assetMapFile = process.env.ASSETS_MAP_FILE || "assets.json";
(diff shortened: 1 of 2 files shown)


webpack/config.js
```diff
@@ -21,8 +21,8 @@ const plugins = [
   )
 ];
 const jsxLoaders = [{ loader: "babel-loader" }];
-const assetsDir = process.env.ASSETS_DIR;
-const assetMapFile = process.env.ASSETS_MAP_FILE;
+const assetsDir = process.env.ASSETS_DIR || "./build/client/assets";
+const assetMapFile = process.env.ASSETS_MAP_FILE || "assets.json";
```
Arique1104, line 25: Hey @schuyler1d, do you think the reason there is a test failing is because we need to upgrade the tests or is there maybe something afoot with the pathway to accessing assets.json?

Arique1104: Maybe not.  I went through the codebase to see if there was a file in our code called `assets.json` and we don't have one.  I don't currently understand why there are tests failing after this change.

schuyler1d: looks like a spurious failure -- i kicked off a re-run.  Annoying that the tests sometimes fail :-( :crossed_fingers:

crayolakat (approved):

PR #2065, opened 2021-10-26
Additional scrub / service manager fixes 

### LINKED ISSUE #2096: Bug: scrub-bad-mobilenums ServiceManager.fullyConfigured error
**Describe the bug**

**To Reproduce**
1. Configure env var SERVICE_MANAGERS=scrub-bad-mobilenums
2. Create new Campaign
3. Load Contacts

**Get error message on server**
 code: 'INTERNAL_SERVER_ERROR',
 error: Error: Resolve function for "ServiceManager.fullyConfigured" returned undefined
 at /spoke/node_modules/graphql-tools/dist/schemaGenerator.js:554:19
 at resolveFieldValueOrError (/spoke/node_modules/graphql/execution/execute.js:531:18)
 at resolveField (/spoke/node_modules/graphql/execution/execute.js:495:16)
 at /spoke/node_modules/graphql/execution/execute.js:364:18
 at Array.reduce (<anonymous>)
 at executeFields (/spoke/node_modules/graphql/execution/execute.js:361:42)
 at collectAndExecuteSubfields (/spoke/node_modules/graphql/execution/execute.js:772:10)

codygordon:
# Fixes

1. Turns out I had to roll back the recent deployment of 11.1 this patch because some campaigns were being sent `fullyConfigured: undefined` from the server causing an error. Our GraphQL settings prevent undefined values being sent to the client so I transformed `undefined` to `null` and this resolved this issue. Maybe there's a deeper reason this was happening specific to WFP's configuration, but I wasn't able to find it.

2. If the scrub job fails delete the job. Still not great user feedback but at least it doesn't appear to be hanging/working, and there's a failure log on the server.

3. I discovered a _very_ edge-case situation where a specific phone number returned a 404 "not found" error from the Twilio API. Left unhandled, the campaign could not proceed without deleting that campaign contact from the db. In this case it makes sense to just delete the phone number because if there's something about it that causes Twilio to choke on it even just attempting validation, I figure it's probably not textable.

### DIFF
--- a/src/extensions/service-vendors/twilio/index.js
+++ b/src/extensions/service-vendors/twilio/index.js
@@ -627,6 +627,19 @@ export async function getContactInfo({
     // console.log('twilio.getContactInfo', contactInfo, phoneNumber);
     return contactInfo;
   } catch (err) {
+    /* oddly, very rarely Twilio returns a 404 error message
+       when looking up what appears to be a valid number
+       https://www.twilio.com/docs/api/errors/20404
+
+       the message appears like:
+      "The requested resource /PhoneNumbers/{cell} was not found"
+
+      the assumption is that this must not be a real number if it can't
+      even be validated, so we should just delete it
+
+      try it with this number:
+      https://lookups.twilio.com/v1/PhoneNumbers/+15056405970?Type=carrier
+    */
     if (err.message.includes("was not found")) {
       return {
         ...contactInfo,
--- a/src/extensions/service-vendors/twilio/index.js
+++ b/src/extensions/service-vendors/twilio/index.js
@@ -583,46 +583,58 @@ export async function getContactInfo({
     // caller-name is more expensive
     types.push("caller-name");
   }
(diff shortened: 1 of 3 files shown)


src/extensions/service-vendors/twilio/index.js
```diff
@@ -583,46 +583,58 @@ export async function getContactInfo({
...
+      contactInfo.status_code = 1;
+    }
+
+    if (phoneNumber.callerName) {
+      contactInfo.lookup_name = phoneNumber.callerName;
+    }
+    // console.log('twilio.getContactInfo', contactInfo, phoneNumber);
+    return contactInfo;
+  } catch (err) {
+    if (err.message.includes("was not found")) {
```
schuyler1d, line 643: I think this is your 404 case -- can you add another line to document when this happens.  Basically, why are we considering this an error that should be preserved/remembered for the phone number instead of maybe a `status_code: 0` where we just don't know.

crayolakat (approved):

PR #2093, opened 2021-12-09
Azure Deployment Documentation

dwbond:
# Fixes #

Adds Azure deployment documentation for Spoke.

## Description

Support initially limited to Azure servers (therefore generally applicable in other contexts) but may expand to include Azure app service support later.


docs/HOWTO_AZURE_DEPLOY.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Enter your virtual machine,
+- Click Review + create
+- Create
+- Click Go to resource
+- We want to ssh into the new server
+- Microsoft will provide you with a public IP address (which is static)
+- In terminal: $ssh <your-username>@<static IP address>
+- Enter your password for this user
+- Note for later(?) for production purposes, you will want to attach permanent disk storage
```
crayolakat: What does `Note for later(?)` mean?

docs/HOWTO_AZURE_DEPLOY.md
```diff
@@ -0,0 +1,149 @@
...
+- Then it’s all up and running
+
+## Install Redis
+- Something set up by default within the Docker container.  It sets up a Redis instance.
+- We’re going to deploy redis separately within the server.
+- Since we’ve already broken the database option out of the Docker File.
+- But need to look for [Azure Cache for Redis instructions] (https://github.com/MoveOnOrg/Spoke/blob/main/docs/HOWTO_CONNECT_WITH_REDIS.md)
+
+## Install Postgres
+- In our documentation there are some information for setting this up.
```
crayolakat: Nitpicky grammar correction: Should be "there is some information"

docs/HOWTO_AZURE_DEPLOY.md
```diff
@@ -0,0 +1,149 @@
...
+- It’s probably doing something wrong with the environment variables.
+
+
+## Heroku Deployment
+- In order for people to download csv’s we used Bucketeer
+- Where the csv files go if you are trying to export data out of spoke.
+- Designated temp directory for csv exports
+- What is the expected workflow is for the data exports
+    - MAILGUN_API_KEY, knows that it works but doesn’t know what’s going on.
+    - USING UBUNTU TO DEPLOY THIS: Email is included as a thing works.
```
crayolakat: These lines sound confusing to me

docs/HOWTO_AZURE_DEPLOY.md
Arique1104: I flushed out the thought more, here are my changes:

- For production purposes, you will want to attach permanent disk storage (Documentation Update Coming Soon)

docs/HOWTO_AZURE_DEPLOY.md
Arique1104: Got it!

docs/HOWTO_AZURE_DEPLOY.md
Arique1104: That's because it was scratch notes from a conversation we were having.  I've updated this section to be more outward facing.

crayolakat (approved):
Due to an existing bug, the automated tests all show as fail in this PR. That bug has been fixed here: https://github.com/MoveOnOrg/Spoke/pull/2090

I tested pulling the changes from that PR into your PR in a separate branch, and the automated tests pass!

BrianVallelunga:
I'm working on getting this set up on Azure right now and came across this post. Is there a reason you went with Docker on a VM instead of AppService to begin with? I'm trying to go the AppServices with container hosting to start, but can't quite get it to work and was wondering if you tried this approach. Thanks.

Arique1104:
@BrianVallelunga 
In a StateVoices Meeting with @dwbond and @crayolakat 

Running an AppService that runs some logic, database, and static-file service is the "right way" but StateVoices went a different route.
@dwbond recommended this video to support the proper integration with an AppService: https://www.youtube.com/watch?v=0BsR9TrR6do

Instead of having 20 data app services with all their own databases, we worked to have all that information on a single server so that we could track capacity.  This way we could update and edit quicker.

The way StateVoices is doing it isn't a full perfect use of Azure services (isn't the way you're "supposed" to do app development in Azure). For the Spoke moveover, the idea was instead of having 20 different AppServices with each on their own database, it was to have 1 server. Limited to the way StateVoices have their stuff set up. Detailed in the docs is the way StateVoices is doing it but it's probably not the right for most people's instances.

Would you be interested in providing your notes for a singular Spoke instance?  StateVoices has to manage many instances and this was how they set it up for themselves.

PR #2099, opened 2021-12-22
Twilio Account Switching

crayolakat:
## Description

Allows for users to switch between Twilio accounts per campaign.

# Checklist:

### DIFF
--- a/src/extensions/service-managers/twilio-account-switching/index.js
+++ b/src/extensions/service-managers/twilio-account-switching/index.js
@@ -45,7 +45,7 @@ export async function onMessageSend({
         twilioAccountSwitchingCreds: {
           accountSid: campaignTwilioAccount.accountSid,
           authToken: await getSecret(
-            `MULTI_TWILIO_AUTH_${campaignTwilioAccount.accountSid}`,
+            `MULTI_TWILIO_AUTH_${campaignTwilioAccount.id}`,
             campaignTwilioAccount.authToken,
             organization
           )
@@ -106,6 +106,12 @@ export async function onCampaignUpdateSignal({
 export async function getOrganizationData({ organization }) {
   const accounts = getFeatures(organization).MULTI_TWILIO;
 
+  // Instantiate orgChanges and orgFeatures upon Settings page load
+  orgChanges = {
+    features: getFeatures(organization)
+  };
+  orgFeatures = JSON.stringify(accounts);
+
   return {
     data: {
       multiTwilio: accounts ? _obscureSensitiveInformation(accounts) : []
@@ -125,11 +131,19 @@ export async function onOrganizationUpdateSignal({
     // Save changes to organization features
     await accessRequired(user, organization.id, "OWNER", true);
     for (let i = 0; i < orgChanges.features.MULTI_TWILIO.length; i++) {
-      orgChanges.features.MULTI_TWILIO[i].authToken = await convertSecret(
-        "MULTI_TWILIO_AUTH_" + orgChanges.features.MULTI_TWILIO[i].accountSid,
-        organization,
-        orgChanges.features.MULTI_TWILIO[i].authToken
-      );
+      const curAccount = orgChanges.features.MULTI_TWILIO[i];
+      const foundAccount = orgFeatures
+        ? JSON.parse(orgFeatures).find(account => account.id == curAccount.id)
+        : null;
+
+      // Only encrypt auth token if it's not already encrypted (new account or updated auth token)
+      if (!(foundAccount && curAccount.authToken == foundAccount.authToken)) {
+        curAccount.authToken = await convertSecret(
+          "MULTI_TWILIO_AUTH_" + curAccount.id,
+          organization,
+          curAccount.authToken
+        );
+      }
     }
     await cacheableData.organization.clear(organization.id);
     await r
@@ -140,13 +154,6 @@ export async function onOrganizationUpdateSignal({
     saveDisabled = true;
   } else {
     // Make changes to organization features
-    if (!orgChanges) {
-      orgChanges = {
-        features: getFeatures(organization)
-      };
-      orgFeatures = JSON.stringify(getFeatures(organization).MULTI_TWILIO);
-    }
-
     orgChanges.features.MULTI_TWILIO = updateData.map(account => {
       const existingAccount = orgChanges.features.MULTI_TWILIO
         ? orgChanges.features.MULTI_TWILIO.find(e => {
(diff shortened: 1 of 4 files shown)


src/extensions/service-managers/twilio-account-switching/index.js
```diff
@@ -0,0 +1,228 @@
...
+let orgChanges, orgFeatures;
+let version = 1;
+
+export const name = "twilio-account-switching";
+
+export const metadata = () => ({
+  displayName: "Twilio Accounts",
+  description: "Add, delete, and edit Twilio accounts",
+  canSpendMoney: false,
+  moneySpendingOperations: ["onCampaignStart"],
```
ibrand: if `canSpendMoney` is false, then why is there a "money spending operation"?

src/extensions/service-managers/twilio-account-switching/index.js
```diff
@@ -0,0 +1,228 @@
...
+
+export async function onMessageSend({
+  message,
+  contact,
+  organization,
+  campaign
+}) {
+  const campaignTwilioAccount = _getCampaignTwilioAccount(
+    campaign,
+    ({ accountSid, authToken, id, messageServiceSids }) => ({
```
ibrand, line 25: huh, what's happening here? I see it takes in the parameters but then immediately returns them?

src/extensions/service-managers/twilio-account-switching/index.js
```diff
@@ -0,0 +1,228 @@
...
+  user,
+  loaders,
+  fromCampaignStatsPage
+}) {
+  // MUST NOT RETURN SECRETS!
+  // called both from edit and stats contexts: editMode==true for edit page
+  if (!fromCampaignStatsPage) {
+    const campaignTwilioAccount = _getCampaignTwilioAccount(
+      campaign,
+      ({ friendlyName, id }) => ({ friendlyName, id }),
```
ibrand, line 71: same question as last time I saw this structure. This is a new pattern for me

ibrand (dismissed):

src/extensions/service-managers/twilio-account-switching/index.js
crayolakat: There should not be. I removed "onCampaignStart" from moneySpendingOperations

src/extensions/service-managers/twilio-account-switching/index.js
crayolakat: That pattern is the "Object Destructuring and Property Shorthand" (https://stackoverflow.com/questions/17781472/how-to-get-a-subset-of-a-javascript-objects-properties)

`({ accountSid, authToken, id, messageServiceSids }) => ({accountSid, authToken, id, messageServiceSids})` gets passed into the `_getCampaignTwilioAccount` function as the `keys` parameter

The `_getCampaignTwilioAccount` function is defined as:
https://github.com/MoveOnOrg/Spoke/blob/fad5492179fb35d3378852bad0d1bf1d343c0980/src/extensions/service-managers/twilio-account-switching/index.js#L196-L217

`keys` defines the keys that the function will return from a Twilio account object. In this case, it will return the accountSid, authToken, id, and messageServiceSids keys from the Twilio account object. In some cases, the `_getCampaignTwilioAccount` function may be used as a helper function to return data to the front-end, such as here in `getCampaignData`: https://github.com/MoveOnOrg/Spoke/blob/fad5492179fb35d3378852bad0d1bf1d343c0980/src/extensions/service-managers/twilio-account-switching/index.js#L72-L76, so the purpose of using the "Object Destructuring and Property Shorthand" pattern is to return only the required keys from the Twilio account object, instead of all the keys

src/extensions/service-managers/twilio-account-switching/index.js
crayolakat: Same response as above

crayolakat:
@ibrand I made a few changes to address the following issues:

1. orgChanges and orgFeatures not being instantiated properly
I moved the instantiation of orgChanges and orgFeatures to getOrganizationData(...):
https://github.com/MoveOnOrg/Spoke/blob/b651d4ee074bb207805556ffb3e86c7447a581ad/src/extensions/service-managers/twilio-account-switching/index.js#L109-L113
This allows the variables to be instantiated upon Settings page load. Previously, they were in onOrganizationUpdateSignal(...), which would not freshly instantiate the variables upon Settings page load.
2. "Re-encryption" of auth token
Prior to these changes, if a user added a Twilio account, then went back to edit the account later (such as change the Friendly Name), the already encrypted auth token would be re-encrypted, thus preventing it from having the correct value when decrypting the token.
https://github.com/MoveOnOrg/Spoke/blob/b651d4ee074bb207805556ffb3e86c7447a581ad/src/extensions/service-managers/twilio-account-switching/index.js#L134-L146
3. Save button does not become enabled if the user only changes the Auth Token of a Twilio account
Prior to these changes, if a user only changes the Auth Token of a Twilio account, the Save button remains disabled and the user isn't able to save their changes
https://github.com/MoveOnOrg/Spoke/blob/b651d4ee074bb207805556ffb3e86c7447a581ad/src/extensions/service-managers/twilio-account-switching/index.js#L164-L168
4. Address failing test
Have to account for the case when the serviceManagerData object is null
https://github.com/MoveOnOrg/Spoke/blob/b651d4ee074bb207805556ffb3e86c7447a581ad/src/extensions/service-vendors/twilio/index.js#L242

src/extensions/service-managers/twilio-account-switching/index.js
ibrand: ahhhh yes that makes a lot of sense. Thank you for this explanation!

ibrand (dismissed):

crayolakat:
There were a couple variables defined at the top of `src/extensions/service-managers/twilio-account-switching/index.js` which were found to be out-of-scope in certain deployment settings. Those variables also do not follow the typical React pattern of rendering data. 012dd1886f444ac116d5f023c9f8c64885d21569 re-writes the code to remove those variables and follow the React pattern of rendering data via state changes.

src/extensions/service-managers/twilio-account-switching/index.js
```diff
@@ -0,0 +1,208 @@
...
+  for (let i = 0; i < updateData.length; i++) {
+    const curAccount = updateData[i];
+    const foundAccount = orgFeatures
+      ? orgFeatures.find(account => account.id == curAccount.id)
+      : null;
+
+    // Only encrypt auth token if it's not already encrypted
+    if (curAccount.authToken == "<Encrypted>") {
+      // Didn't change auth token. Don't re-encrypt
+      curAccount.authToken = foundAccount.authToken;
```
ibrand, line 133: Nice!

ibrand (approved):

PR #2111, opened 2022-01-15
Add isOptedOut field to allContactsCount and set to false

### LINKED ISSUE #2003: Bug: Ghost Replies / Contacts
**Describe the bug**
When 1 contact is in 2 campaigns and is marked as opt out in 1 of the campaigns the other campaign does not properly exclude that contact from the total count of contacts. This means the "Initials" box can never be 100% which is very confusing.

Also if a reply link is set to give out 10 replies but 5 people are opted out it will only give 5 replies to a person.

**To Reproduce**
1. Set up 2 campaigns with the same contact in both and a few other random contacts.
2. In the first campaign start the convo and opt them out.
3. Then get initials from the next campaign and send them all
4. Observe how the campaign never hit 100%

**Expected behavior**
Total Contacts should always represent the number of people who are textable. An op-out from the first campaign should drop the total count from other campaigns.

if a reply link is supposed to give out 10 replies it should find 10 people who can be texted.

n0land0:
# Fixes #2003 

## Description

- Previously, when a texter had a contact listed which included contacts who had opted out from another campaign, those opt-outs were not being excluded from that texter's number of remaining texts to send
- `src/containers/TexterTodo.jsx`: Add `isOptedOut` field to `allContactsCount` and set value to `false`
- Test to ensure opt-out status persists across campaigns when creating a new campaign
- Test to ensure that, if same contacts are loaded to two new campaigns simultaneously, the second campaign's total contacts update when a shared contact opts out of the first campaign

# Checklist:

### DIFF
--- a/src/containers/TexterTodo.jsx
+++ b/src/containers/TexterTodo.jsx
@@ -157,7 +157,7 @@ export const dataQuery = gql`
       contacts(contactsFilter: $contactsFilter) {
         id
       }
-      allContactsCount: contactsCount
+      allContactsCount: contactsCount(contactsFilter: { isOptedOut: false })
       unmessagedCount: contactsCount(contactsFilter: $needsMessageFilter)
       unrepliedCount: contactsCount(contactsFilter: $needsResponseFilter)
       texter {


ibrand (approved):
Thank you for checking those boxes, testing your changes, and a very clear description!

crayolakat (changes requested):
`basic-campaign-e2e.test.js` Cypress test is now failing. Context below:

To run the `basic-campaign-e2e.test.js` Cypress test, following the instructions below:

1. Run this command: `yarn start:test`
2. In a new terminal, run this command: `yarn test-cypress`
3. In the Cypress window that pops up, click the `basic-campaign-e2e.test.js` test to run it
4. You will see the following failure:
image
5. The failure is caused by `allContactsCount` returning the incorrect value of 0 instead of the correct value of 2

Debugging notes:
Before this change, the following code was executed in the `contactsCount` function to get `allContactsCount`:
https://github.com/MoveOnOrg/Spoke/blob/ce5d1290a5c4ef3ea0a7d105dc839a1917545df1/src/server/api/assignment.js#L309-L315
This `filterCount` function returned the correct count.

After this change, the following code is executed in the `contactsCount` function to get `allContactsCount`:
https://github.com/MoveOnOrg/Spoke/blob/ce5d1290a5c4ef3ea0a7d105dc839a1917545df1/src/server/api/assignment.js#L334-L338
This `getContacts` function returns the wrong count. This is due to the following lines of code in the `getContacts` function:
https://github.com/MoveOnOrg/Spoke/blob/ce5d1290a5c4ef3ea0a7d105dc839a1917545df1/src/server/api/assignment.js#L132-L140
These lines of code cause `allContactsCount` be always be filtered by either the `needsResponse` or `needsMessageOrResponse` messageStatus. There doesn't seem to be a way to get all contacts regardless of message status via the `getContacts` function. We need to figure out why this is so, whether it's the desired behavior, and develop a solution this failing test.

Note: @ibrand and I are currently looking into this issue

PR #2113, opened 2022-01-18
automatic TexterTodoList updates (when we have redis+postgres) (Fixes #207)

### LINKED ISSUE #207: Auto-refresh replies and other ToDo-screen data

schuyler1d:
# Fixes #207 

## Description

For a long time, we've wanted the TodoList to refresh automatically. This PR creates the backend infrastructure and a stub implementation on the frontend for auto-updating data. Future work can do browser notifications, e.g.

Depending on Redis and Postgres, the important thing is to make sure polling does not add any database calls which wouldn't scale if we have 100s or 1000s of texters. upon an incoming message update that yields a messageStatus of 'needsResponse' we get the assignmentId and add it to a new cache key that's specific to the user. 

Note that github is saying this is 500 changed lines, but several of them are movements that are registering as adds AND removes.

# Checklist:

### DIFF
--- a/__test__/server/texter.test/texter.test.js
+++ b/__test__/server/texter.test/texter.test.js
@@ -156,11 +156,12 @@ it("should be able to receive a response and reply (using fakeService)", async (
       expect.objectContaining({
         send_status: "DELIVERED",
         text: `responding to ${message.text}`,
-        user_id: testTexterUser.id,
+        user_id: null,
         contact_number: testContact.cell,
         campaign_contact_id: testContact.id
       })
     );
+    expect(dbMessage[1].is_from_contact).toBeTruthy();
   });
 
   await waitForExpect(async () => {
--- a/src/api/user.js
+++ b/src/api/user.js
@@ -16,7 +16,8 @@ export const schema = `
     assignment(campaignId: String): Assignment
     terms: Boolean
     profileComplete(organizationId: String): Boolean
-    cacheable: Boolean
+    notifications(organizationId: String): [Assignment]
+    notifiable: Boolean
   }
 
 type UsersList {
--- a/src/containers/TexterTodoList.jsx
+++ b/src/containers/TexterTodoList.jsx
@@ -4,16 +4,57 @@ import CheckCircleIcon from "@material-ui/icons/CheckCircle";
 import Empty from "../components/Empty";
 import LoadingIndicator from "../components/LoadingIndicator";
 import AssignmentSummary from "../components/AssignmentSummary";
+import Snackbar from "@material-ui/core/Snackbar";
 import loadData from "./hoc/load-data";
 import gql from "graphql-tag";
 import { withRouter } from "react-router";
 
 let refreshOnReturn = false;
 
+// We want to back off on notifications both because if they don't leave the page
+// then there's no point in polling more frequently, and also to stop annoying
+// someone that's clearly ignoring it.
+// So 30sec, 2min, 8min, 32min, 2 hours -- so 4x each time
+const NOTIFICATION_START_DELAY = 7500;
+const NOTIFICATION_MAX_DELAY = 7680000; // 2.13 hours
+let notificationPollDelay = NOTIFICATION_START_DELAY;
+
 class TexterTodoList extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { polling: null };
+    this.state = {};
+  }
+
+  componentWillUpdate(nextProps, nextState) {
+    const notifications =
+      nextProps.notifications && nextProps.notifications.user.notifications;
+    if (notifications && notifications.length) {
+      // Note, we are changing notification polling times only when there's a notification,
+      // otherwise keep polling at the same rate.
+      notificationPollDelay = Math.min(
+        4 * notificationPollDelay,
+        NOTIFICATION_MAX_DELAY
+      );
+      this.props.notifications.stopPolling();
+      this.props.notifications.startPolling(notificationPollDelay);
+      // move the result to state
+      nextProps.notifications.user.notifications = [];
+      // FUTURE: maybe append for a set of assignmentIds to display them
+      nextState.notifications = notifications;
+    }
+  }
+
+  componentDidMount() {
+    if (refreshOnReturn) {
+      this.props.data.refetch();
+    }
+    // reset notification state
+    notificationPollDelay = NOTIFICATION_START_DELAY;
+  }
+
+  componentWillUnmount() {
+    // not state: maintain this forever after
+    refreshOnReturn = true;
   }
 
   renderTodoList(assignments) {
@@ -56,34 +97,6 @@ class TexterTodoList extends React.Component {
       })
       .filter(ele => ele !== null);
   }
-  componentDidMount() {
-    if (refreshOnReturn) {
-      this.props.data.refetch();
-    }
-    // stopPolling is broken (at least in currently used version), so we roll our own so we can unmount correctly
-    if (
-      this.props.data &&
-      this.props.data.user &&
-      this.props.data.user.cacheable &&
-      !this.state.polling
-    ) {
-      const self = this;
-      this.setState({
-        polling: setInterval(() => {
-          self.props.data.refetch();
-        }, 5000)
-      });
-    }
-  }
-
-  componentWillUnmount() {
-    if (this.state.polling) {
-      clearInterval(this.state.polling);
-      this.setState({ polling: null });
-    }
-    // not state: maintain this forever after
-    refreshOnReturn = true;
-  }
 
   termsAgreed() {
     const { data, router } = this.props;
@@ -118,7 +131,18 @@ class TexterTodoList extends React.Component {
       <Empty title="You have nothing to do!" icon={<CheckCircleIcon />} />
     );
 
-    return <div>{renderedTodos.length === 0 ? empty : renderedTodos}</div>;
+    return (
+      <div>
+        <Snackbar
+          open={Boolean(this.state.notifications)}
+          message={"Some campaigns have replies for you to respond to!"}
+          onClose={() => {
+            this.setState({ notifications: false });
+          }}
+        />
+        {renderedTodos.length === 0 ? empty : renderedTodos}
+      </div>
+    );
   }
 }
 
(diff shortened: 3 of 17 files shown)


__test__/server/texter.test/texter.test.js
```diff
@@ -156,11 +156,12 @@ it("should be able to receive a response and reply (using fakeService)", async (
       expect.objectContaining({
         send_status: "DELIVERED",
         text: `responding to ${message.text}`,
-        user_id: testTexterUser.id,
+        user_id: null,
```
schuyler1d, line 159: while adding tests, this bug showed up. This is_from_contact=true shouldn't have a user_id since it's allegedly sent from the contact.

__test__/server/texter.test/texter.test.js
```diff
@@ -156,11 +156,12 @@ it("should be able to receive a response and reply (using fakeService)", async (
...
         contact_number: testContact.cell,
         campaign_contact_id: testContact.id
       })
     );
+    expect(dbMessage[1].is_from_contact).toBeTruthy();
```
schuyler1d, line 164: I could add this as =true above, but sqlite annoyingly returns as `1` instead, so we need to be truthy here.

src/api/user.js
```diff
@@ -16,7 +16,8 @@ export const schema = `
     assignment(campaignId: String): Assignment
     terms: Boolean
     profileComplete(organizationId: String): Boolean
-    cacheable: Boolean
+    notifications(organizationId: String): [Assignment]
+    notifiable: Boolean
```
schuyler1d, line 20: cacheable was always a placeholder for notifiable, and when it depended on postgres returning() as well, it seemed like we should name it something more precise.
`notifications` is the polled api call that can return the updated assignments with the newer data.

src/containers/TexterTodoList.jsx
```diff
@@ -118,7 +131,18 @@ class TexterTodoList extends React.Component {
       <Empty title="You have nothing to do!" icon={<CheckCircleIcon />} />
     );
 
-    return <div>{renderedTodos.length === 0 ? empty : renderedTodos}</div>;
+    return (
```
schuyler1d, line 134: Snackbar is a stub notification -- once we get this in, we could use maybe https://www.npmjs.com/package/react-browser-notifications to do browser notifications

src/containers/TexterTodoList.jsx
```diff
@@ -4,16 +4,57 @@ import CheckCircleIcon from "@material-ui/icons/CheckCircle";
...
 import LoadingIndicator from "../components/LoadingIndicator";
 import AssignmentSummary from "../components/AssignmentSummary";
+import Snackbar from "@material-ui/core/Snackbar";
 import loadData from "./hoc/load-data";
 import gql from "graphql-tag";
 import { withRouter } from "react-router";
 
 let refreshOnReturn = false;
 
+// We want to back off on notifications both because if they don't leave the page
```
schuyler1d, line 14: This file, TexterTodoList is where all the action is on the client-side.  The changes do the following:
* separate the fields needed for an assignment record, so we can poll the same fields for notifications
* hard-code all the stupid count parameters so it's easy to embed them rather than pass them as variables
* back off notifications as a user continues to ignore them (if they click into a page and come back, it resets)

src/containers/TexterTodoList.jsx
```diff
@@ -56,34 +97,6 @@ class TexterTodoList extends React.Component {
       })
       .filter(ele => ele !== null);
   }
-  componentDidMount() {
-    if (refreshOnReturn) {
-      this.props.data.refetch();
-    }
-    // stopPolling is broken (at least in currently used version), so we roll our own so we can unmount correctly
```
schuyler1d, line 63: This is all dead code that was a stub for if/when we implemented this feature

src/containers/TexterTodoList.jsx
```diff
@@ -178,16 +184,55 @@ export const dataQuery = gql`
           sweepComplete
         }
         allContactsCount: contactsCount
-        unmessagedCount: contactsCount(contactsFilter: $needsMessageFilter)
-        unrepliedCount: contactsCount(contactsFilter: $needsResponseFilter)
-        badTimezoneCount: contactsCount(contactsFilter: $badTimezoneFilter)
+        unmessagedCount: contactsCount(contactsFilter: {
```
schuyler1d, line 187: these are all moving constants in the variables{} section below to actual constant within the gql, which I think is more readable anyway.

src/containers/TexterTodoList.jsx
```diff
@@ -226,38 +271,42 @@ const queries = {
...
+      ) {
+        user(organizationId: $organizationId, userId: $userId) {
+          id
+          notifications {
+            ${assignmentQueryData}
+          }
+        }
+      }
+    `,
+    options: ownProps => {
```
schuyler1d, line 292: this gets re-run when `data` property is updated, so we can add conditions for when they come in and then not poll at all in scenarios where e.g. there aren't any todos

src/server/api/message.js
```diff
@@ -4,20 +4,14 @@ import { Message } from "../models";
...
+      ["text", "userNumber", "contactNumber", "createdAt", "isFromContact"],
       Message
     ),
     media: msg =>
       // Sometimes it's array, sometimes string. Maybe db vs. cache?
       typeof msg.media === "string" ? JSON.parse(msg.media) : msg.media || [],
     // cached messages don't have message.id -- why bother
-    id: msg => msg.id || `fake${Math.random()}`
+    id: msg => msg.id || `fake${Math.random()}`,
+    userId: msg => msg.user_id || null
```
schuyler1d, line 15: this was a bug I encountered while testing, where if userId was undefined the message view failed to load.

src/server/api/user.js
```diff
@@ -174,6 +174,129 @@ export async function getUsers(
   }
 }
 
+export const getUserTodos = async (
```
schuyler1d, line 177: this is completely pulling out the `todos` function into a new function, `getUserTodos` for a function and adding `queryFilter` as an argument, so we can use the same code for `notifications` where we pass queryFilter to be only the assignmentIds that were notifified.

src/server/api/user.js
```diff
@@ -250,119 +373,21 @@ export const resolvers = {
...
+      user,
+      { organizationId, withOutCounts },
+      { user: loggedInUser }
+    ) => {
+      const assignments = await getUserTodos(
+        user,
+        { organizationId, withOutCounts },
+        { user: loggedInUser }
+      );
+      if (user.id === loggedInUser.id) {
```
schuyler1d, line 386: if we call this, we should clear any notifications, since the user is getting the newest data.

src/server/models/cacheable_queries/campaign-contact.js
```diff
@@ -494,15 +496,28 @@ const campaignContactCache = {
   ) => {
     // console.log('updateSTATUS', newStatus, contact)
     try {
-      await r
+      const assignmentIds = await r
```
schuyler1d, line 499: This is crucially where we create a notification when a message comes in. In order to avoid any new SQL calls, we need to get the assignmentId returned when we do the campaign-contact update.  Before this moment, during an incoming message, we never know the assignmentId, and we can't cache it because reassignments are too common/easy.

If we do have an assignmentId, then we'll look up the assignment in cache and get the user_id that we should save the cache.  The lookup of user_id from the assignment could, in theory, use the database, but assignments should be cached and stay there during a campaign's lifetime.

src/server/models/cacheable_queries/user.js
```diff
@@ -20,8 +20,8 @@ key = orgId
 value = highest_role:org_name
 
 QUERYS:
-userHasRole(userId, orgId, acceptableRoles) -> boolean
-userLoggedIn(authId) -> user object
+userHasRole(user, orgId, acceptableRoles) -> boolean
```
schuyler1d, line 23: outdated header doc that I'm updating while I added tests.

src/server/models/cacheable_queries/user.js
```diff
@@ -30,6 +30,8 @@ const userRoleKey = userId =>
   `${process.env.CACHE_PREFIX || ""}texterroles-${userId}`;
 const userAuthKey = authId =>
   `${process.env.CACHE_PREFIX || ""}texterauth-${authId}`;
+const notificationsKey = userId =>
```
schuyler1d, line 33: Here's the new notifiation key by-user

src/server/models/cacheable_queries/user.js
```diff
@@ -97,7 +99,11 @@ const loadUserRoles = async userId => {
       const userRoles = {};
       Object.keys(roles).forEach(orgId => {
         const [highestRole, orgName] = roles[orgId].split(":");
-        userRoles[orgId] = { id: orgId, name: orgName, role: highestRole };
+        userRoles[orgId] = {
+          id: Number(orgId),
```
schuyler1d, line 103: while writing tests, this would be a string or a number depending on whether it came from the db or cache, so I'm making it explicit here to be consistent

src/server/models/cacheable_queries/user.js
```diff
@@ -212,11 +218,41 @@ const userLoggedIn = async (field, val) => {
   return user;
 };
 
+const getAndClearNotifications = async userId => {
```
schuyler1d, line 221: getAndClearNotifications and addNotification are the crucial part of the backend.
While unusual, for these calls, we delete while we read, to avoid any 'old' notifications.  Besides making sure it's the user themselves that are doing the retrieving, this should stop any stale notifications from repeating.

schuyler1d (commented):
Adding some comments for easier review

crayolakat (approved):

PR #2118, opened 2022-01-25
Two features to reduce SMS/MMS costs

schuyler1d:
1. Expands characters to convert for cheaper GSM encoding; 
2. Also adds a service-manager: mms-when-cheaper that tries to force MMS when it would be cheaper compared with multiple SMS segments (currently depends on a twilio beta feature)

## Description

Some ad-hoc conversions for smart quotes were already in the code, but Twilio converts 200+ characters -- this does it on the spoke side to be more reliable (and for other vendors).

There is also a frustrating case that e.g. in other languages that don't use the ASCII English alphabet (even Spanish), accents can increase the expense of a message sent of SMS by multiplying 6+ segments.  Standard twilio (and other) rates, MMS is about the cost of 2-3 SMS segments, so it's often cheaper to send an MMS.  MMS does have different rate limits, so as long as it's done consciously (i.e. enabled on-purpose) this should be an improvement.  For Twilio, the `sendAsMms` flag is in beta, so this PR is in draft until I've at least confirmed it works using the beta.

# Checklist:

- [see below] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#your-first-code-contribution), or has a documented reason in the description why it’s longer

Note: the changes are above 300, but only because the full list of characters is 200+ lines of 'data' rather than code.


crayolakat (approved):

joemcl:
This is a great idea, thanks Sky for doing this

PR #2119, opened 2022-01-25
Fix JS frontend error when TZ env var is set

michael-gratton:
Fixes a fatal frontend error on initial page load when the `TZ` environment var is set to a non-builtin value, e.g. `Australia/Sydney`.

Without this, there's an error on initial page load that prevents the UI from showing up in 11.1.

Does this need tests with the TZ var set, or that var added to `REFERENCE-environment_variables.md`?

### DIFF
--- a/src/components/CampaignTextingHoursForm.jsx
+++ b/src/components/CampaignTextingHoursForm.jsx
@@ -8,6 +8,7 @@ import * as yup from "yup";
 import cloneDeep from "lodash/cloneDeep";
 import isEqual from "lodash/isEqual";
 import moment from "moment";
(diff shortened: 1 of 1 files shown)


crayolakat (approved):
Thanks for catching this! Ideally yes, the `TZ` variable should be added to `REFERENCE-environment_variables.md`, and tests should be written with the `TZ` variable set.

Since we are aiming for this PR to be in Spoke release 12, those 2 tasks can be done in a separate PR. I created a GitHub issue to keep track of that here: https://github.com/MoveOnOrg/Spoke/issues/2129

PR #2107, opened 2022-01-07
Stop early zip code addition

n0land0:
# Fixes #1062 

## Description

Delay seeding zip codes until database is created, so that zip codes can be counted

# Checklist:

- [x] I have manually tested my changes on desktop and mobile
- [x] The test suite passes locally with my changes
- [ ] If my change is a UI change, I have attached a screenshot to the description section of this pull request
- [x] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#your-first-code-contribution), or has a documented reason in the description why it’s longer
- [ ] I have made any necessary changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] My PR is labeled [WIP] if it is in progress

PR #2131, opened 2022-03-12
Eliminate exception when starting up against a new database

### LINKED ISSUE #1062: Improve seeding & database initialization
When starting from a fresh install, it's currently necessary to start Spoke and let it fail on attempting to seed zip codes (because the zip code table doesn't exist yet). After that failure, it goes on to create the tables, but doesn't retry seeding. Thus, one has to kill the process and start it again to seed the zip codes and actually run the application.

Unless there's a reason for the current behavior, this should be changed to:
1) Run migrations (#692 introduces a new process for this),
2) Check if seeding is necessary and seed if so.

The relevant code is in `/src/server/index.js` around L43.

lperson:
# Fixes #1062 

## Description

Eliminate the `zip_code table doesn't exist` exception when starting Spoke against a newly created database. 

Previously the code checked whether to seed the zip_code table without regard for whether the database existed. The check involves querying the zip_code table. If it doesn't exist that causes an exception. This change puts some guards around seeding the zip_code table so that the code doesn't attempt it if it's also creating a new database.

# Checklist:

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -15,7 +15,7 @@ import { log } from "../lib";
 import telemetry from "./telemetry";
 import { addServerEndpoints as messagingServicesAddServerEndpoints } from "../extensions/service-vendors/service_map";
 import { getConfig } from "./api/lib/config";
-import { seedZipCodes } from "./seeds/seed-zip-codes";
+import { seedZipCodesIfNecessary } from "./seeds/seed-zip-codes";
 import { setupUserNotificationObservers } from "./notifications";
 import { existsSync } from "fs";
 import { rawAllMethods } from "../extensions/contact-loaders";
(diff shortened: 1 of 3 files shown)


schuyler1d:
I think this was fixed in this PR:
https://github.com/MoveOnOrg/Spoke/pull/2107/files
but it looks like it wasn't included in the 12.0 release candidate. @crayolakat FYI
Not saying which one is better, but we should fix it so people don't keep on hitting it.

lperson:
> I think this was fixed in this PR:
> https://github.com/MoveOnOrg/Spoke/pull/2107/files
> but it looks like it wasn't included in the 12.0 release candidate. @crayolakat FYI
> Not saying which one is better, but we should fix it so people don't keep on hitting it.

@schuyler1d  It's essentially the same fix!

lperson:
Closed in favor of #2107

PR #2133, opened 2022-03-13
2082 stop overwriting campaign edits

### LINKED ISSUE #2082: Bug: Campaign details refresh every minute with data from the server and overwrites edits
**Describe the bug**
The campaignData query is polled every 1 minute to get the latest changes on the server: https://github.com/MoveOnOrg/Spoke/blob/06f743fff9d9764a5f900ee561eccda9ecd82dc2/src/containers/AdminCampaignEdit.jsx#L1068

After the campaignData query is polled, it triggers this React method which updates all the sections except the currently expanded section with the latest changes on the server: https://github.com/MoveOnOrg/Spoke/blob/06f743fff9d9764a5f900ee561eccda9ecd82dc2/src/containers/AdminCampaignEdit.jsx#L196

The problem is that if someone is creating/editing a campaign and doesn’t specifically click “Save” every time they collapse a section and expand a different one, the changes in the section that they didn’t click “Save” in will be lost. The code was purposely written to work that way, although multiple people have expressed annoyance and significant loss of time with their unsaved changes getting wiped out every minute if they don't save the section before moving on to the next one.

This is my proposed approach to the issue: Don’t overwrite form fields that someone has changed, even if the section is collapsed and not saved. Note that Spoke doesn't have an autosave feature, so users must be aware that they still have to go back to the collapsed section and click "Save." There is already a UI indication that a section hasn't been saved because the section header color will be yellow.

**To Reproduce**
I don't know how to reproduce this issue locally, but here are the steps to reproduce the behavior on an instance that's deployed on Lambda with the env variable `JOB_RUNNER` set to `lambda-async`
1. Log in as an admin
2. Navigate to an existing campaign and click `Edit`
3. Expand the `Basics` section
4. Change the value of `Intro HTML`. **Don't** click `SAVE`
5. Expand the `Contacts` section
6. Wait 1 minute
7. **Issue:** The changes that you made in the `Basics` section will be wiped out

**Expected behavior**
The edits that you make to any form fields shouldn't be wiped out unless you leave the page without saving. The form fields that you didn't edit should continue to be updated with the latest data from the server every 1 minute.

 - OS: macOS
 - Browser: Chrome
 - Desktop
 - Version: 95.0.4638.69

lperson:
# Fixes #2082 

### Change that fixes the issue
In `AdminCampaignEdit` we poll for updated data by design. 

Before, the polling would result in unsaved user changes in closed sections being overwritten. We prevented data from being updated in the open section.

This PR makes it so that updates do not overwrite changed data in any section in which the user changed something and closed the section without saving. 

### Additional change

When we had pending jobs, we changed the polling interval from 60 seconds to 2.5 seconds. When there were no more pending jobs, rather than setting the polling interval back to 60 seconds, we stopped polling altogether. This PR sets the polling inteval back to 60 seconds.

# Checklist:

### DIFF
--- a/src/containers/AdminCampaignEdit.jsx
+++ b/src/containers/AdminCampaignEdit.jsx
@@ -41,6 +41,10 @@ import { styles } from "./AdminCampaignStats";
 import AdminScriptImport from "../containers/AdminScriptImport";
 import { makeTree } from "../lib";
 
+const CAMPAIGN_POLLING_INTERVAL = 60000;
+const JOB_WAITING_POLLING_INTERVAL = 2500;
+const ORG_POLLING_INTERVAL = 20000;
+
 const campaignInfoFragment = `
   id
   title
@@ -168,7 +172,7 @@ export class AdminCampaignEdit extends React.Component {
         {
           isPolling: true
         },
-        () => this.props.campaignData.startPolling(2500)
+        () => this.props.campaignData.startPolling(JOB_WAITING_POLLING_INTERVAL)
       );
     }
   };
@@ -181,7 +185,7 @@ export class AdminCampaignEdit extends React.Component {
           isPolling: false
         },
         () => {
-          this.props.campaignData.stopPolling();
+          this.props.campaignData.startPolling(CAMPAIGN_POLLING_INTERVAL);
         }
       );
     }
@@ -1065,7 +1069,7 @@ const queries = {
       variables: {
         campaignId: ownProps.params.campaignId
       },
-      pollInterval: 60000
+      pollInterval: CAMPAIGN_POLLING_INTERVAL
     })
   },
   organizationData: {
@@ -1106,7 +1110,7 @@ const queries = {
       variables: {
         organizationId: ownProps.params.organizationId
       },
-      pollInterval: 20000
+      pollInterval: ORG_POLLING_INTERVAL
     })
   }
 };
--- a/src/containers/AdminCampaignEdit.jsx
+++ b/src/containers/AdminCampaignEdit.jsx
@@ -213,18 +213,29 @@ export class AdminCampaignEdit extends React.Component {
       expandedKeys = expandedSection.keys;
     }
 
-    const unsavedKeys = [];
+    const keysInUnsavedSections = [];
     this.sections().forEach(section => {
       if (!this.checkSectionSaved(section)) {
-        unsavedKeys.push(...section.keys);
+        keysInUnsavedSections.push(...section.keys);
       }
     });
+
+    const newCampaignData = newProps.campaignData.campaign;
+    const changedKeysInUnsavedSections = [];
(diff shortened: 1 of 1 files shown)


src/containers/AdminCampaignEdit.jsx
```diff
@@ -202,17 +206,37 @@ export class AdminCampaignEdit extends React.Component {
...
 
+    const keysInUnsavedSections = [];
+    this.sections().forEach(section => {
+      if (!this.checkSectionSaved(section)) {
+        keysInUnsavedSections.push(...section.keys);
+      }
+    });
+
+    const newCampaignData = newProps.campaignData.campaign;
+    const changedKeysInUnsavedSections = [];
```
schuyler1d: are these *changed keys* or just keys that we don't want to overwrite due to unsaved info?  If the latter, then let's change the variable name.

src/containers/AdminCampaignEdit.jsx
```diff
@@ -202,17 +206,37 @@ export class AdminCampaignEdit extends React.Component {
...
+    keysInUnsavedSections.forEach(key => {
+      if (newCampaignData[key] !== this.state.campaignFormValues[key]) {
```
schuyler1d: Do we need a deep equality comparison here?  Didn't new data come in from the server?  Since we already have lodash in the codebase, maybe use: `_.isEqual()`.
Regardless, let's add a comment for line 226 to explain what we are and are not updating/testing.

schuyler1d (commented):
Need a little help parsing what's happening in this PR, but this is super exciting if this fixes it (or even mitigates the issue significantly). :tada:

src/containers/AdminCampaignEdit.jsx
lperson: @schuyler1d No, we're going key by key to find keys that were changed by the user. If a key was changed by the user we remove it from the update from the server.

src/containers/AdminCampaignEdit.jsx
lperson: Added some comments and renamed the variable. Hopefully this makes my intention clear(er).

schuyler1d: what are the keys here?  Aren't a bunch of them objects and not values (e.g. Interaction Steps array, which is one of the frequent issues)

src/containers/AdminCampaignEdit.jsx
schuyler1d: ok, so this was changed by the user, not by the server.  Should we also see if the original version is different than the new server version to see if the user might be overwriting a change? (Merging is probably WAY too complex, etc, but we might at least want to warn folks).

Let's change the var names so we say "userChanged..." or "serverChanged...."

crayolakat:
Hey @lperson ! I'm wondering if you were able to address the questions that @schuyler1d had? Thanks for this very useful PR

PR #2134, opened 2022-03-17
Update the 300 lines of code link to point more directly to the right section on it

ibrand:
# Description

The old link links to a section header a little high up that doesn't really get to the "300 lines of code or less" piece

# Checklist:


crayolakat (approved):
Thanks for catching this!

PR #2137, opened 2022-03-18
updated SUPRESS_MIGRATIONS=false

joemcl:
updated v 11 migrations to what actually works :) on Heroku

# Fixes # (issue)
fixes an inaccuracy in docs for version 11, will update later for earlier versions
## Description

# Checklist:

### DIFF
--- a/docs/RELEASE_NOTES.md
+++ b/docs/RELEASE_NOTES.md
@@ -84,7 +84,7 @@ Based on work from Larry Person, there is a large refactor of "service-vendors" 
 #### Database upgrades
 * This is a major release and includes a schema change.
 * For small instances simply leave/set SUPPRESS_MIGRATIONS="" or for [AWS Lambda, see the db migration instructions](https://moveonorg.github.io/Spoke/#/HOWTO_DEPLOYING_AWS_LAMBDA?id=migrating-the-database).
-* If you are running Spoke on Heroku and updating to version 11, either remove SUPPRESS_MIGRATIONS="" if it is set in config, *or* set SUPPRESS_MIGRATIONS=false
+* If you are running Spoke on Heroku and updating to version 11 or later, either remove SUPPRESS_MIGRATIONS="" if it is set in config, *or* set SUPPRESS_MIGRATIONS=false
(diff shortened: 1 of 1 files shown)


docs/RELEASE_NOTES.md
```diff
@@ -83,7 +83,8 @@ Based on work from Larry Person, there is a large refactor of "service-vendors"
 
 #### Database upgrades
 * This is a major release and includes a schema change.
-* For small instances simply leave/set SUPPRESS_MIGRATIONS="" or for [AWS Lambda, see the db migration instructions](https://moveonorg.github.io/Spoke/#/HOWTO_DEPLOYING_AWS_LAMBDA?id=migrating-the-database). 
+* For small instances simply leave/set SUPPRESS_MIGRATIONS="" or for [AWS Lambda, see the db migration instructions](https://moveonorg.github.io/Spoke/#/HOWTO_DEPLOYING_AWS_LAMBDA?id=migrating-the-database).
+* If you are running Spoke on Heroku and updating to version 11, either remove SUPPRESS_MIGRATIONS="" if it is set in config, *or* set SUPPRESS_MIGRATIONS=false
```
Arique1104: Thank you for catching this @joemcl!

Arique1104 (dismissed):
Thanks for catching this error in our documentation!

docs/RELEASE_NOTES.md
crayolakat: I believe this same issue would occur in releases after Spoke 11 as well. Can we say "version 11 or later" instead of "version 11"?

```suggestion
* If you are running Spoke on Heroku and updating to version 11 or later, either remove SUPPRESS_MIGRATIONS="" if it is set in config, *or* set SUPPRESS_MIGRATIONS=false
```

crayolakat (changes requested):

docs/RELEASE_NOTES.md
joemcl: The same issue would occur on every version before 11 as well :) I just didn't update the previous entries to reflect that.

crayolakat (approved):

PR #2147, opened 2022-04-05
Auth0 Password Reset - bug

### LINKED ISSUE #1870: Password reset/change with Auth0
**Describe the bug**
The password reset button on the Users list, and change password button on the user edit screen are intended to work for local auth. If you are using the Auth0 passport strategy, they do bad things:

- The change password dialog gives "GraphQL error: undefined" which is confusing for users.
- The Reset Password button is worse. If an admin clicks the button, the user's Auth0 token is replaced in the database with a reset hash, and they will NEVER be able to log in with that account again. If the user does manage to log in with Auth0 (presumably by using A0's password reset) it will create a duplicate account in the Spoke db (same email, name, etc, new user id).

I see that @lperson worked on #1155, but it appears it was closed without merging. Probably a good place to start for working on this issue!

### LINKED ISSUE #2155: Bug: Unable to change Auth0 Email
**Describe the bug**
When the user changes their email in their Spoke profile settings, it won't be updated in Auth0 if the Auth0 passport strategy is enabled. The user would still have to log into Spoke using their old email address, and won't be able to use the "Don't remember your password" link if they forgot their password. Currently, only admins who have access to Auth0 can update a user's email. It would be better if when someone updates their own email in their Spoke profile settings, this update is also reflected in Auth0 and their login credentials.

**To Reproduce**
1. Log in as a user via Auth0
2. Go to the user profile page and change your email
3. Log out
4. Issue 1: If the user tries logging in again they must log in using their old email and password. The new email isn't updated in Auth0
5. Issue 2: If the user clicks "Don't remember your password" at the login screen, a reset password email will be sent to the old email instead of the new email. This doesn't help in the case when someone changed their Spoke email because they lost access to their old email

**Expected behavior**
When the user changes their email in their Spoke profile settings, it should also be updated in Auth0.

**Screenshots**
User profile page
image

Auth0 login screen
image

Arique1104:
# Fixes #1870 

## Description

- Successful backend deployment of a password reset for an Auth0 Administrator to reset other users passwords.
    - [Explanation of Auth0 API call](https://gist.github.com/Arique1104/19f761c07cac7f53ce5e18128a91832c)

- An Auth0 Administrator can now safely reset their user's password, and the user will receive an email from Auth0 to reset their password.
Screen Shot 2022-04-07 at 7 59 00 PM

- Dialog confirmation that an email was sent shows up for an Auth0 Administrator
Screen Shot 2022-04-07 at 7 59 25 PM

- Removal of "Change Password" button at the user profile level, if PASSPORT_STRATEGY is set to auth0
Screen Shot 2022-04-07 at 7 56 43 PM

# Checklist:

### DIFF
--- a/src/containers/PeopleList.jsx
+++ b/src/containers/PeopleList.jsx
@@ -165,25 +165,16 @@ export class PeopleList extends Component {
   resetPassword = async userId => {
     const { currentUser } = this.props;
     if (window.PASSPORT_STRATEGY === "auth0") {
-      let newUrl =
-        "https://manage.auth0.com/dashboard/us/" +
-        window.AUTH0_DOMAIN_MANAGEMENT;
+      let auth0Domain = "/" + window.AUTH0_DOMAIN;
+      let result = auth0Domain.split(".")[0] + "/users";
+      let country =
+        "https://manage.auth0.com/dashboard/" + auth0Domain.split(".")[1];
 
-      // notes for getting rid of AUTH0_DOMAIN_MANAGEMENT from the .env file
+      let newUrl = country + result;
 
-      //  call AUTH0_DOMAIN
+      // see if I can get ""/users" to show up at the end of that URL for better ease to Campaign Manager
 
-      // catch all the letters before the first period
-
-      // example: THIS-INFO.us.auth0.com
-
-      // add it to a local variable called auth0Domain
-
-      // add auth0Domain appropriately to the following url: "https://manage.auth0.com/dashboard/us/${auth0Domain}
-
-      // see if I can get /users to show up at the end of that URL for better ease to Campaign Manager
-
-      // needs information or a message letting them know they are being redirected to Auth0 for password management support
+      // needs information or a message letting administrator know they are being redirected to Auth0 for password management
 
       location.href = newUrl;
     } else if (currentUser.id !== userId) {
--- a/src/containers/PeopleList.jsx
+++ b/src/containers/PeopleList.jsx
@@ -129,21 +129,19 @@ export class PeopleList extends Component {
         }
       }
     ];
-    if (window.PASSPORT_STRATEGY !== "slack") {
-      columns.push({
-        key: "password",
-        name: "",
-        options: {
-          customBodyRender: this.renderChangePasswordButton
-        },
-        label: "",
-        style: {
-          textOverflow: "ellipsis",
-          overflow: "hidden",
-          whiteSpace: "pre-line"
-        }
-      });
-    }
+    columns.push({
+      key: "password",
+      name: "",
+      options: {
+        customBodyRender: this.renderChangePasswordButton
+      },
+      label: "",
+      style: {
+        textOverflow: "ellipsis",
+        overflow: "hidden",
+        whiteSpace: "pre-line"
+      }
+    });
     return columns;
   };
 
@@ -164,20 +162,7 @@ export class PeopleList extends Component {
 
   resetPassword = async userId => {
     const { currentUser } = this.props;
-    if (window.PASSPORT_STRATEGY === "auth0") {
-      let auth0Domain = "/" + window.AUTH0_DOMAIN;
-      let result = auth0Domain.split(".")[0] + "/users";
-      let country =
-        "https://manage.auth0.com/dashboard/" + auth0Domain.split(".")[1];
-
-      let newUrl = country + result;
-
-      // see if I can get ""/users" to show up at the end of that URL for better ease to Campaign Manager
-
-      // needs information or a message letting administrator know they are being redirected to Auth0 for password management
-
-      location.href = newUrl;
-    } else if (currentUser.id !== userId) {
+    if (currentUser.id !== userId) {
       const res = await this.props.mutations.resetUserPassword(
         this.props.organizationId,
         userId
@@ -359,6 +344,7 @@ export class PeopleList extends Component {
               open={!!this.state.passwordResetHash}
               requestClose={this.handlePasswordResetClose}
               passwordResetHash={this.state.passwordResetHash}
+              isAuth0={window.PASSPORT_STRATEGY === "auth0"}
             />
           </div>
         )}
--- a/src/containers/PeopleList.jsx
+++ b/src/containers/PeopleList.jsx
@@ -143,8 +143,8 @@ export class PeopleList extends Component {
           whiteSpace: "pre-line"
         }
       });
-      return columns;
     }
+    return columns;
   };
 
   editUser = userId => {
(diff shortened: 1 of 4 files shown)


src/containers/PeopleList.jsx
```diff
@@ -129,21 +129,19 @@ export class PeopleList extends Component {
         }
       }
     ];
-    if (window.PASSPORT_STRATEGY !== "slack") {
```
crayolakat: Why is this if statement being removed?

src/server/api/schema.js
```diff
@@ -649,23 +650,71 @@ const rootMutations = {
...
-    resetUserPassword: async (_, { organizationId, userId }, { user }) => {
+    resetUserPassword: async (
+      _,
+      { organizationId, userId },
+      { user, loaders }
+    ) => {
       if (user.id === userId) {
         throw new Error("You can't reset your own password.");
       }
+      // might need to impliment user password change available elsewhere.
```
crayolakat, line 661: Can you expand more about this comment?

src/server/api/schema.js
```diff
@@ -649,23 +650,71 @@ const rootMutations = {
...
+      const organization = await loaders.organization.load(organizationId);
 
-      const userRes = await r
-        .knex("user")
-        .where("id", userId)
-        .update({
-          auth0_id
+      const passportStrategy = getConfig("PASSPORT_STRATEGY", organization);
+      if (passportStrategy === "auth0") {
+        // BEGINNING
```
crayolakat: I think this comment can be removed

src/server/api/schema.js
```diff
@@ -649,23 +650,71 @@ const rootMutations = {
...
+          });
+          res = await res.text();
+          console.log(res, email);
+          return res;
+        } catch (err) {
+          //handles error and sends it to the client
+          throw new Error(err);
+        }
+      } else {
+        //returns one of the options
```
crayolakat: Can you expand more on this comment?

crayolakat (changes requested):
Thanks for the PR! A question: Instead of removing the "Change Password" button, can we make it do the same thing as when an admin clicks the "Reset Password" button?

I also left some additional questions on specific lines of code. Also, I encountered this edge case:

1. Log in as a user via Auth0
2. Go to the user profile page and change your email
3. Log out
5. Issue 1: If the user tries logging in again they must log in using their old email and password. The new email isn't updated in Auth0
6. Log in as an admin
7. Click the "Reset Password" button of the user who just changed their email
8. Issue 2: A message saying "The user has been sent an Auth0 password reset link!" appears but nothing actually happens. The user won't receive an email to reset their password because the new email isn't linked to an Auth0 account. The admin must log into Auth0 and manually change the user's email

Issue 1 is already present in the main repo, so doesn't have to be addressed in this PR, but I just wanted to point it out. I'll create a new issue for Issue 1 in GitHub.

codygordon:
@crayolakat fwiw I co-signed on @Arique1104's justification of removing the user's ability to reset their own password using that button: we were thinking the reset can be triggered from the login form, and there's not much of a reason for them to do it within the app if they are already in with a working password. But it could probably be implemented pretty easily (and you were right, it can be done from the client just fine!) if we do think users frequently want to reset their password when logged in.

As far as the changing of the email, that's a good observation. Allowing Auth0 users to change their emails is somewhat complicated and requires using the [Management API](https://auth0.com/docs/api/management/v2#!/Users/patch_users_by_id), and we'd probably want to validate the email is owned by the user by verifying it via auth0 OTP before the change is accepted.

Maybe we should just disable editing the email for auth0 users in the meantime, since it doesn't actually work?

crayolakat:
@codygordon Thanks for clarifying! I'm fine either way with if the "Change Password" button is removed for Auth0 or if it does the same thing as the "Reset Password" button. So up to @Arique1104 if they want to make that change in this PR or keep it as is

I made an [GitHub Issue](https://github.com/MoveOnOrg/Spoke/issues/2155) for the changing email problem, and am going to add your observations about how changing email is somewhat complicated and maybe should just be disabled for Auth0 users

src/server/api/schema.js
Arique1104: Oh!  Currently an administrator is not allowed to change their own password.  Cody suggested that we should change that, so I made a note of it in the code.

src/server/api/schema.js
Arique1104: I believe this is psuedo code that I failed to remove.  Thanks for catching that!

src/containers/PeopleList.jsx
Arique1104: It must've gotten lost in the refactoring.  Great catch Kathy!

Arique1104:
@codygordon, thanks for explaining our process!

@crayolakat, I've removed the buttons "Save" and "Cancel" by checking if the Passport Strategy isn't "auth0" (as you can see below)

Screen Shot 2022-04-12 at 8 21 59 AM

I also returned the if statement that was lost during refactor on PeoplesList line 132 and it compiles successfully!!
Screen Shot 2022-04-12 at 8 28 58 AM

src/containers/PeopleList.jsx
```diff
@@ -143,8 +143,8 @@ export class PeopleList extends Component {
           whiteSpace: "pre-line"
         }
       });
+      return columns;
     }
-    return columns;
```
crayolakat: I don't think you meant to move this line, right?

src/containers/UserEdit.jsx
```diff
@@ -374,26 +374,33 @@ export class UserEditBase extends React.Component {
...
+              !fieldsNeeded &&
+              window.PASSPORT_STRATEGY !== "auth0" && (
+                <Button
+                  className={css(styles.cancel)}
+                  variant="outlined"
+                  onClick={onCancel}
+                >
+                  Cancel
+                </Button>
+              )}
```
crayolakat: If the `Save` and `Cancel` buttons are removed, then users won't be able to update their first name, last name, texting alias, or cell phone number. Instead of removing the `Save` and `Cancel` buttons, can we just disable the ability to edit the Email field if Auth0 is enabled?

crayolakat (changes requested):

src/containers/UserEdit.jsx
Arique1104: We have changed this code so that it reflects the changes you requested. :)

crayolakat (dismissed):
Looks great, thank you!!

crayolakat:
This was merged into the `main` branch for the 12.2 release, but some rebasing caused the PR to not close automatically. Manually closing the PR.

PR #2189, opened 2022-06-23
JWT-powered Delegated Authentication

reyattrestle:
## Description

This PR adds a new passport strategy based on `passport-jwt` that will use JSON Web Tokens to delegate authentication to a third-party app. This essentially is a stripped down version of what Auth0 provides, but meant to use where both systems are maintained by the same group.

Note about length of this PR: it includes docs and `package.json` changes, which makes it go over the 300 lines limit.

# Checklist:

- [ ] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#submitting-your-pull-request), or has a documented reason in the description why it’s longer


.env.example
```diff
@@ -50,3 +50,10 @@ DST_REFERENCE_TIMEZONE='US/Eastern'
 PASSPORT_STRATEGY=local
 TEXTER_SIDEBOXES=celebration-gif,default-dynamicassignment,default-releasecontacts,contact-reference,tag-contact,freshworks-widget,default-editinitial,take-conversations,hide-media,texter-feedback
 OWNER_CONFIGURABLE=ALL
+
+
+#PASSPORT_STRATEGY=token
+#TOKEN_AUTH_URL=
+#TOKEN_AUTH_SHARED_SECRET=secret
+#TOKEN_AUTH_ISSUER=issuer
+#TOKEN_AUTH_AUDIENCE=audience
```
crayolakat: Can you move this example configuration to the `docs/HOWTO_INTEGRATE_TOKEN_AUTH.md` file?

Dockerfile
```diff
@@ -21,6 +21,7 @@ RUN yarn install --ignore-scripts --non-interactive --frozen-lockfile && \
 FROM ${RUNTIME_IMAGE}
 WORKDIR /spoke
 COPY --from=builder /spoke/build build
+COPY --from=builder /spoke/migrations migrations
```
crayolakat: Why is this needed?

package.json
```diff
@@ -197,6 +199,7 @@
     "regenerator-runtime": "^0.10.5",
     "selenium-webdriver": "^3.6.0",
     "sqlite3": "^4.1.1",
+    "supertest": "^6.2.3",
```
crayolakat, line 202: What gains are you getting from using `supertest` vs the current libraries present in the codebase?

crayolakat (changes requested):
Hey Rey, thanks for this PR! Very well written and useful new passport strategy.

I just had a few small questions, and a request to move the example configuration out of the .env.example file and into the `docs/` folder.

.env.example
reyattrestle: Sure thing!

Dockerfile
reyattrestle: It is not! I was just confirming, but because of how I have this sequenced in my docker-compose file, the postgres database is creating while Spoke is spinning up, which I was having trouble with because even with `SUPPRESS_DATABASE_AUTOCREATE` the migrations were still being called.

I just read 10 lines-ish below and saw that I can just have the same effect by setting `SUPPRESS_MIGRATIONS` and `SUPPRESS_SEED_CALLS` to truthy values.

I'll take this out.

package.json
reyattrestle: The main feature I am leveraging from supertest is the ability to test out http requests in process without having to set up a server. I am using this feature in a test to allow me to set up the configuration params of the JWT issuer to be something unique to the test, and supertest allows me to assert on the response.

Basically these two assertions here:
```js
       // Login callback will send user to the appropriate page with the right token
        await request(app)
          .get(`/login-callback?nextUrl=/admin/${orgId}/campaigns`)
          .set("Authorization", `Bearer ${signedToken}`)
          .expect(302)
          .expect("Location", `/admin/${orgId}/campaigns`);

        // Bad tokens get rejected
        await request(app)
          .get(`/login-callback`)
          .set("Authorization", `Bearer ${invalidToken}`)
          .expect(401);
 ```

.env.example
reyattrestle: As a matter of fact, it's already there, so I'll just remove from the `.env.example` file

crayolakat (approved):

Issue #1037, opened 2019-01-17
Create option to include qualitative notes to a conversation


deasterdaywfp:
Allow collection of qualitative data by texters and an option to flag a convo they may need to reassign or need help on

harpojaeger:
This is starting to venture into CRM territory. So far, Spoke has not been concerned with persisting data and has preferred to leave that to an org's external CRM. What kind of data do you think it'd be useful to collect here?

Separately, users being able to flag a contact as needing reassignment sounds like a good idea, but separate from this one.

deasterdaywfp:
Short free text that would include survey questions that could be like
primary language, gender identity, race/ethnicity identity, etc. Basically
there are lots of uses that we have for this included past campaigns where
we collect short responses of how a policy or action impacts that
individual that can be used in a social media storytelling campaign or
testimony.

I think this is hardly CRM level info to collect a short free text giving
more context to an interaction.

*Dylan Easterday*
*National Volunteer Manager, Digital - Working Families Party*
*Working Families <http://workingfamilies.org>*
*Text *(347) 460-9451 or Schedule https://calendly.com/deasterday
Email deasterday@workingfamilies.org
Done with email? me too - join me here <http://bit.ly/WFP4theMany>

DinahSanders:
A few questions that come to mind immediately:
- What is the expected life-cycle of that piece of free text?
- What methods of export are required for it?
- Where is it stored?
- How is the association with the interaction maintained? (since that seems to be a requirement) and what does the life-cycle of interaction data mean for this piece of free text?
- Since this could be sensitive personal information associated with contact information (at least a first name and phone number), what privacy protections need to be in place, both in regards to its storage and who can view it?

deasterdaywfp:
Great questions!

For what I expect would be the use cases with WFP and our allies we work
directly with here are the thoughts I have:

   - What is the expected life-cycle of that piece of free text? Likely no
   longer than the period of the campaign, so typically anywhere from a week
   to 12 months.
   - What methods of export are required for it? I imagined export as text
   in a field exported as csv or txt file much like the message and survey
   export files format them currently. We would like to integrate these right
   into our AWS or Action Network or similar CRM at some point, but not
   immediately.
   - Where is it stored? Why not in the same table as other survey data?
   - How is the association with the interaction maintained? (since that
   seems to be a requirement) and what does the life-cycle of interaction data
   mean for this piece of free text? cannot this not be associated via
[...]

PR #2192, opened 2022-07-04
Texter Sidebox Contact Notes Feature

codygordon:
Issue https://github.com/MoveOnOrg/Spoke/issues/1037

## Description

Adds a new `contact-notes` Texter Sidebox feature.

When enabled, allows entry of notes in a multiline text input in the sidebox. Notes are saved
to a campaign contact's `customFields` object using a `notes` key.

Notes can be persisted across campaigns for the same contact by simply loading the notes data to persist into a column called `notes` in contacts list uploads. All texters will see any `notes` data loaded in this way in the input when viewing the contact if this setting is enabled.

This PR is longer than 300 lines because it required adding a new mutation and cacheable query update for campaign contact data. There is no generic `update` function for campaign contacts, as such I implemented the updates here only for `customFields` rather than a more generic update to be in line with the other more specific update functions.

# Checklist:


src/components/AssignmentTexter/ContactController.jsx
```diff
@@ -232,7 +232,16 @@ export class ContactController extends React.Component {
   getContact(contacts, index) {
     if (contacts.length > index) {
       // console.log('getcontact', index, (contacts[index]||{}).id, (contacts.length > index + 1 ? 'next' + (contacts[index+1]||{}).id : 'end'))
-      return contacts[index];
+      const contact = contacts[index];
+      if (typeof contact.customFields === "string") {
+        try {
+          contact.customFields = JSON.parse(contact.customFields || "{}");
```
ibrand: Am I understanding this right that you're changing it from just returning a contact to parsing the JSON as a side effect and then returning the contact?

If so, one nit / personal preference that you could ignore would be to make a tiny function called `tryParsingContactCustomFields` that does this just to help the code self-document a little more. If not, what is happening here?

src/components/AssignmentTexter/ContactController.jsx
```diff
@@ -261,6 +270,19 @@ export class ContactController extends React.Component {
     );
   };
 
+  updateContactData = (contactId, newData) => {
```
ibrand, line 263: is this basically a cache busting / overwriting function?

src/components/AssignmentTexter/Controls.jsx
```diff
@@ -171,6 +171,12 @@ export class AssignmentTexterContactControls extends React.Component {
   };
 
   onKeyUp = evt => {
+    if (evt.target && evt.target.id === "contact-notes-input") {
```
ibrand: oh this is a good catch and is nice and neat. One follow up here: is this the only texter sidebox that should overwrite this behavior? One thought would be to do:

```suggestion
    const texterSideboxesThatAllowInput = ["contact-notes-input"]
    if (evt.target && texterSideboxesThatAllowInput.includes(evt.target.id)) {
```

Or something similar if we think we'd want this to be reused. Otherwise, this is good!

src/extensions/texter-sideboxes/contact-notes/react-component.js
```diff
@@ -0,0 +1,166 @@
+import PropTypes from "prop-types";
+import React from "react";
+import { withRouter } from "react-router";
+import gql from "graphql-tag";
+import _ from "lodash";
+// import { compose } from "recompose";
```
ibrand: Nit: delete unused import

src/extensions/texter-sideboxes/contact-notes/react-component.js
```diff
@@ -0,0 +1,166 @@
...
+    super(props);
+    this.state = {
+      notes: props.contact.customFields.notes || "",
+      isSaving: false,
+      hasError: false
+    };
+  }
+
+  componentDidUpdate() {
+    // console.log(this.props.contact.customFields)
```
ibrand: nit: delete comment

src/extensions/texter-sideboxes/contact-notes/react-component.js
```diff
@@ -0,0 +1,166 @@
...
+
+TexterSideboxClass.propTypes = {
+  router: PropTypes.object,
+  mutations: PropTypes.object,
+  contact: PropTypes.object,
+  updateContactData: PropTypes.func
+};
+
+export const mutations = {
+  updateContactCustomFields: ownProps => customFields => ({
```
ibrand, line 129: I've never seen 2 `=>` in a row, is that a typo? or it is syntax for a closure or something?

src/lib/scripts.js
```diff
@@ -54,7 +54,7 @@ export const allScriptFields = (customFields, includeDeprecated) =>
 const capitalize = str => {
   const strTrimmed = str.trim();
   if (
-    strTrimmed.charAt(0).toUpperCase() == strTrimmed.charAt(0) &&
+    strTrimmed.charAt(0).toUpperCase() === strTrimmed.charAt(0) &&
```
ibrand, line 57: This seems like a little bit of a dangerous change, but I'm not against it. Just raising that it could break things in unexpected places. I do notice that it's only used once in this file on line 97 so it's probably fine

src/lib/scripts.js
```diff
@@ -105,11 +105,17 @@ export const applyScript = ({ script, contact, customFields, texter }) => {
   let appliedScript = script;
 
   for (const field of scriptFields) {
-    const re = new RegExp(`${delimit(field)}`, "g");
-    appliedScript = appliedScript.replace(
-      re,
-      getScriptFieldValue(contact, texter, field)
-    );
+    if (field === "notes") continue;
```
ibrand: this is deeper than a sidebox normally needs to go into core code. Is this because fields were being replaced in the notes box?

src/server/models/cacheable_queries/campaign-contact.js
```diff
@@ -562,6 +562,58 @@ const campaignContactCache = {
     } catch (err) {
       console.log("contact updateStatus Error", newStatus, contact, err);
     }
+  },
+  updateCustomFields: async (contact, customFields, campaign, organization) => {
```
ibrand, line 566: This is the part of the code that I understand the least because of my knowledge of caching. I'm trying to think of if there are any particularly clever ways to test this...

ibrand (commented):
Some questions and nit picks but looking really good!

src/components/AssignmentTexter/ContactController.jsx
codygordon: yeah, the `ContactController` is basically a client in-memory cache for campaign contacts using React state, and this updates a single contact in the client state with new data (in this case, just `customFields`, but it could be used for other things in the future). This allows updated notes to show up if navigating between contacts using the arrows, which relies on this client state to populate as it does not re-fetch from the server.

src/lib/scripts.js
codygordon: I believe this was an automatic change made by my editor as defined by our lint rules (or maybe the error was bothering me and I changed it 😛).  The **[eqeqeq](https://eslint.org/docs/latest/rules/eqeqeq)** rule calls for `===` to always be used in our code.  Really, `==` is potentially dangerous and should basically never be used in JS:

> The reason for this is that == and != do type coercion which follows the rather obscure [Abstract Equality Comparison Algorithm](https://www.ecma-international.org/ecma-262/5.1/#sec-11.9.3). For instance, the following statements are all considered true:
>
> 
> [] == false
> [] == ![]
> 3 == "03"
> If one of those occurs in an innocent-looking statement such as a == b the actual problem is very difficult to spot.

If you ever see `==` in our JS code it's a result of folks not setting up a linter in their editor or ignoring that rule error when they see it.

In this case, I committed it because I could tell the intent was not to do an abstract equality check and the result would be the same using either operator, but this cleared all the lint errors from this particular file. 100s more to go...

src/lib/scripts.js
codygordon: This line eliminates the `notes` field from being usable in scripts, imo you don't want to accidentally populate them into a text!

Below is a somewhat unrelated bugfix; catching a regexp parse error which was causing crashes very rarely with certain characters loaded into custom field names.

src/extensions/texter-sideboxes/contact-notes/react-component.js
codygordon: Not a typo, this is how we set up callable mutations in the client! I do believe it's shorthand for a closure. Here's a more verbose illustration of what's happening:

```
const firstFunction(props) => {
  const secondFunction(data) => {
    // do something with props and data
  }
  return secondFunction
}

const secondFunctionWithProps = firstFunction(props)
secondFunctionWithProps(data)
```

In this case, our [`load-data` hoc](https://github.com/MoveOnOrg/Spoke/blob/main/src/containers/hoc/load-data.jsx) sets up mutations; the first call injects the props on render and returns the second function to be callable when we're ready to send data (in this case,`customFields`).

src/components/AssignmentTexter/Controls.jsx
codygordon: excellent suggestion, implemented with a slight tweak :)
https://github.com/MoveOnOrg/Spoke/pull/2192/commits/257400e30a26544526f97cc4a454cf38b24eb162

src/lib/scripts.js
codygordon: I actually went ahead and implemented a better filter here that first checks for the sidebox being enabled to hide the `notes` field. Also, the basic check I had in the for loop wasn't removing it from the script creator, just disallowing its value to be sent to the texter!

https://github.com/MoveOnOrg/Spoke/pull/2192/commits/af7aec7e10c83ce42285b69c579356beec70f423
https://github.com/MoveOnOrg/Spoke/commit/fcf7e7470b80319fb7bbc313f600f83d24ee5919

src/components/AssignmentTexter/ContactController.jsx
codygordon: turns out this wasn't necessary, so changed it back! https://github.com/MoveOnOrg/Spoke/pull/2192/commits/dbd58d41c98a8e0b49b16c14b1385f73b01156be

src/extensions/texter-sideboxes/contact-notes/react-component.js
```diff
@@ -0,0 +1,164 @@
...
+export const displayName = () => "Contact notes";
+
+export const showSidebox = () => true;
+
+export class TexterSideboxClass extends React.Component {
+  constructor(props) {
+    /* eslint-disable no-param-reassign */
+    if (typeof props.contact.customFields === "string") {
+      try {
+        props.contact.customFields = JSON.parse(props.contact.customFields);
```
ibrand: Is this block still necessary re: https://github.com/MoveOnOrg/Spoke/pull/2192#discussion_r917423549?

codygordon: Yep this is what makes the other one unnecessary. I realized other places in the client expect the `customFields` to be stringified, so the conversion should happen only here rather than on the server.

That said, [I discovered and fixed a regression here](https://github.com/MoveOnOrg/Spoke/pull/2192/commits/47df5ca30e1bf4267dc6d48e699615d0c190fffa) – I was mutating the `customFields` in the props like I had it which was causing issues in other places that expected it to be a string. A reminder to myself never to directly reassign values to props or state objects in React, always create a copy!

src/extensions/texter-sideboxes/contact-notes/react-component.js
ibrand: aha yes that is a closure! Neat. Thanks for the explanation

src/lib/scripts.js
ibrand: Totally agree with this. My only fear was that now that it's in there, something surprising might be relying on a faulty type check like `==` -- but I agree with what you're saying and I'm in favor

ibrand (dismissed):
This looks nice! This has been manual QA'd and the code looks great 🎉

crayolakat (dismissed):

crayolakat (approved):

PR #2232, opened 2022-10-20
Improve accessibility for texters using screen reader

crayolakat:
## Description

Improve accessibility for texters using a screen reader. These changes were tested using JAWS 2022.

# Checklist:

- [X] [My change is 300 lines of code or less](https://github.com/MoveOnOrg/Spoke/blob/main/CONTRIBUTING.md#submitting-your-pull-request), or has a documented reason in the description why it’s longer

### DIFF
--- a/src/components/AssignmentTexter/ContactToolbar.jsx
+++ b/src/components/AssignmentTexter/ContactToolbar.jsx
@@ -128,12 +128,10 @@ const ContactToolbar = function ContactToolbar(props) {
         title={global.ASSIGNMENT_CONTACTS_SIDEBAR ? "Toggle Contact List" : ""}
       >
         <IconButton
+          aria-label="Toggle Contact List"
           className={css(styles.contactToolbarIconButton)}
-          onClick={() => {
-            global.ASSIGNMENT_CONTACTS_SIDEBAR
-              ? props.toggleContactList()
-              : null;
-          }}
+          disabled={global.ASSIGNMENT_CONTACTS_SIDEBAR ? false : true}
+          onClick={props.toggleContactList}
         >
           <FaceIcon style={{ width: 42 }} htmlColor="white" />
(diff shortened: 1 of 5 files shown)


src/components/AssignmentTexter/ContactToolbar.jsx
```diff
@@ -128,12 +128,10 @@ const ContactToolbar = function ContactToolbar(props) {
         title={global.ASSIGNMENT_CONTACTS_SIDEBAR ? "Toggle Contact List" : ""}
       >
         <IconButton
+          aria-label="Toggle Contact List"
           className={css(styles.contactToolbarIconButton)}
-          onClick={() => {
-            global.ASSIGNMENT_CONTACTS_SIDEBAR
```
ibrand, line 133: What is this `global. ASSIGNMENT_CONTACTS_SIDEBAR` ? and why do we not need to check it anymore?

crayolakat: `ASSIGNMENT_CONTACTS_SIDEBAR` is an environment variable. Previously, if `ASSIGNMENT_CONTACTS_SIDEBAR` is set, the "Toogle Contact List" button will run the function `props.toggleContactList()` when it's clicked, else it does nothing. I added the condition property `disabled={global.ASSIGNMENT_CONTACTS_SIDEBAR ? false : true}` to disable the "Toggle Contact List" button if `ASSIGNMENT_CONTACTS_SIDEBAR` isn't set. I did this to prevent confusion because the JAWS screen reader reads it as a clickable button even though it actually doesn't do anything when clicked. If the `disabled` property to set to `true`, the `onClick` method automatically doesn't run, so no need to check for the `global.ASSIGNMENT_CONTACTS_SIDEBAR` condition anymore.

PR #2262, opened 2023-02-21
One-Click Heroku Bug Fix

Arique1104:
# Fixes
- Fixes Heroku Deprecation
- Fixes Node error on Dockerfile.

## Description
_Upon the submission of Version 13 Update, a bug was introduced to the codebase regarding deprecated heroku settings AND an issue regardingg the Docker Node16 update._

# Checklist:

- [ ] I have QA'd my fix with three successful Heroku deployments that successfully deliver text messages.


app.json
```diff
@@ -181,12 +181,17 @@
...
+      "description": "A twilio service message account SID",
+      "required": false
     }
   },
   "addons": [
-    "heroku-postgresql:hobby-dev",
+    "heroku-postgresql:mini",
     {
-      "plan": "heroku-redis:hobby-dev",
+      "plan": "heroku-redis:mini",
```
lastminutediorama, line 194: If I remember right, we have documentation that provides estimated costs for running Spoke in Heroku. Since this looks like a different postgresql tier, should we be updating that documentation along with this change?

lastminutediorama (dismissed):
This seems straightforward enough, but I have one non-blocking question about documentation.

Arique1104:
Hey @lastminutediorama, thank you so much for pairing with me today.  Below is my notes, but I like to work backwards so let's start with what I wrote down on what our next steps are:

_Mark wants to know_
- Why do the jobs want to stop?  Clearly github is upset about something.
- Will review error message and why Github thinks stopping the tests is a good idea

_Arique wants to know_
- Is this pull request doing too much?  It seems like the node issue and the heroku deprecation issue are seperate.  And if I were to start over and isolate those two edits, we might be able to improve our Heroku deployments while also continuing to figure out what our Docker file needs to pass these github deployment tests. 

*NOTES FROM OUR MEETING TODAY*

- Node versions are failing
- Made me think about my docker update.  Updated RUNTIME_IMAGE to 16.8.0-alpine
- Tests failed more
- NODEJS UPDATES require Migration steps, what would be the steps for our most resent update?
- The VAN error feels like a red-herring, but it should be something we look into once we've isolated the bug.

app.json
Arique1104: That makes a lot of sense @lastminutediorama !  Yes, I will create a ticket to provide that update and deliver that to this branch.

lastminutediorama (dismissed):
ok, looks good to me

lastminutediorama (approved):

PR #2298, opened 2023-09-28
Mobile compatible chat

### LINKED ISSUE #2269: Mobile-Only Update
Make the Texter UI available on Mobile

KarimAl-Rashdan:
# Fixes #2269 

## Description

Adds mobile update to volunteer chat feature.
Screen Shot 2023-09-28 at 8 51 25 PM

Screen Shot 2023-09-28 at 8 50 13 PM

Screen Shot 2023-09-28 at 8 51 00 PM

# Checklist:

### DIFF
--- a/src/components/AssignmentTexter/StyleControls.js
+++ b/src/components/AssignmentTexter/StyleControls.js
@@ -62,7 +62,13 @@ export const flexStyles = StyleSheet.create({
     right: 0,
     display: "flex",
     flexDirection: "column",
-    height: "100%"
+    //messages can be scrolled through and height/width is responsive 
+    height: "100%",
+    "@media(max-width: 420px)": {
+      fontFamily: "Poppins",
+      overflowY: "scroll !important",
+      width: "100%"
+    }
   },
   popoverSideboxesInner: {
     // expand to fill the whole popover
@@ -114,7 +120,7 @@ export const flexStyles = StyleSheet.create({
     height: "100%"
   },
   superSectionMessageBox: {
-    flex: "1 2 auto",
+    height: "100%",
     overflowY: "scroll",
     overflow: "-moz-scrollbars-vertical",
     overflowX: "hidden",
@@ -124,6 +130,9 @@ export const flexStyles = StyleSheet.create({
   },
   /// * Section Scrolling Message Thread
   sectionMessageThread: {
+    display: "flex",
+    justifyContent: "center",
+    alignItems: "flex-start",
     flex: "1 1 auto",
     overflowY: "scroll"
   },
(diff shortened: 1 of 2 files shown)


src/components/AssignmentTexter/StyleControls.js
```diff
@@ -169,19 +183,19 @@ export const flexStyles = StyleSheet.create({
...
   /// * Section Reply/Exit Buttons
   sectionButtons: {
     // TODO: maybe make this contingent on whether there are answer buttons
-    "@media(max-height: 600px)": {
-      flexBasis: "96px" // TODO
-    },
-    "@media(min-height: 600px)": {
-      flexBasis: "144px" // TODO
-    },
+    // "@media(max-height: 600px)": {
```
crayolakat: Is there a reason this block of code is commented out instead of deleted?

src/components/AssignmentTexter/StyleControls.js
```diff
@@ -197,7 +211,7 @@ export const flexStyles = StyleSheet.create({
     // height:105: webkit needs constraint on height sometimes
     //   during the inflection point of showing the shortcut-buttons
     //   without the height, the exit buttons get pushed down oddly
-    height: "15px", //TODO
+    // height: "15px", //TODO
```
crayolakat: Is there a reason this line of code is commented out instead of deleted?

crayolakat (changes requested):
Hi Karim, thank you for submitting this PR! I just have 1 question:

Is there a reason why some blocks/lines of code are commented out instead of deleted?

src/components/AssignmentTexter/StyleControls.js
KarimAl-Rashdan: Good catch Kathy! I don't remember commenting this out, but yes it should be deleted.

src/components/AssignmentTexter/StyleControls.js
KarimAl-Rashdan: This block of code was impacting my styling; however, since I didn't write this code I wanted to touch base with the team and see if it's ok to delete it.

crayolakat: Yup! It's fine to delete

crayolakat (approved):

Arique1104 (changes requested):
**Notes from Community Coalition QA**
TLDR;
- On a cell phone, after a volunteer opts out a phone number within a texting campaign, can  the back arrow on the top half of the screen successfully override the opt-out and allow the volunteer to return back to an opted-out phone number and successfully send a text message?
- If the above is true, can we resolve this by removing the second bar from the top and ensuring that a functional home button and skip button already provide the back and forth functionality we would lose by removing those arrows?
- Is there functionality that we lose when getting rid of that second bar?  If so, what would we want to keep?

Notes Continued:
- These changes work well, although we hit some interesting moments during the QA for some edge cases.
- Some how, between using the opt-out button, we lose the ability to use the <- arrow at the top of the screen.  
- And when pressed back enough times, it does look like our Spoke instance will permit to sending a text to someone who has already been opted-out.
- We would need to try this on a production texting framework to confirm that this only occurs in development because we are using fake data.  Would like to run a live QA on a production site to validate this edge case concern.
- We have to check and see if the "All Responses" tab actually scrolls down if it contains 20+ canned responses within the campaign.
- Esthetically we are deciding if the Second Tab from the top of the screen with the <- and -> buttons is strategically necessary when there is a skip button and a home button to return to text messages that received responses.  Removing that second box from the top (or make it more useful) would help with the back-arrow opt-out bug we found in QA.  Removing those back and forth arrow and simply keeping the "skip" button would protect our functionality and rid us of the opt-out bug.
- @crayolakat, what would you suggest we prioritize?  It's difficult to predict if this is a real issue in production without a live API key to test it.  Is this something that Travis can include as part of his QA?  Checking the mobile feature update for these edge cases?

crayolakat:
Hi @Arique1104, I tested this on fake data on MoveOn's staging website, and was not able to replicate the issue with the <- arrow or being able to text someone who is already opted out.

Also, I verified that the "All Responses" tab scrolls down if it contains 20+ canned responses within the campaign.

I will coordinate with you further on Slack

Arique1104 (approved):

Arique1104:
After careful review in our Staging Branch, we were able to confirm that no new bugs were introduced to the code base with these changes.

A bug that currently exists in main is now [reported correctly here.](https://github.com/MoveOnOrg/Spoke/issues/2309)

This new mobile update functionality will now be part of the Spoke codebase.  Thank you so much for your work on this @KarimAl-Rashdan!

PR #2360, opened 2024-04-03
New node 20 branch texter side boxes

engelhartrueben:
# Fixes # (issue)
- Added support for forms that require integers (`GSIntegerFile.jsx`)
- Adjusted .env.example to better represent the dynamic assignment features
- Added clarifying text and linked the github doc for batch strategies
- Added clarifying text to Dynamic Assignment drop down about adding multiple strategies

## Description

Mostly QOL, adding comments and what not.

The `ctrl+click` text is new. I didn't know I could do this, which probably means others don't as well
image

But the GSIntegerField.jsx was necessary due to the stricter GraphQL update that just occurred. Previously, options looking to gather user input for numbers, like batch size, would use GSFormField, which would return a string that contained a number. Now, with GSIntegerField, the user input can be passed directly as an integer. 

# Checklist:

# NOTE
I have not looked at merge conflicts - just getting this out here before class. Will be able to come back at 4/3/2024 4:30 EST to look that over.


src/components/CampaignDynamicAssignmentForm.jsx
```diff
@@ -103,7 +102,7 @@ class CampaignDynamicAssignmentForm extends React.Component {
                 switch to replying.
               </p>
               <Form.Field
-                as={GSTextField}
+                as={GSIntegerField}
```
mau11, line 105: nice!

engelhartrueben:
There is a cypress test failing, which is odd. Will need more time to look into it prior to merging

engelhartrueben:
Cypress Test resolved.

__test__/cypress/integration/basic-campaign-e2e.test.js
```diff
@@ -101,7 +101,7 @@ describe("End-to-end campaign flow", () => {
     cy.get("[data-test=editorInteraction] textarea[name=script]").click();
     cy.wait(400);
     cy.get(".DraftEditor-root").type(
-      "Hi {{}firstName{}} this is {{}texterFirstName{}}, how are you?"
+      "Hi {{}firstName{}} this is {{}texterAliasOrFirstName{}}, how are you?"
```
mau11, line 104: good catch, weird that this failed, but passed on the parent branch

.env.example
```diff
@@ -49,6 +49,7 @@ TWILIO_MESSAGE_VALIDITY_PERIOD=
 DST_REFERENCE_TIMEZONE='US/Eastern'
 PASSPORT_STRATEGY=local
 TEXTER_SIDEBOXES=celebration-gif,default-dynamicassignment,default-releasecontacts,contact-reference,tag-contact,freshworks-widget,default-editinitial,take-conversations,hide-media,texter-feedback,contact-notes
+DYNAMICASSIGNMENT_BATCHES=finished-replies-tz,vetted-texters,finished-replies
```
mau11, line 52: I believe we'll need to update the docs for this: docs/REFERENCE-environment_variables.md

engelhartrueben: Good catch. This change might not be entirely necessary. It is only when trying to get "vetted-texterassignment" to work. So if you were to just do `DYNAMICASSIGNMENT_BATCHES=vetted-texterassignment", that entire feature would brick as it would need those other three.

src/components/forms/GSIntegerField.jsx
```diff
@@ -0,0 +1,103 @@
+import React from "react";
+import TextField from "@material-ui/core/TextField";
+import GSFormField from "./GSFormField";
+import theme from "../../styles/mui-theme";
+
+export default class GSIntegerField extends GSFormField {
```
mau11, line 6: we can discuss this more later, but I'd like to start moving towards building functional components over class components, especially with modern react, we'll have access to more features (you don't need to make changes for this, just something we can talk about after this upgrade)

mau11 (approved):
🎉

PR #2408, opened 2024-07-30
Grouped survey stats

schuyler1d:
## Description

Requested by Jim N (I can't find his Github account) -- This supports two groups of pie chart reports for survey/questions.

1. If multiple *questions* have text surrounded by `[]`s that is the same, then those question's answers will be grouped together at the top with a title of "Grouped Questions with [question text]"
    * example: question 1 = [Will you signup?] w/ answers "Yes", "No" and "Not sure" with a Not sure response nagging again and then asking [Will you signup?] really? and then "Yes" "No", then those questions and their "Yes"s and "No"s will be grouped into a single question and the "Not sure" will be added in-tandem (this makes the pie-chart a bit of a lie, if one of the answers leads to follow-on answers)
2. Furthermore, and separately, if multiple *answers* have text surrounded by `[]`s that are the same, then those answers will be grouped all together.
    * example: Question 1 = "Will you sign up?" w/ answers [Yes] and [No] and a later question also has answers [Yes] and [No] (or just one of those), then all of those will be grouped into "Responses Grouped by Answers (with []'s)" -- and a single pie chart will have an aswer [Yes] that is all the answers with [Yes] in them combined.

Finally, button text will also work if you make it .e.g "[Yes] I like it" => "Yes" just like "Yes: I like it" would do so before.  This makes it easier for buttons across questions to map to the grouped answer values in (2).

I'd like feedback on this feature before adding documentation to make sure one or the other or both features are what's desired and that it can work.

# Checklist: 

spoke_surveysgrouped

### DIFF
--- a/src/components/AssignmentTexter/Controls.jsx
+++ b/src/components/AssignmentTexter/Controls.jsx
@@ -825,8 +825,8 @@ export class AssignmentTexterContactControls extends React.Component {
       currentQuestionAnswered = questionResponses[currentInteractionStep.id];
       const dupeTester = {};
       const shortener = answerValue => {
-        // label is for one-word values or e.g. "Yes: ...."
-        const label = answerValue.match(/^(\w+)([^\s\w]|$)/);
+        // label is for one-word values or e.g. "Yes: ...." or "[Yes] ..."
+        const label = answerValue.match(/^\W?(\w+)([^\s\w]|$)/);
         return label ? label[1] : answerValue;
       };
       currentQuestionOptions = currentQuestion.answerOptions
--- a/src/components/CollapsibleCard.jsx
+++ b/src/components/CollapsibleCard.jsx
@@ -1,0 +1,75 @@
+import PropTypes from "prop-types";
+import React from "react";
+import Chart from "../components/Chart";
+
+import Card from "@material-ui/core/Card";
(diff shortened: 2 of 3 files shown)


engelhartrueben:
@electiondaze

schuyler1d:
At @electiondaze 's request I have implemented this with grouping the sections and default-collapsing the main survey section when grouped responses exist.  See screenshot.
spoke-collapsiblestats

engelhartrueben:
Hi Sky, two things here that need addressing and then I can approve for staging!

When first opening the campaign stats page, the Pie charts are super large and don't change size until the drop down is closed and reopened again. 

https://github.com/user-attachments/assets/3b8a6c40-63e5-4435-b0a5-297276f19e9c

The `children` PropTypes are failing in `CollapsibleCard` with the error drop down due to this string being passed in the second index. This looks like an issue with how we store error codes in the props. 
image

schuyler1d:
@engelhartrueben latest commit should fix those issues.  I couldn't figure out how to make the pie charts perfectly the same size, but they are at least not enormous now.We can bump the 50% to smaller still if you like, but it might mess up surveys with very many options

engelhartrueben (approved):
🎉

PR #2410, opened 2024-08-01
AWS-SDK: Fixing broken exports

engelhartrueben:
# Fixes # (issue)
Fixes broken exports.

## Description
During the upgrade to Node 20, `aws-sdk` had to be upgraded to V3. This was simple in nature, considering a code-mod existed. However, our specific way of initiating an S3 bucket was not covered by this code-mod, and subsequently failed. Theoretically, our implementation (still in V2) was still supported, but clearly not.

This new implementation uses `aws-sdk` V3 api calls to create the S3 client and bucket. Additionally, we can set the region for the bucket, which is called from the environment variables. 

This PR:
- Checks if an AWS bucket exists, if not creates one of the specificed name found in `AWS_S3_BUCKET_NAME=`
- Brings v3 AWS-SDK update to export logic
- Only shows UI related to the type of export:
Local Export: 
image
AWS Export: 
image
(__NOTE__) this UI is fixed with #2423  
- Places a hard check on if an email (SMTP or MailGun) is set up for exports
- Adjusts the expiration for cached `campaignExportUrl` to match AWS expiration (1 day)

Tested on:
| Mode | AWS Bucket and Download| Local download | 
| --- | --- | --- |
| production | ✔️ | ❌*|
| development | ✔️|✔️|

_*Local download is not supported in production._
# Checklist:

### DIFF
--- a/src/containers/AdminCampaignStats.jsx
+++ b/src/containers/AdminCampaignStats.jsx
@@ -426,7 +426,7 @@ class AdminCampaignStats extends React.Component {
             <span>
               Export started -
               {(this.props.organizationData &&
-                process.env.EMAIL_HOST) ?
+                this.props.organizationData.organization.emailEnabled) ?
                 " we'll e-mail you when it's done. " :
               (campaign.cacheable && (
                 <span>
--- a/src/containers/AdminCampaignStats.jsx
+++ b/src/containers/AdminCampaignStats.jsx
@@ -346,8 +346,8 @@ class AdminCampaignStats extends React.Component {
             {campaign.exportResults.error && (
               <div>Export failed: {campaign.exportResults.error}</div>
             )}
-            {campaign.exportResults.campaignExportUrl &&
-            campaign.exportResults.campaignExportUrl.startsWith("http") ? (
+            {campaign.exportResults.campaignExportUrl && (
+            (campaign.exportResults.campaignExportUrl.startsWith("http")) ? (
               <div>
                 Most recent export:
                 <a href={campaign.exportResults.campaignExportUrl} download>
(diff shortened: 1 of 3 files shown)


src/containers/AdminCampaignStats.jsx
```diff
@@ -424,9 +424,10 @@ class AdminCampaignStats extends React.Component {
           message={
             <span>
               Export started -
-              {this.props.organizationData &&
-                this.props.organizationData.emailEnabled &&
-                " we'll e-mail you when it's done. "}
+              {(this.props.organizationData &&
+                this.props.organizationData.emailEnabled) ?
+                " we'll e-mail you when it's done. " :
+                " check your Spoke directory. "}
```
mau11: I just tried this locally, if we have AWS S3 set up, even if email isn't enabled, the data will save to S3, not in the spoke directory

src/workers/jobs.js
```diff
@@ -861,46 +861,74 @@ export async function exportCampaign(job) {
...
-        // The transformation for params is not implemented.
-        // Refer to UPGRADING.md on aws-sdk-js-v3 for changes needed.
-        // Please create/upvote feature request on aws-sdk-js-codemod for params.
-        params: { Bucket: process.env.AWS_S3_BUCKET_NAME }
+      const client = new S3Client({
+        region: process.env.AWS_REGION
       });
+      const bucketName = process.env.AWS_S3_BUCKET_NAME;
+      const command = new CreateBucketCommand({ Bucket : bucketName });
+
```
mau11, line 875: I don't think we want to create a new bucket each time we export data, it's likely someone will already have bucket created for the data and want to upload directly to it

we can add a check, if the bucket exists, save the exported data, if it doesn't, create a new one and save

src/workers/jobs.js
```diff
@@ -865,14 +872,46 @@ export async function exportCampaign(job) {
...
+      try {
+        // Check if the S3 bucket already exists
+        const verifyBucketCommand = new HeadBucketCommand({
+          Bucket: bucketName
+        });
+        await client.send(verifyBucketCommand);
+
+        console.log(`S3 bucket "${bucketName}" already exists.`);
+      } catch (error) {
+        if (error.name === "NotFound") {
```
engelhartrueben, line 885: `if (error.code === "NoSuchBucket")`

[Source](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html)

mau11: When the bucket is not found, it doesn't look like we get this error code, only the error name

src/containers/AdminCampaignStats.jsx
mau11: disregard my previous comment, the last change to this check fixed the issue I was seeing

engelhartrueben:
@mau11 this is set up for `main`, but can rebase to `stage-main-14.1` upon request!

mau11 (approved):
🎉

Issue #1368, opened 2020-01-14
Feature Request: Dark Mode


arena:
**Problem**
Sometimes my eyes hurt and I want to conserve electricity while I give them a rest. 

**Solution**
Dark Mode
sub-features
- way to toggle to dark mode
- user setting: 
     - light mode
     - dark mode
     - automatic based on time of day
     - automatic based on system setting
- light to dark color map

**Design Note**
Rather than inverting the light mode the colors should be dark but slightly lower contrast. I will add a mockup. ...need to take some inspiration from other dark mode apps and experiment a little to get the right colors.

azuzunaga:
Great idea, @arena! I'd add `based on user OS preference` to the `user setting` list.

arena:
@azuzunaga Great suggestion! Is that something that is possible to do with a web app (as opposed to a native mobile app).

azuzunaga:
Yes! See support for the feature in web browsers here: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

arena:
@stefanhayden is there a way to do this via a Material UI theme now that Material UI is updated?

stefanhayden:
@arena yes. I've actually done most of the work already. we just need a UI for users and dark mode should mostly just work.

PR #2418, opened 2024-08-10
1368 dark mode

mcverter:
# Fixes
 [Feature Request: Dark Mode #1368](https://github.com/StateVoicesNational/Spoke/issues/1368)

## Description
Adds a toggle switch to the upper right corner for dark mode.  It's not as fully fleshed out as the feature request.
Please note that the switch is visible on the login screen.  
This is still a draft

https://github.com/user-attachments/assets/b66795cf-3ed1-450b-a306-0e9bff84973b

# Checklist:

- [ ] [My change is 300 lines of code or less](https://github.com/StateVoicesNational/Spoke/blob/main/CONTRIBUTING.md#submitting-your-pull-request), or has a documented reason in the description why it’s longer

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -5,22 +5,31 @@ import CssBaseline from "@material-ui/core/CssBaseline";
 
 import { defaultTheme } from "../styles/mui-theme";
 import ThemeContext from "../containers/context/ThemeContext";
+import { gql } from "@apollo/client";
+import loadData from "../containers/hoc/load-data";
 
 /**
  * We will let users customize the colors but not other
  * parts of the theme object. Here we will take the string,
  * parse it, and merge it with other app theme defaults
  */
-const formatTheme = newTheme => {
+const formatTheme = (newTheme, darkMode) => {
   return {
     ...defaultTheme,
-    palette: newTheme.palette
+    palette: {...newTheme.palette,
+      type: darkMode ? "dark" : "light"}
   };
 };
 
-const App = ({ children }) => {
-  const [theme, setTheme] = useState(defaultTheme);
-  let muiTheme = createTheme(defaultTheme);
+const App = ({ children, user }) => {
+  const darkMode = user?.currentUser?.darkMode;
+  const defaultThemeWithMode = {
+    ...defaultTheme,
+    palette: { ...defaultTheme.palette, "type": darkMode ? "dark" : "light" }
+  };
+
+  const [theme, setTheme] = useState(defaultThemeWithMode);
+  let muiTheme = createTheme(defaultThemeWithMode);
   try {
     // if a bad value is saved this will fail.
     muiTheme = createTheme(theme);
@@ -30,10 +39,10 @@ const App = ({ children }) => {
   const handleSetTheme = newPalette => {
     if (newPalette === undefined) {
       // happens when OrganizationWrapper unmounts
-      setTheme(defaultTheme);
+      setTheme(defaultThemeWithMode);
     } else {
       try {
-        const newTheme = formatTheme(newPalette);
+        const newTheme = formatTheme(newPalette, darkMode);
         setTheme(newTheme);
       } catch (e) {
         console.error("Failed to parse theme: ", newPalette);
@@ -55,4 +64,16 @@ App.propTypes = {
   children: PropTypes.object
 };
 
-export default App;
+const queries = {
+  user: {
+    query: gql`
+      query getCurrentUser {
+        currentUser {
+          darkMode
+        }
+     }`
+  },
+}
+export const operations = { queries };
+
+export default loadData(operations)(App);
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -28,6 +28,7 @@ const App = ({ children }) => {
     console.error("failed to create theme", theme);
   }
   const handleSetTheme = newPalette => {
+    console.log('handle set theme')
     if (newPalette === undefined) {
       // happens when OrganizationWrapper unmounts
       setTheme(defaultTheme);
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -28,7 +28,6 @@ const App = ({ children }) => {
     console.error("failed to create theme", theme);
   }
   const handleSetTheme = newPalette => {
-    console.log('handle set theme')
     if (newPalette === undefined) {
       // happens when OrganizationWrapper unmounts
       setTheme(defaultTheme);
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -1,33 +1,22 @@
 import PropTypes from "prop-types";
-import React, { useEffect, useState } from "react";
-import { ThemeProvider } from "@material-ui/core/styles";
-import { createTheme } from "@material-ui/core/styles";
+import React, { useCallback, useEffect, useState } from "react";
+import { ThemeProvider , createTheme } from "@material-ui/core/styles";
 import CssBaseline from "@material-ui/core/CssBaseline";
+import Sun from "@material-ui/icons/Flare";
+import Moon from "@material-ui/icons/NightsStay";
 
+import Switch from "@material-ui/core/Switch";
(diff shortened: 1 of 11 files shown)


stefanhayden:
The logic for how you are flipping from dark to light seems sound to me. Though I don't think folks will even want the toggle in the top nav. I would guess it should either be moved the be in the account dropdown (i'm not sure material ui supports a toggle there) or the user account page.

some things to check on:
- how does it behave as you switch from one org to another org?
- what should the behavior be on page refresh? do we preserve the last choice or go back to light mode?
- can you document how every page looks? This is hard as there are lots of different extensions and texter-sideboxes to turn on that might not come configured by default.
- Does this just work for future texter-sideboxes and what documentation do we need for those developers?
- Should there be an admin feature to turn this feature off for everyone? I'm usually not all the things spoke let admins toggle off but that is often how this app roles. An admin toggle would either need to be a new environment variable or for a nicer in app toggle a database entry.

mcverter:
> The logic for how you are flipping from dark to light seems sound to me. Though I don't think folks will even want the toggle in the top nav. I would guess it should either be moved the be in the account dropdown (i'm not sure material ui supports a toggle there) or the user account page.
> 
> some things to check on:
> 
> * how does it behave as you switch from one org to another org?
> * what should the behavior be on page refresh? do we preserve the last choice or go back to light mode?
> * can you document how every page looks? This is hard as there are lots of different extensions and texter-sideboxes to turn on that might not come configured by default.
> * Does this just work for future texter-sideboxes and what documentation do we need for those developers?
> * Should there be an admin feature to turn this feature off for everyone? I'm usually not all the things spoke let admins toggle off but that is often how this app roles. An admin toggle would either need to be a new environment variable or for a nicer in app toggle a database entry.

This implementation is a somewhat heavy-handed, client-side implementation.  It switches the entire site globally from light to dark mode.  I was assuming that this would make more sense to the user -- that they would want an easily accessible switch they could flip at any time that would change the scheme for the entire site rather than having to configure light/dark for every organization. 
If we want to configure light/dark per organization, we can easily add that to the organization settings panel.

These changes are not persisted to any data store. So on refresh everything will revert to light mode.  I guess I could try stashing this light/dark option in local storage to deal with the refresh.  

We could persist changes to the database either in user settings or organization settings.  My assumption was that a user would want to switch from light to dark with an easily accessible toggle switch rather than going into their settings.  But I am willing to be corrected on this.

I don't know much about the various pages in Spoke so I can't document the various pages.  In the current implementation, it is globally applied to all UIs.  

So to summarize, I see these as being the various possibiities:
1.  A client side switch to globally toggle all pages between light and dark that is persisted to LocalStorage.
2. A user setting to globally toggle all pages between light and dark that is persisted to Database.
3. Settings to toggle each organization between light and dark.  For this we can just reuse the UI we already have for changing organizational themes.

stefanhayden:
I'm generally going to disagree with a globally available toggle. Though I'm not the final arbiter. This is a nitch feature. Most people don't care about this. Many spoke users are not experts at websites and any additional global button will lead to confusion. This is why I think it should either live in the account dropdown menu or the a user's account page.

I don't think the toggle needs to be per organization. But if a user is switching between organization we should just make sure nothing breaks.

mcverter:
> I'm generally going to disagree with a globally available toggle. Though I'm not the final arbiter. This is a nitch feature. Most people don't care about this. Many spoke users are not experts at websites and any additional global button will lead to confusion. This is why I think it should either live in the account dropdown menu or the a user's account page.
> 
> I don't think the toggle needs to be per organization. But if a user is switching between organization we should just make sure nothing breaks.

Okay.  Thanks for the feedback.  I will add a per-user dark mode option.  I think I can figure out the persistence layer.  I will look at the UserSettings page and the dropdown.  It will be fundamentally the same logic in either place.  Let me circle back in a bit. Thanks so much for all of your feedback.

mcverter:
I have redone this feature to display a checkbox on the User Settings page.  When it is clicked, it saves the `darkMode` value in the user table.

There are a lot of questionable features about this PR and I'm not familiar enough with the codebase yet to resolve them.

* I added a database migration to add a new column in the user table : `darkMode`. 
* If for some reason, the Spoke user updates the code but does not migrate the database, they will get an error when they try to save UserSettings and be unable to save their changes.  I did not observe any other errors.  
* The retrieval of the darkMode value from the graphQL endpoint happens in App.jsx.  I tried to do this through OrganizationWrapper but found a few admin routes that aren't routed through organizations.  I am not wild about adding more code to App.jsx.  Also, because we are retrieving graphQL in App, we have to wrap it in the `loadData` HOC and mock that in tests.
* I added a file for a form element GSCheckbox.  I wasn't able to fully replicate the structure of other form elements.  Let me know if you want me to rework this.
* I wanted to display the UI changes after the user presses the SAVE button.  I was unable to use setTheme() to do this and so I am forcing a reload if the darkMode is toggled.

src/components/App.jsx
```diff
@@ -56,4 +64,16 @@ App.propTypes = {
   children: PropTypes.object
 };
 
-export default App;
+const queries = {
+  user: {
+    query: gql`
+      query getCurrentUser {
+        currentUser {
+          darkMode
```
engelhartrueben, line 73: Add 'id' here before darkmode to clear up this message:
image

engelhartrueben:
Hi @mcverter -
I am unable to save a new theme in the settings page when dark mode is enabled:
image

I am a bit concerned about when orgs have a dark palette theme in place already, and how that might be jarring and difficult to navigate

Here is a valid palette in the features column as an example of my concerns: {"theme":{"palette":{"primary":{"main":"#363d4a"},"secondary":{"main":"#421a1a"},"info":{"main":"#282a2c"},"success":{"main":"#313631"},"warning":{"main":"#44433e"},"error":{"main":"#38302f"}}}}
This primary color is used in one of our orgs. 

Example with above palette and dark mode:
image
and without dark mode:
image

I don't know what the exact solution is. I just know that I would like to try to keep the color palette intact throughout the entire application, regardless of light or dark mode. Maybe changing these specific links to buttons where we can then control the background color there, but there are plenty of links just like this throughout the application that wouldn't make sense to be buttons. Just throwing ideas out there. 

Also might be worth looking into caching this in Redis if it is not already. 

Looking forward to seeing a solution! We do have a hacknight tonight, if you would like to join: https://bit.ly/SpokeHacknight24 , where we can talk further

Thanks for working on this

mcverter:
> Hi @mcverter - I am unable to save a new theme in the settings page when dark mode is enabled: image
> 
> I am a bit concerned about when orgs have a dark palette theme in place already, and how that might be jarring and difficult to navigate
> 
> Here is a valid palette in the features column as an example of my concerns: {"theme":{"palette":{"primary":{"main":"#363d4a"},"secondary":{"main":"#421a1a"},"info":{"main":"#282a2c"},"success":{"main":"#313631"},"warning":{"main":"#44433e"},"error":{"main":"#38302f"}}}} This primary color is used in one of our orgs.
> 
> Example with above palette and dark mode: image and without dark mode: image
> 
> I don't know what the exact solution is. I just know that I would like to try to keep the color palette intact throughout the entire application, regardless of light or dark mode. Maybe changing these specific links to buttons where we can then control the background color there, but there are plenty of links just like this throughout the application that wouldn't make sense to be buttons. Just throwing ideas out there.
> 
> Also might be worth looking into caching this in Redis if it is not already.
> 
> Looking forward to seeing a solution! We do have a hacknight tonight, if you would like to join: https://bit.ly/SpokeHacknight24 , where we can talk further
> 
> Thanks for working on this

I'm not entirely clear what you are seeing.  But that's because I don't know how to navigate Spoke that well.
I am able to create a new org with a new Palette.  And I'm not entirely sure which screens you have screenshot above.  
I do have a blue-themed organization and it's definitely hard to see blue against the black.

One possible solution is to allow dark palettes per organization rather than having a user settings for it.  
Another option is to have an on-screen toggle for this.  But I had a difficul time figuring how to non-invasively insert that into the UI and Stefan did no like that solution

PR #2459, opened 2024-09-20
Remove attempted reassignment of read-only object, nextProps

mau11:
# Fixes # (issue)

## Description
We noticed that each time a text response was received while on the texter view page (.../app/[org-id]/todos), the page would crash (white screen + error in the console). This was caused by an attempt to clear the user.notifications property of nextProps. nextProps however is read only and cannot be updated in this way. Removal of this line allows the notification pop up to display, however it does not disappear until the user clicks on respond or any other action that navigates away from the page.

Screenshot 2024-09-20 at 12 20 32 PM

# Checklist:


engelhartrueben (approved):
🎉

PR #2478, opened 2024-10-04
Allow pausing of initial messages

jeffm2001:
## Description

We've had instances where we needed to pause all our campaigns but still wanted texters to be able to respond to replies for messages that have already gone out. This makes it possible to do that by setting a `DOWNTIME_NO_INITIAL` environment variable. When the environment variable is set, instead of seeing the normal texting interface texters will see a message saying "Sending initial messages is currently paused. You can still reply to any incoming messages."

# Checklist:


engelhartrueben:
I really like the idea, but am hesitant with with instances that have many organizations running/texting concurrently. Also, having to restart the application and set `DOWNTIME_NO_INITIAL` may be too disruptive (again, when there are many orgs). 

Has this been mostly used when there has been something wrong with the application? Or to stop texting in general?

jeffm2001:
The goal was to have something slightly less disruptive than setting `DOWNTIME` or `DOWNTIME_TEXTER`. I think we used it twice, once when we were switching from Twilio to Bandwidth and needed to wind down active campaigns, and once when we were dealing with scaling/instability issues and wanted to scale back activity without shutting down entirely.

Agree that it would be nice if it could be done on a per-organization or even per-campaign basis, but the implementation of that would get trickier.

engelhartrueben (approved):
🎉

